PR #3, opened 2022-07-18
Add CI

richpjames:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,50 @@
...
+        env:
+          POSTGRES_PASSWORD: password
+          POSTGRES_DB: approved-premises-test
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          - 5432:5432
```
pezholio, line 32: We won't be using a database, so we won't need this

.github/workflows/ci.yml
```diff
@@ -0,0 +1,50 @@
...
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+      - develop
+
+env:
+  NODE_ENV: test
+  DATABASE_URL: postgres://postgres:password@localhost:5432/approved-premises-test
```
pezholio, line 12: Same here - we can probably bin this line off

.github/workflows/ci.yml
```diff
@@ -0,0 +1,50 @@
...
+  API_CLIENT_ID: approved-premises
+  API_CLIENT_SECRET: clientsecret
+  OPENSEARCH_URL: http://localhost:9200/
+  OPENSEARCH_INDEX_NAME: bed_availability_test
```
pezholio, line 16: We'll probably be using opensearch later, but more than likely in the backend, so can get rid of these lines

pezholio (approved):
Nice work - LGTM, just a couple of legacy things that sneaked in from Alpha

PR #18, opened 2022-07-19
add import sorting

richpjames:
the current linting rules disallow having local imports above external imports but prettier previously didn’t resolve this manually


richpjames:
We can manually sort for now 😅

PR #24, opened 2022-07-20
Add an ADR to outline our validation approach

pezholio:
This codifies how we’re going to handle validation in the application.

TLDR: We’re going to let the API handle it, and spit back a lightweight JSON response that the frontend app can translate and marshal into a usable format.

### DIFF
--- a/doc/architecture/decisions/0003-handle-validations-errors-on-the-server-side.md
+++ b/doc/architecture/decisions/0003-handle-validations-errors-on-the-server-side.md
@@ -1,0 +1,54 @@
+# 3. Handle validations errors on the server side
+
+Date: 2022-07-20
+
+## Status
+
+Accepted
+
+## Context
+
+We need to figure out a way of how to handle validation errors in the application. As we have
+a UI application and a backend API, it doesn't make sense to do the validation on both the frontend
+and the backend, so we need to identify the best place to do this validation and how to play back
+the errors to the user.
+
+## Decision
+
+As a general rule, we will validate any form requests at the API level. Any invalid requests will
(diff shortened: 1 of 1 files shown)


doc/architecture/decisions/0003-handle-validations-errors-on-the-server-side.md
```diff
@@ -0,0 +1,41 @@
...
+  "errors" : [
+    "arrivalDateBlank",
+    "departureDateBlank",
+    "CRNBlank"
+  ]
+}
+```
+
+The error messages in the `errors` array will be simple camelcase attributes, which can then be
+translated into human-readable errors by the frontend application.
```
edavey, line 33: So I think in this example the UI calling the API will have sent a payload which includes the following properties:

- `arrivalDate`
- `departureDate`
- `CRN`

and then the UI will parse the errors given so that `departureDateBlank` becomes something like:

`field: "departureDate", errorType: "Blank"`

It will then look up the appropriate user-facing error message from config such as YML 'locale' files?

And also use the property name to identify the part of the DOM which needs to be highlighted for correction?

Would it be helpful (to avoid the need for the UI to split strings like `departureDateBlank`) if the API provided the property name and error type separately e.g.

```json
{
 "message": "Validation error",
 "errors" : [
   {
      "propertyName": "arrivalDate",
      "errorType": "Blank"
    },
   {
      "propertyName": "departureDate",
      "errorType": "Invalid"
    },
   {
      "propertyName": "CRN"
      "errorType": "Blank"
   }
 ]
}
```

edavey (approved):
👍 looks good! we can update the Swagger spec in the API repo to show an example...

PR #25, opened 2022-07-20
(445) View premises detail

pezholio:
This adds a simple view page to premises to show some basic details about a given premises when accessing the `/premises/${id}` URL.

## Screenshot

image

### DIFF
--- a/server/@types/approved-premises/index.d.ts
+++ b/server/@types/approved-premises/index.d.ts
@@ -1,9 +1,39 @@
 declare module 'approved-premises' {
   export type Premises = schemas['Premises']
 
-  export type TableCell = { text: string; attributes?: { [key: string]: string } } | { html: string }
+  export interface HtmlAttributes {
+    [key: string]: string
+  }
+
+  export interface TextItem {
+    text: string
+  }
+
+  export interface HtmlItem {
+    html: string
+  }
+
+  export type TableCell = { text: string; attributes?: HtmlAttributes } | { html: string }
   export interface TableRow {
     [index: number]: TableCell
+  }
+
+  export interface SummaryListActionItem {
+    href: string
+    text: string
+    visuallyHiddenText: string
+  }
+
+  export interface SummaryListItem {
+    key: TextItem | HtmlItem
+    value: TextItem | HtmlItem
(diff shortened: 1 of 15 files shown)


server/@types/approved-premises/index.d.ts
```diff
@@ -1,11 +1,41 @@
...
+  export interface HtmlAttributes {
+    [key: string]: string
+  }
+
+  export interface TextItem {
+    text: string
+  }
+
+  export interface HtmlItem {
+    text: string
```
richpjames, line 13: I think this should be `html`

richpjames: wishful thinking but it would be ideal if we could download GDS DS types from an NPM package 😄

server/views/premises/show.njk
```diff
@@ -0,0 +1,22 @@
...
+{% set pageTitle = applicationName + " - " + name %}
+
+{% block content %}
+
+	{{ govukBackLink({
+		text: "Back",
+		href: "/premises"
+	}) }}
+
+	<main class="app-container govuk-body">
```
richpjames, line 14: I know this is used elsewhere in the template repo but we shouldn't be using a `<main>` element inside the `content` block as there is already a `<main>` element in the `content` block and [the HTML spec states there should only be one](https://html.spec.whatwg.org/multipage/grouping-content.html#the-main-element) `<main>` in the document at any time. 

We can set the classes onto that `<main>` in the `content` block by doing `{% set mainClasses = "app-container govuk-body" %}`

Do you think I should put in a PR to fix this in the template repo at some point?

richpjames: This also slipped past me in your previous PR :D

richpjames (approved):
Looks good 🕶 just left a few comments

server/@types/approved-premises/index.d.ts
pezholio: I'm planning on extracting these at some point I think, would be really useful for other folks I'm sure

server/views/premises/show.njk
pezholio: Nice catch. Yeah, definitely worth fixing in the template repo too if you get a chance 👍

PR #26, opened 2022-07-21
446 create a booking

richpjames:
# Context
We need to be able to take a booking. This PR creates that functionality. 
We have a simple form that takes in the requisite details for a booking and then on submission posts the details to the backend.
There is no validation on the data in either the view or controller level currently as this will be handled entirely in the API.
This PR doesn't have the unhappy path scenario where the data isn't valid as that will be added in a subsequent PR.

[Trello Ticket](https://trello.com/c/o5ADpFax/446-create-a-booking)

### After
image
image

### DIFF
--- a/integration_tests/e2e/booking.cy.ts
+++ b/integration_tests/e2e/booking.cy.ts
@@ -1,0 +1,52 @@
+import premisesFactory from '../../server/testutils/factories/premises'
+import bookingFactory from '../../server/testutils/factories/booking'
+import BookingPage from '../pages/booking'
+import PremisesPage from '../pages/premises'
+import Page from '../pages/page'
+
+context('Booking', () => {
+  beforeEach(() => {
+    cy.task('reset')
+    cy.task('stubSignIn')
+    cy.task('stubAuthUser')
+  })
+
+  it('should show booking form', () => {
+    const booking = bookingFactory.build({
+      CRN: '1bee477b-462f-47c1-8f71-7835a76a2c42',
+      arrivalDate: new Date(Date.UTC(2022, 5, 1, 0, 0, 0)),
+      expectedDepartureDate: new Date(Date.UTC(2022, 5, 3, 0, 0, 0)),
+      keyWorker: 'Lizeth Hagenes',
+    })
+
+    const premises = premisesFactory.buildList(5)
+    cy.task('stubBookingCreate', { premisesId: 'premisesId', booking })
+    cy.task('stubPremises', premises)
+
+    // Given I am signed in
+    cy.signIn()
+
+    // When I visit the booking page
+    const page = BookingPage.visit()
+
+    // And I fill in the booking form
+    page.completeForm(booking)
+    page.clickSubmit()
+
+    // Then I should be redirected to the premises page
+    Page.verifyOnPage(PremisesPage)
+
+    PremisesPage.shouldShowFlashMessage()
+
+    // And the booking should be created in the API
+    cy.task('verifyBookingCreate', { premisesId: 'premisesId' }).then(requests => {
+      expect(requests).to.have.length(1)
+      const requestBody = JSON.parse(requests[0].body)
+
+      expect(requestBody.CRN).equal('1bee477b-462f-47c1-8f71-7835a76a2c42')
+      expect(requestBody.arrivalDate).equal((booking.arrivalDate as Date).toISOString())
+      expect(requestBody.expectedDepartureDate).equal((booking.expectedDepartureDate as Date).toISOString())
+      expect(requestBody.keyWorker).equal('Lizeth Hagenes')
+    })
+  })
+})
--- a/integration_tests/e2e/booking.cy.ts
+++ b/integration_tests/e2e/booking.cy.ts
@@ -1,8 +1,8 @@
 import premisesFactory from '../../server/testutils/factories/premises'
 import bookingFactory from '../../server/testutils/factories/booking'
 import BookingPage from '../pages/booking'
-import PremisesPage from '../pages/premises'
 import Page from '../pages/page'
+import PremisesPage from '../pages/premisesList'
 
 context('Booking', () => {
   beforeEach(() => {
@@ -35,7 +35,6 @@ context('Booking', () => {
 
     // Then I should be redirected to the premises page
     Page.verifyOnPage(PremisesPage)
-
     PremisesPage.shouldShowFlashMessage()
 
     // And the booking should be created in the API
--- a/integration_tests/pages/premisesList.ts
+++ b/integration_tests/pages/premisesList.ts
@@ -23,4 +23,8 @@ export default class PremisesListPage extends Page {
         })
     })
   }
+
+  static shouldShowFlashMessage(): void {
+    cy.get('.govuk-notification-banner__content').should('contain', 'Booking made successfully')
+  }
 }
--- a/server/@types/approved-premises/index.d.ts
+++ b/server/@types/approved-premises/index.d.ts
@@ -1,5 +1,12 @@
 declare module 'approved-premises' {
   export type Premises = schemas['Premises']
+  export type Booking = schemas['Booking']
+
+  export type BookingDto = Omit<Booking, 'id'>
+
+  export type ObjectWithDateParts<K extends string | number> = { [P in `${K}-${'year' | 'month' | 'day'}`]: string } & {
+    [P in K]?: Date
+  }
 
   export interface HtmlAttributes {
     [key: string]: string
@@ -44,6 +51,13 @@ declare module 'approved-premises' {
       postcode: string
       bedCount: number
       apAreaId: string
+    }
+    Booking: {
+      id: string
+      CRN: string
+      arrivalDate: Date | string
+      expectedDepartureDate: Date | string
+      keyWorker: string
     }
   }
 }
--- a/server/controllers/bookingsController.test.ts
+++ b/server/controllers/bookingsController.test.ts
@@ -35,7 +35,7 @@ describe('bookingsController', () => {
     it('given the expected form data, the posting of the booking is successful should redirect to the "premises" page', async () => {
       const booking = BookingFactory.build()
       bookingService.postBooking.mockResolvedValue(booking)
-
+      const mockFlash = jest.fn()
       const requestHandler = bookingController.create()
 
       await requestHandler(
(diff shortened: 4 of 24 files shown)


edavey:
This looks great -- nice work breaking the ice here on our first form!

A couple of high level initial comments/questions:

1. Probably the "key worker" is going to be a select list of names (and `keyworkerId`s) so maybe we should embrace that sooner rather than later? e.g. using a list of 3 dummy keyworkers to illustrate? What do you think? And would that be a job for now or a later PR?
2. are you thinking that in a later PR we'll change the controller to redirect to a "booking view" so that the user can get an immediate confirmation of the details of the booking they've just made?

richpjames:
Thanks @edavey! 

1. That is useful to know, I didn't know what format the Key Worker input would take so I went with the simplest option. A select list makes a lot more sense. I agree that it would be good to get this in as soon as possible but my preference would be for a later PR. 

2. That sounds like a good idea. I asked @pezholio what should happen on submitting and he suggested the approach I've taken (if I understood him correctly!) but long term we will absolutely need some kind of booking confirmation view. I think it would make sense to do this in a later PR on this same ticket.

I'm interested as to whether these seems like sensible boundaries to draw for this PR or whether you'd like to see all these changes in one easy to revert PR? My preference as a writer and reviewer would be for PRs to be as small as possible whilst still containing a useful piece of functionality but I can see an argument that both these changes (and handling a 400 from the server) should be in one PR. Interested to hear what you think too @pezholio

server/@types/approved-premises/index.d.ts
```diff
@@ -45,5 +46,12 @@ declare module 'approved-premises' {
       bedCount: number
       apAreaId: string
     }
+    Booking: {
+      id?: string
```
richpjames, line 50: As mentioned in the commit message this Id is nullable as it won't be there on the request but it will be on the response (assuming it's created on the server). 
I'm not sure if this is a good approach or if we should have two seperate types to more clearly express what's going on

pezholio: Yeah, I think having a `BookingDto` would be better. What I've done in my PR is using the `Omit` utility class, so for booking, it'd look something like this:

```
type BookingDto = Omit<Booking, 'id'>
```

wiremock/stubApis.ts
```diff
@@ -37,6 +37,21 @@ premises.forEach(p => {
...
+      request: {
+        method: 'POST',
+        url: `/premises/${p.id}/booking/new`,
+      },
+      response: {
+        status: 200,
+        headers: {
+          'Content-Type': 'application/json;charset=UTF-8',
+        },
+        body: "{id: 'bookingID', {{request.body}}}",
```
pezholio, line 51: Would `{id: 'bookingID', ...request.body}` not do the same thing here?

server/controllers/bookingsController.ts
```diff
@@ -0,0 +1,42 @@
...
+      const { CRN, keyWorker }: Booking = req.body
+      const arrivalDay = req.body['arrival-day']
+      const arrivalMonth = req.body['arrival-month']
+      const arrivalYear = req.body['arrival-year']
+      const expectedDepartureDay = req.body['expected-departure-day']
+      const expectedDepartureMonth = req.body['expected-departure-month']
+      const expectedDepartureYear = req.body['expected-departure-year']
+
+      const arrivalDate = createDate(arrivalDay, arrivalMonth, arrivalYear)
+      const expectedDepartureDate = createDate(expectedDepartureDay, expectedDepartureMonth, expectedDepartureYear)
```
pezholio, line 30: I've got another way of creating dates in my branch. Would be good if you could take a look at my method and then standardise on one way or another 👍

integration_tests/e2e/booking.cy.ts
```diff
@@ -0,0 +1,48 @@
...
+    page.arrivalMonth().type('07')
+    page.arrivalYear().type('22')
+
+    page.getLegend('What is the expected departure date?')
+    page.expectedDepartureDay().type('30')
+    page.expectedDepartureMonth().type('12')
+    page.expectedDepartureYear().type('22')
+
+    page.getLabel('Key Worker')
+    page.getByInputByIdAndEnterDetails('keyWorker', 'John Doe')
```
pezholio, line 41: I'd prefer it if we could hide the implementation in the page object and pass in a booking that has been created by a factory, so somthing like:

```typescript
const booking = bookingFactory.build()
page.completeBookingForm(booking)
```

Then the page object would have a `completeBookingForm` method that looked like this:

```typescript
this.getLabel('CRN')
this.getByInputByIdAndEnterDetails('CRN', booking.crn)

this.getLegend('What is the arrival date?')
this.arrivalDay().type(booking.arrivalDate.day())
this.arrivalMonth().type(booking.arrivalDate.month())
this.arrivalYear().type(booking.arrivalDate.year())

this.getLegend('What is the expected departure date?')
this.expectedDepartureDay().type(booking.departureDate.day())
this.expectedDepartureMonth().type(booking.departureDate.month())
this.expectedDepartureYear().type(booking.departureDate.year())

this.getLabel('Key Worker')
this.getByInputByIdAndEnterDetails('keyWorker', booking.keyWorker)
```

integration_tests/e2e/booking.cy.ts
```diff
@@ -0,0 +1,48 @@
...
+    cy.task('stubBooking', booking)
+    cy.task('stubPremises', premises)
+
+    // Given I am signed in
+    cy.signIn()
+
+    // When I visit the booking page
+    const page = BookingPage.visit()
+
+    // Then I should see the booking form and enter details
```
pezholio, line 26: I feel like this a should be `And I fill in the booking form`

integration_tests/e2e/booking.cy.ts
```diff
@@ -0,0 +1,48 @@
...
+    page.getLegend('What is the expected departure date?')
+    page.expectedDepartureDay().type('30')
+    page.expectedDepartureMonth().type('12')
+    page.expectedDepartureYear().type('22')
+
+    page.getLabel('Key Worker')
+    page.getByInputByIdAndEnterDetails('keyWorker', 'John Doe')
+
+    // And be able to click submit
+    page.clickSubmit()
```
pezholio, line 44: This could be rolled into the previous step

integration_tests/e2e/booking.cy.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    Page.verifyOnPage(PremisesPage)
```
pezholio, line 46: This should be an expectation on its own, something like:

```
// Then I should be redirected to the premises page
```

Could we also set an expectation that the expected data gets passed to the API too?

integration_tests/pages/premisesList.ts
```diff
@@ -23,4 +23,8 @@ export default class PremisesListPage extends Page {
         })
     })
   }
+
+  static shouldShowFlashMessage(): void {
+    cy.get('.govuk-notification-banner__content').should('contain', 'Booking made successfully')
```
pezholio, line 28: 👍

pezholio (approved):
This is ace, thanks! Sorry for the bombardment of comments, just want to make sure we're all aligned style wise. Feel free to push back if you think I'm barking up the wrong tree on anything 👍

wiremock/stubApis.ts
richpjames: to be honest I'm not sure, I was following the [Wiremock docs](https://wiremock.org/docs/response-templating/) for this.
We don't have access to a 'request' variable but I think the double braces syntax allows Wiremock to do some magic behind the scenes.

server/controllers/bookingsController.test.ts
```diff
@@ -0,0 +1,136 @@
...
+      expect(bookingService.postBooking).toHaveBeenCalledWith('premisesIdParam', {
+        CRN: 'CRN',
+        arrivalDate: '2022-02-01',
+        expectedDepartureDate: '2023-02-01',
+        keyWorker: 'John Doe',
+      })
+
+      expect(response.redirect).toHaveBeenCalledWith('/premises')
+    })
+    it('given the form is submitted with no data the posting of the booking is successful should render the "new" page', async () => {
```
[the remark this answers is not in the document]
richpjames: Agree, we should find a prettier setting for this

integration_tests/e2e/booking.cy.ts
richpjames: Yeah I agree that your suggested approach is more readable

integration_tests/e2e/booking.cy.ts
richpjames: I think that would be wise. Do you suggest doing this via Wiremock?

pezholio: Yeah, I reckon so

pezholio: Should be able to use `getMatchingRequests`

server/utils/utils.ts
```diff
@@ -21,3 +23,25 @@ export const initialiseName = (fullName?: string): string | null => {
...
+    Number(dateInputObj[`${key}-day`]),
+    Number(dateInputObj[`${key}-month`]),
+    Number(dateInputObj[`${key}-year`]),
+  ]
+  const o: { [P in K]?: Date } = dateInputObj
+  if (day && month && year) {
+    o[key] = new Date(Date.UTC(year, month - 1, day, 0, 0, 0))
+  } else {
+    o[key] = new Date(0, 0, 0)
+  }
```
pezholio, line 45: I wonder if we should raise an error if there is no date present. We don't really want to pass an invalid date to the API

pezholio (approved):
LGTM 👍 I do think we need to think about how we handle missing dates and if we should even pass missing / invalid dates to the API, but we can think about that in more detail later

server/utils/utils.ts
richpjames: I did think about that... this would result in a strange check in the API but otherwise we would have to show a validation error to the user which might be awkward to implement if all the other validaiton is handled via the API

edavey:
Just to say that I think it makes sense (certainly at this point) to leave all the validation to the API as previously discussed.

Is there something particularly problematic about a missing or invalid date which is concerning you?

pezholio:
@edavey The issues I can see at the moment are:

1) If someone provides invalid date parts (i.e. 99/99/9999), this will currently create a seemingly valid date (because JavaScript is weird)
2) If someone leaves a date blank, the parser will create a date anyway, which won't necessarily be what we want

pezholio:
I'm definitely all for handling validation at the API side, but I do think some things that are obviously wrong (like invalid date parts) might have to be something we handle in the frontend, because this is the point where we're transforming user input into something API-shaped. I certainly don't think we should POST the individual date parts to the API and handle the transformation there, because we're allowing the frontend to dictate the shape of the API requests, and this may change in future and potentially make things brittle.

PR #34, opened 2022-07-26
446 Add key worker select input

richpjames:
# Context
[From feedback on a previous PR](https://github.com/ministryofjustice/approved-premises-ui/pull/26#issuecomment-1191759262) I learned that the key worker input on the create a booking form will need to allow the user to select a key worker from a list of key workers and we will need to send their ID to the API.
[Trello Card](https://trello.com/c/o5ADpFax/446-create-a-booking)

# Changes in this PR
In order to get this input in sooner rather than later I have create an select input using mocked Key Worker names and hardcoded IDs.
In future the Key Worker names and their IDs will have to come from somewhere else but this should be easier to implement now we have the correct input in place. 

### Before
image

### After
image


pezholio (approved):
LGTM - we probably should look at returning key workers as an array when we get to displaying this data, but that can come later

PR #36, opened 2022-07-26
(503) Rework the date parser

pezholio:
After some discussion, we've decided to make a best attempt to transform date parts into an ISO8601 date string, and then pass it to the API. If the string is invalid, we will return an error from to API.

With this in mind, I've also updated all the objects that currently return either a Date or an ISO string, and updated them to only expect a string. If we want to do any conversion when the data hits the frontend, we can do this in a more predictable way.

### DIFF
--- a/integration_tests/e2e/arrivals.cy.ts
+++ b/integration_tests/e2e/arrivals.cy.ts
@@ -18,7 +18,10 @@ context('Arrivals', () => {
     // And I have a booking for a premises
     const premises = premisesFactory.build()
     const bookingId = 'some-uuid'
-    const arrival = arrivalFactory.build()
+    const arrival = arrivalFactory.build({
+      dateTime: new Date(2022, 1, 11).toISOString(),
+      expectedDeparture: new Date(2022, 11, 11).toISOString(),
+    })
 
     cy.task('stubSinglePremises', premises)
     cy.task('stubArrivalCreate', { premisesId: premises.id, bookingId, arrival })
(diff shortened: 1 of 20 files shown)


pezholio:
[Rework the Date parser](https://trello.com/c/CkKFFOTv/503-rework-the-date-parser)

server/controllers/bookingsController.ts
```diff
@@ -18,32 +18,14 @@ export default class BookingsController {
...
-        arrivalDate,
-        expectedDepartureDate,
-        keyWorker,
-      })
+
+      const booking = {
+        ...req.body,
+        ...convertDateInputsToIsoString(req.body, 'arrivalDate'),
+        ...convertDateInputsToIsoString(req.body, 'expectedDepartureDate'),
+      } as BookingDto
```
richpjames, line 26: I would avoid casting the `booking` object, I think we can just type it on line 22.

integration_tests/e2e/arrivals.cy.ts
```diff
@@ -32,19 +35,11 @@ context('Arrivals', () => {
...
-      const dateTime = arrival.dateTime as Date
-      const expectedDeparture = arrival.expectedDeparture as Date
-
-      const expectedDateTime = new Date(
-        Date.UTC(dateTime.getFullYear(), dateTime.getMonth(), dateTime.getDate(), 0, 0, 0),
-      )
-      const expectedExpectedDeparture = new Date(
-        Date.UTC(expectedDeparture.getFullYear(), expectedDeparture.getMonth(), expectedDeparture.getDate(), 0, 0, 0),
-      )
+      const { dateTime, expectedDeparture } = arrival
```
richpjames, line 38: Well this is a big improvement!

integration_tests/pages/premisesShow.ts
```diff
@@ -31,12 +31,11 @@ export default class PremisesShowPage extends Page {
 
   shouldShowBookings(bookings: Array<Booking>): void {
     bookings.forEach((item: Booking) => {
+      const arrivalDate = new Date(Date.parse(item.arrivalDate))
```
richpjames, line 34: Could we use `date-fns` `parseISO` here?

richpjames (approved):
Looks good, thanks for doing this! 
I've left some comments

PR #37, opened 2022-07-27
446 booking confirmation page

richpjames:
# Context
We need to be able to show users the booking they've just created so that they can confirm the details.
Previously we redirected them to a list of the premises but now we are taking them to a new booking confirmation screen.
[Trello ticket](https://trello.com/c/o5ADpFax/446-create-a-booking)

# Changes in this PR

## Screenshots of UI changes
image

### DIFF
--- a/integration_tests/e2e/booking.cy.ts
+++ b/integration_tests/e2e/booking.cy.ts
@@ -2,7 +2,7 @@ import premisesFactory from '../../server/testutils/factories/premises'
 import bookingFactory from '../../server/testutils/factories/booking'
 import BookingPage from '../pages/booking'
 import Page from '../pages/page'
-import PremisesPage from '../pages/premisesList'
+import BookingConfirmation from '../pages/bookingConfirmation'
 
 context('Booking', () => {
   beforeEach(() => {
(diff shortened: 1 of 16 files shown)


pezholio:
Looking good - sorry to be a huge pain, but I do wonder if we should maybe not send the booking details to the API until they've been confirmed? So when we submit from the form, we store them as session variables, so the user can go back and make changes. Only when they've hit the button on the confirmation screen, does it get sent to the API - does that make sense?

pezholio:
I know this probably contradicts something I said previously, sorry!

richpjames:
> I know this probably contradicts something I said previously, sorry!

It does but it's fine. Do you think I should leave the `getBooking` changes (the client + service) in this PR? or YAGNI them out of here? I think it makes the PR a bit strange but probably worth keeping around as we will probably need that endpoint in future

pezholio:
Yeah, let's keep it for now - sorry to mess you around! 👍

server/controllers/bookingsController.ts
```diff
@@ -30,4 +32,17 @@ export default class BookingsController {
...
+
+  confirm(): RequestHandler {
+    return async (req: Request, res: Response) => {
+      const { premisesId, bookingId } = req.params
+      const booking = await this.bookingService.getBooking(premisesId, bookingId)
+
+      return res.render('premises/bookings/confirm', {
+        ...booking,
+        arrivalDate: parseISO(booking.arrivalDate).toLocaleDateString('en-GB'),
+        expectedDepartureDate: parseISO(booking.expectedDepartureDate).toLocaleDateString('en-GB'),
```
pezholio, line 44: I think I'd prefer to see this conversion in the service, in the interests of keeping our controllers as slim as possible 👍

integration_tests/e2e/booking.cy.ts
```diff
@@ -19,26 +19,24 @@ context('Booking', () => {
...
     page.clickSubmit()
 
-    // Then I should be redirected to the premises page
-    Page.verifyOnPage(PremisesPage)
-    PremisesPage.shouldShowFlashMessage()
-
-    // And the booking should be created in the API
-    cy.task('verifyBookingCreate', { premisesId: 'premisesId' }).then(requests => {
+    // Then I should be redirected to the confirmation page and the booking should be created in the API
+    Page.verifyOnPage(BookingConfirmation)
```
pezholio, line 38: I think we're OK to have the booking confirmation assertion in this part. I get that it makes the test a bit longer, but I think it's an OK tradeoff vs having another spec file to maintain.

pezholio (approved):
LGTM - Just a couple of minor things 👍

server/controllers/bookingsController.ts
richpjames: Yep will do this

integration_tests/e2e/booking.cy.ts
richpjames: Sure!

PR #40, opened 2022-07-28
450 group bookings into types

pezholio:
This groups bookings into types in a tabbed view. To get this to work properly, we needed to add a `status` field to the API, which has been documented in Swagger. I've also added some traits to the booking factory, which makes it easier to create types of bookings in test and for stubs.

# Screenshot

image

### DIFF
--- a/server/testutils/factories/booking.ts
+++ b/server/testutils/factories/booking.ts
@@ -1,16 +1,77 @@
 import { Factory } from 'fishery'
 import { faker } from '@faker-js/faker/locale/en_GB'
+import { addDays, startOfToday } from 'date-fns'
 
 import type { Booking } from 'approved-premises'
 import arrivalFactory from './arrival'
 
-export default Factory.define<Booking>(() => ({
+const today = startOfToday().toISOString()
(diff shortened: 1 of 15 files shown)


pezholio:
[group bookings into types](https://trello.com/c/ec0nyUSB/450-group-bookings-into-types)

server/testutils/factories/booking.ts
```diff
@@ -1,15 +1,76 @@
 import { Factory } from 'fishery'
 import { faker } from '@faker-js/faker/locale/en_GB'
+import { addDays } from 'date-fns'
 
 import type { Booking } from 'approved-premises'
 import arrivalFactory from './arrival'
 
-export default Factory.define<Booking>(() => ({
+const today = new Date(new Date().setHours(0, 0, 0, 0)).toISOString()
```
richpjames, line 8: It took me a while to work out what this is doing. 
I suggest we use [startOfToday](https://date-fns.org/v2.28.0/docs/startOfToday) from `date-fns` for clarity. 
Or we could save the `new Date().setHours(0, 0, 0, 0)` as a variable called `startOfToday` and use that in the second `new Date()`

richpjames (approved):
LGTM
Left a couple of minor comments

PR #41, opened 2022-07-29
(500) Create a non arrival

richpjames:
# Context
People can arrive to their AP bookings but they can also not arrive.
We need to be able to record non-arrivals on a booking.

# Changes in this PR

When a user clicks 'Manage' on a booking on the AP detail view they can select if the person who is booked to arrived or did not. Previously when a user select that a user did not arrive nothing happened. 
This PR adds a mini form where the user can select a reason for the non-arrival and add notes about the non-arrival.
The client, service and controller are very similar to the components for arrivals. The UI is similar but with the fields 'non-arrival date', 'reason' and 'notes'.
[Trello ticket](https://trello.com/c/anTARtGW/500-creating-non-arrival)
## Screenshots of UI changes
image

### DIFF
--- a/integration_tests/pages/arrivalCreate.ts
+++ b/integration_tests/pages/arrivalCreate.ts
@@ -17,20 +17,20 @@ export default class ArrivalCreatePage extends Page {
 
     cy.log('arrival', arrival)
 
-    const dateTime = new Date(Date.parse(arrival.dateTime))
-    const expectedDeparture = new Date(Date.parse(arrival.expectedDeparture))
+    const date = new Date(Date.parse(arrival.date))
+    const expectedDeparture = new Date(Date.parse(arrival.expectedDepartureDate))
 
-    cy.get('input[name="dateTime-day"]').type(String(dateTime.getDate()))
-    cy.get('input[name="dateTime-month"]').type(String(dateTime.getMonth() + 1))
-    cy.get('input[name="dateTime-year"]').type(String(dateTime.getFullYear()))
+    cy.get('input[name="date-day"]').type(String(date.getDate()))
+    cy.get('input[name="date-month"]').type(String(date.getMonth() + 1))
+    cy.get('input[name="date-year"]').type(String(date.getFullYear()))
 
-    cy.get('input[name="expectedDeparture-day"]').type(String(expectedDeparture.getDate()))
-    cy.get('input[name="expectedDeparture-month"]').type(String(expectedDeparture.getMonth() + 1))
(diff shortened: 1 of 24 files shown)


integration_tests/pages/arrivalCreate.ts
```diff
@@ -28,8 +28,26 @@ export default class ArrivalCreatePage extends Page {
     cy.get('input[name="expectedDeparture-month"]').type(String(expectedDeparture.getMonth() + 1))
     cy.get('input[name="expectedDeparture-year"]').type(String(expectedDeparture.getFullYear()))
 
-    cy.get('textarea[name="notes"]').type(arrival.notes)
+    cy.get('#conditional-arrived > form >> textarea[name="notes"]').type(arrival.notes)
```
pezholio, line 31: It might be easier to use the nunjucks `attributes` option to add `data-cy-*` attributes to make selecting these elements simpler and less brittle. Perhaps even adding `data-cy-arrival` and `data-cy-non-arrival` attributes to the form might be simplest.

server/views/arrivals/_non-arrival_form.njk
```diff
@@ -0,0 +1,48 @@
+<form action="/premises/{{premisesId}}/bookings/{{bookingId}}/nonArrivals" method="post">
+  <input type="hidden" name="_csrf" value="{{ csrfToken }}"/>
+  {{ govukDateInput({
+    id: "date",
```
pezholio, line 4: These IDs are now duplicated on the page, which would result in invalid HTML (and probably some accessiblity issues). Might be better to namespace the IDs with the attribute type (maybe `nonArrival[date]` in this context)

pezholio (approved):
LGTM - just a couple of things around the form

integration_tests/pages/arrivalCreate.ts
richpjames: Yeah this is a good point, I will do the `data-cy` attributes as you suggest

server/views/arrivals/_non-arrival_form.njk
richpjames: a very good point!

richpjames:
@pezholio rather than add data attributes in the nunjucks template I have opted to namespace the reason and notes for consistency. I wish we could do the same with notes but it's not possible using the govuk macro. 
I think it makes sense and could be a pattern we carry forward

pezholio (commented):
Cool - looks good. Do you think it's worth adding the namespacing to the arrivals form in this PR for consistency?

pezholio (approved):
LGTM 👍

PR #42, opened 2022-07-29
Add code coverage checking

richpjames:
# Context
It's [dxw policy](https://github.com/dxw/tech-team-rfcs/blob/main/rfc-173-testing-our-work.md) to have 100% test coverage for our code.
Previously this check wasn't automated so we could've merged untested code but this didn't happen.
I have opted to go for 100% coverage of functions as that seemed sensible but [we can set coverage by branches, lines and statements too](https://jestjs.io/docs/configuration#coveragethreshold-object).
I have chosen to ignore some config/setup files as it doesn't make sense to test them. The overwhelming majority of these files come from [the template](https://github.com/ministryofjustice/hmpps-template-typescript) and the others are files that connect things together.
Most of these can be ignored by ignoring the `/middleware/` and `/authentication/` folders but I have had to manually ignore some files too.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,9 +37,17 @@
   },
   "jest": {
     "preset": "ts-jest",
+    "collectCoverage": true,
     "collectCoverageFrom": [
-      "server/**/*.{ts,js,jsx,mjs}"
+      "server/**/*.{ts,js,jsx,mjs}",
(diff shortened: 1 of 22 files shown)


pezholio (approved):
This is *chef's kiss*. Thanks for doing this!

PR #45, opened 2022-08-01
(449) Add current resident info to premises

richpjames:
# Context
We need to be able to see who is currently resident in an AP in order to perform tasks on that booking - for example soon we will want to mark them as departed.
[Trello ticket](https://trello.com/c/65asE3GA/449-add-current-resident-information-to-premises)

# Changes in this PR
We have added a table of the current residents, their expected departure date and a link to 'Manage' them.
The 'Manage' link does nothing for now but later we will add functionality to mark the resident as departed.

### Before
image

### After
image

### DIFF
--- a/server/controllers/premisesController.test.ts
+++ b/server/controllers/premisesController.test.ts
@@ -1,7 +1,7 @@
 import type { Request, Response, NextFunction } from 'express'
 import { createMock, DeepMocked } from '@golevelup/ts-jest'
 
-import type { SummaryListItem, GroupedListofBookings } from 'approved-premises'
+import type { SummaryListItem, GroupedListofBookings, TableRow } from 'approved-premises'
 import PremisesService from '../services/premisesService'
 import BookingService from '../services/bookingService'
 import PremisesController from './premisesController'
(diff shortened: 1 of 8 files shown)


server/services/bookingService.ts
```diff
@@ -82,6 +82,34 @@ export default class BookingService {
     ])
   }
 
+  async currentResidents(premisesId: string): Promise<Array<TableRow>> {
+    const token = 'FAKE_TOKEN'
+    const bookingClient = this.bookingClientFactory(token)
+
+    const bookings = await bookingClient.allBookingsForPremisesId(premisesId)
+
+    return this.currentResidentsToTableRows(bookings, premisesId)
```
pezholio, line 91: This doesn't seem to filter for residents who have arrived?

server/controllers/premisesController.test.ts
```diff
@@ -36,18 +36,21 @@ describe('PremisesController', () => {
     it('should return the premises detail to the template', async () => {
       const premises = { name: 'Some premises', summaryList: { rows: [] as Array<SummaryListItem> } }
       const bookings = createMock<GroupedListofBookings>()
+      const currentResidents = createMock<Array<{ text: string; html: string }>>()
```
pezholio, line 39: Could we do `createMock<Array<TableRow>>` here?

server/services/bookingService.ts
richpjames: oops!

server/controllers/premisesController.test.ts
richpjames: Yes

server/services/bookingService.test.ts
```diff
@@ -141,6 +141,20 @@ describe('BookingService', () => {
...
 
+      expect(bookingClient.allBookingsForPremisesId).toHaveBeenCalledWith(premisesId)
+    })
+  })
+  describe('currentResidents', () => {
+    it('should return table rows of the current residents', async () => {
+      const currentResidents = bookingFactory.arrived().buildList(2)
+
+      const premisesId = 'some-uuid'
+      bookingClient.allBookingsForPremisesId.mockResolvedValue(currentResidents)
```
pezholio, line 152: Wonder if it's worth returning some non-arrived bookings here, so we know we're filtering?

pezholio (approved):
LGTM 👍

PR #50, opened 2022-08-02
(535) Create a departure

richpjames:
# Context
We need to be able to communicate that a current resident has departed the AP.

For now this will be done by clicking 'Manage' against their name in the list of current residents in the Premises detail view. 'Manage' is perhaps not the best verb for now but functionality will be added later to enable multiple actions on a residents booking so will make more sense once this work is completed.
Once 'Manage' has been clicked the user has to fill in a form. As with all forms in this repo so far no validation is performed on the data as validation will be handled by the API.
Once the form has been filled in the user is taken to a confirmation page which confirms the details they have just submitted. 

There isn't any error handling in this PR as it will be completed later.

The form uses radio buttons for some of the inputs which mean the form takes up a lot of vertical space. I believe this is preferable to `select` inputs for usability and accessibility reasons and [GDS guidance states that select inputs should only be used as 'a last resort'](https://design-system.service.gov.uk/components/select/)

On the confirmation page the value of the radio button is presented back to the user as opposed to the more human-readable 'text' property of the input. Later work will included pulling in the values of these radio buttons dynamically from the API so I suggest that once that is done we display the 'text' property - but for now I think it will cause more short term work that will need to be redone in the medium term.

[Trello ticket](https://trello.com/c/xCzYhBuA/535-create-a-departure)

## Screenshots of UI changes
image

### After
image

### DIFF
--- a/server/services/bookingService.ts
+++ b/server/services/bookingService.ts
@@ -101,7 +101,7 @@ export default class BookingService {
         text: convertDateString(booking.expectedDepartureDate).toLocaleDateString('en-GB'),
       },
       {
-        html: `<a href="/premises/${premisesId}/bookings/${booking.id}/">
+        html: `<a href="/premises/${premisesId}/bookings/${booking.id}/departures/new">
         Manage
         <span class="govuk-visually-hidden">
           booking for ${booking.CRN}
--- a/server/services/premisesService.test.ts
+++ b/server/services/premisesService.test.ts
@@ -75,6 +75,24 @@ describe('PremisesService', () => {
     })
   })
 
+  describe('premisesSelectList', () => {
+    it('returns the list mapped into the format required by the nunjucks macro and sorted alphabetically', async () => {
+      const premisesA = premisesFactory.build({ name: 'a' })
+      const premisesB = premisesFactory.build({ name: 'b' })
+      const premisesC = premisesFactory.build({ name: 'c' })
+      premisesClient.getAllPremises.mockResolvedValue([premisesC, premisesB, premisesA, premisesC])
+
+      const result = await service.getPremisesSelectList()
+
+      expect(result).toEqual([
+        { text: 'a', value: 'a' },
+        { text: 'b', value: 'b' },
+        { text: 'c', value: 'c' },
+        { text: 'c', value: 'c' },
+      ])
+    })
+  })
+
   describe('getPremisesDetails', () => {
     it('returns a title and a summary list for a given Premises ID', async () => {
       const premises = premisesFactory.build({ name: 'Test', apCode: 'ABC', postcode: 'SW1A 1AA', bedCount: 50 })
--- a/server/services/premisesService.ts
+++ b/server/services/premisesService.ts
@@ -34,6 +34,25 @@ export default class PremisesService {
     return { name: premises.name, summaryList }
   }
 
+  async getPremisesSelectList(): Promise<Array<{ text: string; value: string }>> {
+    const premisesClient = this.premisesClientFactory(this.token)
+    const premises = await premisesClient.getAllPremises()
+
+    return premises
+      .map(p => {
+        return { text: `${p.name}`, value: `${p.name}` }
+      })
+      .sort((a, b) => {
+        if (a.text < b.text) {
+          return -1
+        }
+        if (a.text > b.text) {
+          return 1
+        }
+        return 0
+      })
+  }
+
   async summaryListForPremises(premises: Premises): Promise<SummaryList> {
     return {
       rows: [
--- a/server/views/departures/new.njk
+++ b/server/views/departures/new.njk
@@ -1,0 +1,185 @@
+{% from "govuk/components/date-input/macro.njk" import govukDateInput %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
+{% from "govuk/components/textarea/macro.njk" import govukTextarea %}
+{% from "govuk/components/radios/macro.njk" import govukRadios %}
+{% from "govuk/components/select/macro.njk" import govukSelect %}
+{% from "govuk/components/summary-list/macro.njk" import govukSummaryList %}
+
+{% extends "../partials/layout.njk" %}
+
+{% set pageTitle = applicationName + " - Home" %}
+{% set mainClasses = "app-container govuk-body" %}
+
+{% block content %}
+
+    <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds">
+            <h1>Log a departure</h1>
+            {{ govukSummaryList({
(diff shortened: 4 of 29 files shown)


server/services/premisesService.test.ts
```diff
@@ -75,6 +75,20 @@ describe('PremisesService', () => {
     })
   })
 
+  describe('getPremisesRadioList', () => {
```
pezholio, line 78: Did this sneak in from another commit?

server/services/premisesService.ts
```diff
@@ -34,6 +34,15 @@ export default class PremisesService {
     return { name: premises.name, summaryList }
   }
 
+  async getPremisesRadioList(): Promise<Array<{ text: string; value: string }>> {
```
pezholio, line 37: I think having these items as a select is fine. As a rule I tend to go for radios for enum-like values, and selects for things that are likely to change, or are fetched from the database.

server/services/premisesService.ts
```diff
@@ -34,6 +34,15 @@ export default class PremisesService {
...
+    const premisesClient = this.premisesClientFactory(this.token)
+    const premises = await premisesClient.getAllPremises()
+
+    return premises.map(p => {
```
pezholio, line 41: Could we order these alphabetically?

server/views/departures/new.njk
```diff
@@ -0,0 +1,160 @@
...
+                    },
+                    hint: {
+                        text: "For example, 27 3 2007"
+                    }
+                }) }}
+
+                <label class="govuk-label govuk-label--m" for="time">
+                    What is the time of departure?
+                </label>
+                <input type="time" id="dateTime-time" name="dateTime-time" class="govuk-input govuk-input--width-5 govuk-!-margin-bottom-4">
```
pezholio, line 36: I think this is fine for now - would be good to add this as a macro, so if we do end up changing the approach, we only have to do it in one place 👍

pezholio (approved):
Looks ace - thanks so much for doing this! Just a couple of tiny things 👍

server/services/bookingService.ts
```diff
@@ -101,7 +101,7 @@ export default class BookingService {
         text: convertDateString(booking.expectedDepartureDate).toLocaleDateString('en-GB'),
       },
       {
-        html: `<a href="/premises/${premisesId}/bookings/${booking.id}/">
+        html: `<a href="/premises/${premisesId}/bookings/${booking.id}/departures/new">
```
pezholio, line 104: Not for now, but with this in mind, we should probably update the `Manage` links in the upcoming departures and departing today tabs to link to the departures page

server/views/departures/new.njk
```diff
@@ -0,0 +1,160 @@
...
+{% extends "../partials/layout.njk" %}
+
+{% set pageTitle = applicationName + " - Home" %}
+{% set mainClasses = "app-container govuk-body" %}
+
+{% block content %}
+
+    <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds">
+            <h1>Log a departure</h1>
```
pezholio, line 15: I do wonder if it's worth playing back the details of the user we're logging the departure for here (via a definition list or similar)

server/views/departures/new.njk
richpjames: Yeah I think it makes sense to do that next time we use a time input

server/services/premisesService.ts
richpjames: Sure, I think ideally later down the line they would be sorted by region

server/services/premisesService.ts
richpjames: Yeah that makes sense. 
The `reason`, `moveOnCategory` and `destinationProvider` are going to be fetch from the DB soon too, should they be selects? 
I think it would make sense if the `destinationProvider` was but I see the other two as enums

server/views/departures/new.njk
richpjames: Yeah I think that is a good idea

server/services/premisesService.ts
pezholio: Yeah, they're all enum-like I think. There's definitely a more clear-cut way of seperating enum-like and database-y, but I'm not sure how to express it 😆

pezholio: I think it's maybe if there's more than 10 options, go with a select, otherwise stick with a radio. I'll see if there's some better guidance.

pezholio: Another thing to possibly think about with this, do we maybe want to look at using https://alphagov.github.io/accessible-autocomplete/?

@beth-dixxon Would love to know your take on it 👍

server/services/premisesService.ts
pezholio: 👍

server/views/departures/new.njk
pezholio: Yeah, probably a bit premature for now

server/services/premisesService.ts
richpjames: See I had in my head the opposite was true, that users struggle with selects when there are too many options. 
It's worth bearing in mind that (judging from the pre-prod data) that there will be quite a lot of options: 

  - Move on Category: 12, 
  - Reason: 15,
  - Destination Provider: 20+ 

I don't think it's something we need to decide now but I think it's something worth thinking about. Ideally we'd narrow down the options in so we don't have present such a long list but as we're replacing an existing form we don't really have that option. Agree that Beth's opinion would be super helpful here

beth-dixxon: I think selects can be difficult for users to navigate through if there’s a lot of options, so if there were 20+ it feels like the Autocomplete would be the better option. 

The only concern would be a scenario where users weren't sure of the options included, however with these coming directly from Delius it feels like we could use this option as users should be familiar with the options in the list. We could also use guidance to support this. 

Could we use the autocomplete component to show all of the values (as a dropdown) which will allow users to both browse through the list and type ahead if they wanted to?

server/services/premisesService.ts
```diff
@@ -34,6 +34,25 @@ export default class PremisesService {
...
+      })
+      .sort((a, b) => {
+        if (a.text < b.text) {
+          return -1
+        }
+        if (a.text > b.text) {
+          return 1
+        }
+        return 0
+      })
```
pezholio, line 53: I hate that we have to do this to sort alphabetically in JS / TS

pezholio (approved):
LGTM - nice work! 👍

server/services/premisesService.ts
richpjames: Yeah it's mind boggling

PR #53, opened 2022-08-03
(496) Handling validation errors

pezholio:
This adds the ability to cope with errors that come back from the API, we expect a `400` response, which raises an error with the body of the response. We can then rescue from the error, turn the response into something we can use to show the errors to the user, and then re-render the pages with the errrors presented.

## Screenshots

image

image

image

### DIFF
--- a/server/utils/formUtils.ts
+++ b/server/utils/formUtils.ts
@@ -1,0 +1,26 @@
+import type { ErrorMessages } from 'approved-premises'
+
+export default function dateFieldValues(
+  fieldName: string,
+  context: Record<string, unknown>,
+  errors: ErrorMessages = {},
+) {
+  const errorClass = errors[fieldName] ? 'govuk-input--error' : ''
+  return [
+    {
+      classes: `govuk-input--width-2 ${errorClass}`,
+      name: 'day',
+      value: context[`${fieldName}-day`],
+    },
+    {
+      classes: `govuk-input--width-2 ${errorClass}`,
+      name: 'month',
+      value: context[`${fieldName}-month`],
+    },
+    {
+      classes: `govuk-input--width-4 ${errorClass}`,
+      name: 'year',
+      value: context[`${fieldName}-year`],
+    },
+  ]
+}
--- a/server/views/premises/bookings/new.njk
+++ b/server/views/premises/bookings/new.njk
@@ -3,6 +3,7 @@
 {% from "govuk/components/input/macro.njk" import govukInput %}
 {% from "govuk/components/select/macro.njk" import govukSelect %}
 {% from "govuk/components/button/macro.njk" import govukButton %}
+{% from "../../partials/showErrorSummary.njk" import showErrorSummary %}
 
 {% extends "../../partials/layout.njk" %}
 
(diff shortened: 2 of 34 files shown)


pezholio:
[Handling validation errors](https://trello.com/c/ESogLbGq/496-handling-validation-errors)

server/utils/formUtils.ts
```diff
@@ -0,0 +1,25 @@
+import type { ErrorMessages } from 'approved-premises'
+
+export default function dateFieldValues(
+  fieldName: string,
+  context: Record<string, unknown>,
+  errors: ErrorMessages = {},
+) {
+  return [
```
richpjames, line 8: Feel free to ignore as it's minor but I think it could be clearer to define `errors[fieldName] ? 'govuk-input--error' : ''` as a variable (EG `errorClass`) and use that on lines 10, 15 and 20 instead of duplicating the logic

wiremock/bookingStubs.ts
```diff
@@ -0,0 +1,91 @@
+import { stubFor } from './index'
+import bookingDtoFactory from '../server/testutils/factories/bookingDto'
+
+const getCombinations = (arr: Array<string>) => {
+  const result: Array<Array<string>> = []
+  const f = (prefixes: Array<string>, suffixes: Array<string>) => {
```
richpjames: It might be overkill but I would like to see some unit tests on this. I trust it works as expected but would be good for documenation purposes

wiremock/stubApis.ts
```diff
@@ -2,11 +2,12 @@
 import { stubFor } from './index'
 
 import premises from './stubs/premises.json'
-import bookingDtoFactory from '../server/testutils/factories/bookingDto'
 import bookingFactory from '../server/testutils/factories/booking'
 import arrivalFactory from '../server/testutils/factories/arrival'
 import nonArrivalFactory from '../server/testutils/factories/nonArrival'
 
+import bookingStubs from './bookingStubs'
```
richpjames, line 7: Good to see, this file is getting big!

server/views/premises/bookings/new.njk
```diff
@@ -11,33 +12,38 @@
 
 {% block content %}
 
-    <h1>Make a booking</h1>
-    <div class="govuk-grid-row">
-        <div class="govuk-grid-column-two-thirds">
-            <form action="/premises/{{premisesId}}/bookings" method="post">
-                <input type="hidden" name="_csrf" value="{{ csrfToken }}"/>
+	<h1>Make a booking</h1>
```
richpjames, line 15: I think we have different formatters set for `.njk` files, I don't mind which one we used but would be align. 
I use "okitavera.vscode-nunjucks-formatter"

richpjames (approved):
Looks great! really like the approach. 
I've only one request for some unit tests but other than that this is awesome 🙏

server/views/premises/bookings/new.njk
pezholio: Cool - I'll stick with yours - I was using a Twig formatter that wasn't great, but the one you use looks much better!

wiremock/bookingStubs.ts
pezholio: Nice - I got lazy because it was for internal purposes, but writing a test would've helped me a lot when figuring out how to do this!

PR #55, opened 2022-08-04
Stub all APIs at once

pezholio:
Before we were stubbing each individual stub in a seperate request, which Wiremock struggled with sometimes, now we’re using the Bulk Import API (https://wiremock.org/docs/stubbing/#bulk-importing-stubs), which sends them all in one HTTP request, which is much faster and more reliable.

### DIFF
--- a/wiremock/arrivalStubs.ts
+++ b/wiremock/arrivalStubs.ts
@@ -5,6 +5,7 @@ import { getCombinations, errorStub } from './utils'
 const arrivalStubs: Array<Record<string, unknown>> = []
 
 arrivalStubs.push({
+  priority: 99,
   request: {
     method: 'POST',
     urlPathPattern: `/premises/${guidRegex}/bookings/${guidRegex}/arrivals`,
--- a/wiremock/arrivalStubs.ts
+++ b/wiremock/arrivalStubs.ts
@@ -1,28 +1,27 @@
-import { stubFor, guidRegex } from './index'
+import { guidRegex } from './index'
 import arrivalFactory from '../server/testutils/factories/arrival'
 import { getCombinations, errorStub } from './utils'
 
-const arrivalStubs = [
-  async () =>
-    stubFor({
-      request: {
-        method: 'POST',
-        urlPathPattern: `/premises/${guidRegex}/bookings/${guidRegex}/arrivals`,
-      },
-      response: {
-        status: 201,
-        headers: {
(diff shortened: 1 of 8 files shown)


richpjames (dismissed):

pezholio:
So, I've figured out what's going on here - Wiremock has the concept of priorities. It usually prioritises the first match, but we can override this, before it kinda worked by accident, but now the order of the stubs is more predictable, so if we want this to work as we expect, then we need to set priorities manually (See https://wiremock.org/docs/stubbing/#stub-priority). This may just be too much work for very little reward though, and we may not want to stub the errors at all in dev.

richpjames:
> So, I've figured out what's going on here - Wiremock has the concept of priorities. It usually prioritises the first match, but we can override this, before it kinda worked by accident, but now the order of the stubs is more predictable, so if we want this to work as we expect, then we need to set priorities manually (See https://wiremock.org/docs/stubbing/#stub-priority). This may just be too much work for very little reward though, and we may not want to stub the errors at all in dev.

Are the 'overlapping' requests we'll have the errors and non-error responses? 
I don't think we really want to set a priority, we just want the the closest match. 
But if we need to set a priority (and I agree that we don't we want to spend much more time on this) would it be possible to bulk load the 2xx responses first and then do the 4xx ones?

pezholio:
> Are the 'overlapping' requests we'll have the errors and non-error responses?
I don't think we really want to set a priority, we just want the the closest match.
But if we need to set a priority (and I agree that we don't we want to spend much more time on this) would it be possible to bulk load the 2xx responses first and then do the 4xx ones?

The issue wasn't really that the stub wasn't working, it was much (MUCH) faster and more reliable, but for reasons of Javascript, we were getting the errors being POSTed first, followed by the happy path, now it's just an array, the happy path was getting loaded first, followed by the unhappy path. The way Wiremock works is that it looks through the mocks and finds the first one that vaguely matches that, which is now the happy path. I've got around this by nailing in a `99` priority for the happy path, and then ordering the fields by the biggest amount of missing fields first (so the "unhappiest" path first) and then working down the tree from there. 

It seems to largely work now, but not worth spending much more time on, as this is only a development thing to make the application behave in a more "realistic" way.

andrewrlee (dismissed):
LGTM

andrewrlee (approved):
LGTM

PR #58, opened 2022-08-05
(552) Departure error handling

richpjames:
# Context
We need to be able to show errors when the form fields are submitted without the data we need on the departures form.

[Trello ticket](https://trello.com/c/9UXVb7jA/552-departure-error-handling)

# Changes in this PR
This PR closely follows the patterns outlined in @pezholio's [earlier error handling PR](https://github.com/ministryofjustice/approved-premises-ui/pull/53). 
The only issue with this I see is that the time input doesn't have the same errors showing despite being a necessary field. We're using nunjucks macros from the GDS DS for all the inputs apart from time and I ran out of time to implement something similar for it. But this is a start.

## Screenshots of UI changes
image

### DIFF
--- a/integration_tests/e2e/departure.cy.ts
+++ b/integration_tests/e2e/departure.cy.ts
@@ -55,4 +55,36 @@ context('Departures', () => {
 
     departureConfirmationPage.verifyConfirmedDepartureIsVisible(departure, booking)
   })
+  it('should show errors', () => {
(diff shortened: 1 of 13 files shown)


pezholio (approved):
LGTM - just had to squash two commits, as you're off @richpjames. Hopefully I can still approve / merge this, even though I've pushed to it

PR #60, opened 2022-08-08
Add the path to the `npm test` args

pezholio:
For some reason, the unit tests are hanging when running `npm test`. This seems to fix it. A bigger question is whether we need to be running tests on two platforms. I tend to prefer Github Actions, as it keeps the CI closer to the code, but having the deploy secrets in Circle seems to make sense. I’ll keep both running for now, but may drop Github Actions in the near future.


andrewrlee (approved):
LGTM

pezholio:
Thanks @andrewrlee 👍

PR #70, opened 2022-08-10
(579) Get reference data from the api and display in forms

pezholio:
To keep things consistent, we need to be able to get a list of all reference fields from the API, and present them to the user as radio buttons. This then allows us to know their IDs and also keep the naming consistent across both applications. This adds a reference data client to fetch different types of reference data, adds a service method to get the relevant reference data for a departure, and turns this reference data into a form that will work with nunjucks macros.

## Screenshot

image

image


pezholio:
[Get reference data from the API and display in forms](https://trello.com/c/lcu92Xuf/579-get-reference-data-from-the-api-and-display-in-forms)

server/utils/nunjucksSetup.ts
```diff
@@ -50,4 +50,12 @@ export default function nunjucksSetup(app: express.Express, path: pathModule.Pla
   njkEnv.addGlobal('dateFieldValues', function (fieldName: string, errors: ErrorMessages) {
     return dateFieldValues(fieldName, this.ctx, errors)
   })
+
+  njkEnv.addGlobal(
+    'convertObjectsToRadioItems',
+    // eslint-disable-next-line func-names
+    function (items: Array<Record<string, string>>, textKey: string, valueKey: string, fieldName: string) {
```
andrewrlee, line 57: meh.. wondering if avoiding the function name could cause issues with debugging

integration_tests/mockApis/departure.ts
```diff
@@ -1,9 +1,10 @@
-import { SuperAgentRequest } from 'superagent'
+import { SuperAgentRequest, Response } from 'superagent'
```
andrewrlee, line 1: super meh, could choose to import these by type

server/utils/nunjucksSetup.ts
pezholio: Yeah, this is part of some strangeness around adding nunjucks globals - I can't get access to `this` without using a `function`. If there's a better way, I'd love to know, because I'm not 100% happy putting `eslint-disable`s everywhere

server/services/departureService.ts
```diff
@@ -29,4 +34,13 @@ export default class DepartureService {
 
     return { ...departure, dateTime: parseISO(departure.dateTime).toLocaleDateString('en-GB') }
   }
+
+  async getReferenceData(): Promise<DepartureReferenceData> {
+    const referenceDataClient = this.referenceDataClientFactory(this.token)
+    return {
+      departureReasons: await referenceDataClient.getReferenceData('departure-reasons'),
+      moveOnCategories: await referenceDataClient.getReferenceData('move-on-categories'),
+      destinationProviders: await referenceDataClient.getReferenceData('destination-providers'),
```
andrewrlee, line 43: Guess there's an opportunity to load these in parallel with `Promise.all`

server/utils/nunjucksSetup.ts
pezholio: Ah, course. I could just do `function sendContextConvertObjectsToRadioItems(....` here

integration_tests/mockApis/departure.ts
pezholio: Good catch, ta!

server/services/departureService.ts
pezholio: Yeah, nice. I'll do that 👍

server/@types/approved-premises/index.d.ts
```diff
@@ -20,7 +21,16 @@ declare module 'approved-premises' {
     nonArrival: Omit<NonArrival, 'id' | 'bookingId'>
   }
 
-  export type DepartureDto = ObjectWithDateParts<'dateTime'> & { departure: Omit<Departure, 'id' | 'bookingId'> }
+  export type DepartureDto = Omit<
+    Departure,
+    'id' | 'bookingId' | 'reason' | 'moveOnCategory' | 'destinationProvider' | 'destinationAp'
+  > &
+    ObjectWithDateParts<'dateTime'> & {
```
psoleckimoj, line 28: I'm not clear what `ObjectWithDateParts` is defining even after looking at it's definition?

server/controllers/departuresController.test.ts
```diff
@@ -1,7 +1,7 @@
 import type { Request, Response, NextFunction } from 'express'
 import { createMock, DeepMocked } from '@golevelup/ts-jest'
```
psoleckimoj, line 2: Just noticed this and another module (fakerjs) are in your normal dependency tree, shouldn't they be in devDependencies? Was just having a nosey at what this module did 👍

server/controllers/departuresController.ts
```diff
@@ -19,30 +19,31 @@ export default class DeparturesController {
...
-      }
+      } as DepartureDto
 
       try {
         const { id } = await this.departureService.createDeparture(premisesId, bookingId, departure)
         res.redirect(`/premises/${premisesId}/bookings/${bookingId}/departures/${id}/confirmation`)
       } catch (err) {
         const booking = await this.bookingService.getBooking(premisesId, bookingId)
         const premisesSelectList = await this.premisesService.getPremisesSelectList()
+        const referenceData = await this.departureService.getReferenceData()
```
psoleckimoj, line 44: Is there not a chance that any of these 3 `await` statements could fail in the same way as the one in the `try` block?

server/data/departureClient.ts
```diff
@@ -9,11 +9,7 @@ export default class DepartureClient {
     this.restClient = new RestClient('departureClient', config.apis.approvedPremises as ApiConfig, token)
   }
 
-  async create(
-    premisesId: string,
-    bookingId: string,
-    departure: Omit<Departure, 'id' | 'bookingId'>,
-  ): Promise<Departure> {
+  async create(premisesId: string, bookingId: string, departure: DepartureDto): Promise<Departure> {
```
psoleckimoj, line 12: Confused as to why a departure is of type `DepartureDto` and not `Departure` but returns a `Departure` from the function?

server/services/departureService.ts
```diff
@@ -29,4 +34,18 @@ export default class DepartureService {
 
     return { ...departure, dateTime: parseISO(departure.dateTime).toLocaleDateString('en-GB') }
   }
+
+  async getReferenceData(): Promise<DepartureReferenceData> {
+    const referenceDataClient = this.referenceDataClientFactory(this.token)
+
+    return Promise.all([
```
psoleckimoj, line 41: This would be easier to read if you used `await`

server/services/premisesService.ts
```diff
@@ -40,7 +40,7 @@ export default class PremisesService {
 
     return premises
       .map(p => {
```
psoleckimoj, line 42: Being pickie but `p` doesn't mean anything as a variable, maybe call it `singlePremises` or something?

server/controllers/departuresController.ts
andrewrlee: We tend to use store errors and input in flash and redirect back to the GET - that has the benefit of meaning you don't have to maintain all the code required to render a page in two places.

andrewrlee: maybe `renderWithErrors` would be better named `renderAnyValidationErrorsOrPropagate` or something - something to hint it will only render errors it knows what to do with

andrewrlee: I think with this pattern, it means if you get a one off error or timeout, to render the standard error page, it would have to do these 3 calls, wait for a response, throw the responses away and then let the standard error handler do its thing.

server/controllers/departuresController.test.ts
pezholio: Good catch! Dunno how they snuck in there

server/controllers/departuresController.ts
pezholio: Oh that's interesting. I'll give that a try. Mind if I do that in a seperate PR to stop this getting bloated? That way I can change the pattern everywhere

server/data/departureClient.ts
pezholio: So, `DepartureDto` is a `Departure` without the ID and BookingID, and with `moveOnCategory`,  `destinationProvider` and `destinationAp` as UUID strings. When we POST to the API, the Departure is created and returned as a "proper" `Departure` object with the ID and BookingID, as well as the `moveOnCategory`,  `destinationProvider` and `destinationAp` as objects.

server/@types/approved-premises/index.d.ts
pezholio: I'll add some comments to it. It's basically defining the shape of an object that has date parts coming from the GDS Date Input (so `${date}-day`, `${date}-month`, ${date}-year`)  I'm not convinced how useful it in in DTOs actually, as we should have converted them by the point that the request body gets converted to a Dto object. I'll tidy this up in a future PR.

server/services/departureService.ts
pezholio: I'm not sure I follow - sorry!

pezholio: So literally just?:

```
return {
      departureReasons: await referenceDataClient.getReferenceData('departure-reasons'),
      moveOnCategories: await referenceDataClient.getReferenceData('move-on-categories'),
      destinationProviders: await referenceDataClient.getReferenceData('destination-providers')
}
```

I had this before, but @andrewrlee suggestes using `Promise.all`. Can put it back if you think it's more readable

server/services/premisesService.ts
pezholio: Yeah, pluralisation is hard 😊

server/services/departureService.ts
psoleckimoj: Individual preferences 😁 I find await more readable.

server/controllers/departuresController.ts
pezholio: What happens with the user's previous answers? Do you store the body in the flash too?

pezholio: Ah, have seen what happens elsewhere. Yep. Agreed, this is a much better pattern

psoleckimoj (approved):
I"m going to LGTM as it looks mostly OK, although for me it's too large to process as it's a project I don't know, so YMMV as to whether it works 100%.

pezholio:
Cool, thanks. I'll make sure any future PRs are easier to digest 👍

andrewrlee (approved):
LGTM

andrewrlee (approved):
LGTM

PR #74, opened 2022-08-10
Redirect `/` to the `/premises` page

pezholio:
Now we actually have a service, we don't need the "Under Construction" page. This will more than likely become a Start Page or similat further down the line, but this will do for now.


Thource (approved):

pezholio:
Thanks @Thource 👍

PR #77, opened 2022-08-11
Redirect with errors on validation error in arrivals

pezholio:
This alters the approach for error handling to redirect to the `GET` method and render errors when an error is detected. This means that we don't have to repeat ourselves when doing any setup/teardown and reduce any risk of any expensive setup operations being wasted if the error is not a validation error and ends up getting bubbled up. In the interests of keeping this PR focussed, I've used this approach in the arrivalsController for now, but if all is well, I'll use it in the other controllers too and add that to another PR.

### DIFF
--- a/server/controllers/arrivalsController.test.ts
+++ b/server/controllers/arrivalsController.test.ts
@@ -3,9 +3,9 @@ import { createMock, DeepMocked } from '@golevelup/ts-jest'
 
 import ArrivalService from '../services/arrivalService'
 import ArrivalsController from './arrivalsController'
-import renderWithErrors from '../utils/renderWithErrors'
+import renderWithErrors from '../utils/validation'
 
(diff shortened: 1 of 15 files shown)


server/controllers/arrivalsController.test.ts
```diff
@@ -28,10 +29,41 @@ describe('ArrivalsController', () => {
         bookingId: 'bookingId',
         premisesId: 'premisesId',
       }
+      ;(fetchErrorsAndUserInput as jest.Mock).mockImplementation(() => {
```
andrewrlee, line 32: meh don't need to mockImplementation if just returning a value can just `mockReturn`

andrewrlee (approved):
LGTM

brownbearsoftware (approved):

PR #78, opened 2022-08-11
Redirect with errors on validation error

pezholio:
This builds on #77 to apply the new redirection approach in all endpoints that create an entity


andrewrlee (approved):
LGTM

PR #85, opened 2022-08-15
Refactor routes

pezholio:
As we're making heavy use of controllers and Express Request handlers, rather than directly using routes, I've updated the application to add a factory that initializes the controllers, together with the services they rely on in `controllers/index.ts`. The initialized controllers are then injected into the application on boot, and matched against routes in the `routes/index.ts` file. 

This means we're defining and initializing controllers in one place, and also sets us up for further optimisation, where we can add our routes in a config file, which will allow us to use route helpers in our views and controllers, so we don't have to repeat ourselves when setting redirects and outputting paths in views. This can be done later though.


pezholio:
[Refactor Routes](https://trello.com/c/oW7uld95/605-refactor-routes)

Thource (approved):

pezholio:
Thanks @Thource! 👍

PR #96, opened 2022-08-18
(614) Smarten up our token sharing

pezholio:
Previously, we were just passing a stub `TOKEN` variable to the stub APIs, now we're starting to think about integrating with the API, we need to pass actual tokens from HMPPS Auth. I've done this by adding a `withTokenFromRequest` method to all of our services, this accepts an Express Request object, extracts the token from the session, passes it to the initialized service, then returns the object (to allow for method chaining). This ensures we can seperate the concerns of our service methods and ensure we don't have to carry out repetitive actions in our controllers.

### DIFF
--- a/server/services/premisesService.ts
+++ b/server/services/premisesService.ts
@@ -1,12 +1,11 @@
 import type { Premises, TableRow, SummaryList } from 'approved-premises'
 import type { RestClientBuilder, PremisesClient } from '../data'
+import Service from './service'
 
-export default class PremisesService {
-  // TODO: We need to do some more work on authentication to work
-  // out how to get this token, so let's stub for now
-  token = 'FAKE_TOKEN'
-
-  constructor(private readonly premisesClientFactory: RestClientBuilder<PremisesClient>) {}
+export default class PremisesService extends Service {
+  constructor(private readonly premisesClientFactory: RestClientBuilder<PremisesClient>) {
+    super()
+  }
 
   async tableRows(): Promise<Array<TableRow>> {
     const premisesClient = this.premisesClientFactory(this.token)
--- a/server/services/service.ts
+++ b/server/services/service.ts
@@ -1,0 +1,10 @@
+import type { Request } from 'express'
+
+export default class Service {
+  token: string
+
+  withTokenFromRequest(request: Request): typeof this {
+    this.token = request.user.token
+    return this
+  }
+}
--- a/server/controllers/arrivalsController.test.ts
+++ b/server/controllers/arrivalsController.test.ts
@@ -5,6 +5,7 @@ import type { ErrorsAndUserInput } from 'approved-premises'
 import ArrivalService from '../services/arrivalService'
 import ArrivalsController from './arrivalsController'
 import { catchValidationErrorOrPropogate, fetchErrorsAndUserInput } from '../utils/validation'
+import servicesShouldGetTokenFromRequest from './shared_examples'
 
 jest.mock('../utils/validation')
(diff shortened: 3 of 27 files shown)


Thource (approved):
Nice!

server/services/service.ts
```diff
@@ -0,0 +1,10 @@
+import type { Request } from 'express'
+
+export default class Service {
+  token: string
```
andrewrlee, line 4: I don't think this is thread safe. 

Services are singletons so can't store mutable request scoped data.

Concurrent requests from multiple users may modify this between calling `withTokenFromRequest ` and actually using the token, meaning one user's token can be used to make API calls on behalf of another.

server/services/service.ts
```diff
@@ -0,0 +1,10 @@
+import type { Request } from 'express'
```
andrewrlee, line 1: meh.. generally lower levels of your application shouldn't know about details of higher levels.

The service shouldn't necessarily know that it is living in an express app as dependencies only flow in one direction.

server/services/service.ts
```diff
@@ -0,0 +1,10 @@
...
+
+export default class Service {
+  token: string
+
+  withTokenFromRequest(request: Request): typeof this {
```
andrewrlee, line 6: It's very easy to forget to call this method - and any place that forgets will default to using whatever token was stored by a user from the last request.  

One of the benefits of using typed languages is that you can use the compiler to make illegal states unrepresentable (or at least strongly encourage other developers to go down the right path!).

One way of doing this would be by using builder or factory, so services would no longer be singletons.
You could use the compiler to ensure that a valid service can only be instantiated with a token.

On the other hand, it would be much simpler if the token was considered as a required value and passed from the controller to the service as an additional parameter. This would reduce the complexity in your controllers and their tests - meaning you don't need to have special mocking code.

andrewrlee: For what it's worth - we did look at using continuation local storage (e.g [cls-hooked](https://github.com/Jeff-Lewis/cls-hooked)) and using as an equivalent to thread-local storage.

We'd extract tokens in some middleware and make them accessible directly in the rest clients, to avoid having to pass tokens about but it seemed a bit flaky and other libraries that also used CLS used to trounce over them.

server/services/premisesService.ts
```diff
@@ -1,12 +1,11 @@
...
 import type { RestClientBuilder, PremisesClient } from '../data'
+import Service from './service'
 
-export default class PremisesService {
-  // TODO: We need to do some more work on authentication to work
-  // out how to get this token, so let's stub for now
-  token = 'FAKE_TOKEN'
-
-  constructor(private readonly premisesClientFactory: RestClientBuilder<PremisesClient>) {}
+export default class PremisesService extends Service {
```
andrewrlee, line 5: meh.. I'd personally tend to avoid using [inheritance solely for code sharing](https://www.infoworld.com/article/2073649/why-extends-is-evil.html)

server/services/service.ts
pezholio: Yeah, you're quite right. I didn't really think about services behaving like a singleton across the whole lifecycle of the application. Back to the drawing board it is!

server/services/service.ts
pezholio: Yeah, I get where you're coming from with this, it seemed a bit smelly when I was writing it, but I am very much a fan of DRY code, but typing `req.locals.token` everywhere is probably preferable to including external dependencies.

PR #103, opened 2022-08-22
Fix thread safety with tokens

pezholio:
Although services are classes, they behave like singletons, as they’re initialized once when the application starts. Passing the token around as we had before is not necessarily thread safe, and there’s a risk that users could act as another user, so it’s simpler and safer to pass the token into the methods directly.


andrewrlee (approved):
LGTM

richpjames (approved):

PR #115, opened 2022-08-26
(537) Extension of bookings

richpjames:
# Context
AP Managers need to be able to extend a booking if neccessary. 
https://trello.com/c/M2MH0WYC/537-extensions-of-bookings

# Changes in this PR
From the booking details screen I have added an 'action' that allows the users to extend a booking by entering a new departure date. 
To post this data to the backend we use the BookingClient and BookingService but I have created a new BookingExtensionsController.
Once the extension has been posted successfully the user is presented with a confirmation screen which lets them know the action has been successful. 

### After
image
image
image

### DIFF
--- a/integration_tests/e2e/bookingExtension.cy.ts
+++ b/integration_tests/e2e/bookingExtension.cy.ts
@@ -1,0 +1,74 @@
+import premisesFactory from '../../server/testutils/factories/premises'
+import bookingFactory from '../../server/testutils/factories/booking'
+
+import BookingExtensionsConfirmation from '../pages/booking/extension/confirmation'
+import BookingExtensionCreatePage from '../pages/booking/extension/create'
+
+context('BookingExtension', () => {
+  beforeEach(() => {
+    cy.task('reset')
+    cy.task('stubSignIn')
+    cy.task('stubAuthUser')
+  })
+
+  it('should show booking extension form', () => {
+    const booking = bookingFactory.build({
+      expectedDepartureDate: new Date(Date.UTC(2022, 5, 3, 0, 0, 0)).toISOString(),
+    })
+    const newDepartureDate = new Date(Date.UTC(2022, 6, 3, 0, 0, 0)).toISOString()
+    const premises = premisesFactory.build()
+
+    cy.task('stubBookingExtensionCreate', { premisesId: premises.id, booking })
+    cy.task('stubBookingGet', { premisesId: premises.id, booking })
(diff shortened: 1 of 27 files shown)


server/@types/approved-premises/index.d.ts
```diff
@@ -37,6 +37,8 @@ declare module 'approved-premises' {
       destinationAp: string
     }
 
+  export type NewDepartureDate = ObjectWithDateParts<'newDepartureDate'>
```
pezholio, line 40: I think we should have notes here too

server/services/bookingService.ts
```diff
@@ -49,6 +49,19 @@ export default class BookingService {
     }
   }
 
+  async extendBooking(
+    token: string,
+    premisesId: string,
+    bookingId: string,
+    newDepartureDate: ObjectWithDateParts<'newDepartureDate'>,
```
pezholio, line 56: Should this just be a `NewDepartureDate` type?

integration_tests/pages/booking/confirmation.ts
```diff
@@ -1,7 +1,7 @@
 import parseISO from 'date-fns/parseISO'
 import type { Booking } from 'approved-premises'
-import { formatDate } from '../../server/utils/utils'
-import Page from './page'
+import { formatDate } from '../../../server/utils/utils'
```
pezholio, line 3: Good call 👍

integration_tests/e2e/bookingExtension.cy.ts
```diff
@@ -0,0 +1,75 @@
...
+    cy.signIn()
+
+    // When I visit the booking extension page
+    const page = BookingExtensionCreatePage.visit(premises.id, booking.id)
+
+    // And I fill in the extension form
+    page.completeForm(newDepartureDate)
+    page.clickSubmit()
+
+    // Then I should be redirected to the confirmation page and the extension should be created in the API
```
pezholio, line 36: I prefer each step to be expressed as a different comment before a code block, so this should be:

```
// Then I should be redirected to the confirmation page
```

And line 40 should be:

```
// And the extension should be created in the API
```

pezholio (commented):
Looks largely good - just worried about the absence of a notes field. Can we get that added pls?

server/@types/approved-premises/index.d.ts
richpjames: Yep! my bad

pezholio (approved):
LGTM 👍

PR #118, opened 2022-08-26
Success messages for arrivals and non arrivals

richpjames:
We currently don't give the user any feedback when they sucessfully mark someone as having arrived or non-arrived which could create confusion.
A confirmation screen seemed like a step too far for a small action so we just show a flash message on the Premises detail page for now.

image
image

### DIFF
--- a/integration_tests/e2e/arrivals.cy.ts
+++ b/integration_tests/e2e/arrivals.cy.ts
@@ -44,7 +44,8 @@ context('Arrivals', () => {
     })
 
     // And I should be redirected to the premises page
-    PremisesShowPage.verifyOnPage(PremisesShowPage, premises)
+    const premisesPage = PremisesShowPage.verifyOnPage(PremisesShowPage, premises)
+    premisesPage.shouldShowBanner('Arrival logged')
   })
 
   it('show arrival errors when the API returns an error', () => {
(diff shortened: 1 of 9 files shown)


integration_tests/e2e/arrivals.cy.ts
```diff
@@ -44,7 +44,8 @@ context('Arrivals', () => {
     })
 
     // And I should be redirected to the premises page
-    PremisesShowPage.verifyOnPage(PremisesShowPage, premises)
+    const premisesPage = PremisesShowPage.verifyOnPage(PremisesShowPage, premises)
+    premisesPage.shouldShowBanner('Arrival logged')
```
pezholio, line 48: Love it!

server/utils/utils.ts
```diff
@@ -76,3 +76,7 @@ export const convertDateAndTimeInputsToIsoString = <K extends string | number>(
 
   return dateInputObj
 }
+
+export const formatBanner = (copy: string) => {
+  return `<h3 class="govuk-!-margin-top-2">${copy}</h3>`
```
pezholio, line 81: I'm a bit wary about us including presentation markup in code that gets used in the controller. Maybe a better way would be to add the `<h3>` in the call to `govukNotificationBanner` in the `_messages` partial. Maybe something like this?

```
{% if successMessages %}
  {% for message in successMessages %}
    {{ govukNotificationBanner({
      html: '<h3 class="govuk-!-margin-top-2">' + message + '</h3>',
      type: 'success'
    }) }}
  {% endfor %}
{% endif %}
```

That way we don't have to remember to user `formatBanner` everywhere too.

richpjames: Another great idea 💡

pezholio (approved):
LGTM 👍

PR #128, opened 2022-08-30
Rename *Dto types to new*

richpjames:
We have agreed that objects we create before they are sent to the server should be referred to as `new*` as opposed to `*Dto` as this more accurately describes their nature and matches their names in the API and API spec. 

Rather than muddy the waters in a ticket's PR I think it makes sense to isolate this change to its own PR as it touches so many files.


pezholio (approved):
LGTM 👍 Nice one!

PR #135, opened 2022-08-31
Use static paths

pezholio:
This uses the [Static Path](https://github.com/garybernhardt/static-path) library to allow us to define paths in one place. This gives us two benefits:

- We don't have to hardcode paths and change them everywhere if anything changes; and
- The path objects are type-safe, so we will be told if a parameter is missing!

This will also help us in Apply a lot, as we'll be redirecting to paths a lot, and it makes sense to have them defined in one place.


pezholio:
[Use Static Paths to define routes](https://trello.com/c/rUvv19Y5/683-use-static-paths-to-define-routes)

richpjames:
this is wonderful 😭 it's going to save us tonnes of time.
any reason for skipping the backlinks in `bookings/show.njk` and `bookings/new.njk`?

richpjames (approved):

pezholio:
Yeah, that was a complete oversight on my part. Will fix 👍🏻

PR #136, opened 2022-09-01
Use identity bar to show actions on premises page

pezholio:
We don’t have a way of creating a new booking at the moment, and this groups the actions more nicely at the top of the page. This may need some design input when @beth-dixxon gets back, but will do for now

# Screenshots

## Before

image

## After

image

### DIFF
--- a/assets/sass/application.sass
+++ b/assets/sass/application.sass
@@ -6,3 +6,5 @@ $path: "/assets/images/"
 
 @import './components/header-bar'
 @import './local'
+
+@import './pages/premises/show'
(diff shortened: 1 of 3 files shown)


richpjames:
I'm worried that users will struggle to find this *but* it's definitely an improvement over what we have at the moment and we can iterate from here

richpjames (approved):

pezholio:
Yeah, it's deffo something we should iterate on. It was just bugging me we had no way to create a new booking without navigating manually! 😆

PR #139, opened 2022-09-02
(598) Update new booking to search by crn

richpjames:
## Context

AP Managers need to be able to create a booking for a person and be assured that the booking has been created for the correct person. Our service also needs to ensure that we know exactly who the person the booking has been created for. 
People in prison/probation are tracked by a UUID - their 'CRN', so we need AP managers to enter this when creating a booking. To ensure they have entered the correct CRN they need to be presented with the name of the person that CRN relates to when creating a booking.

This will neccessitate us calling our API with a CRN which is something we will also need to do for the 'Apply' stage of the Approved Premises service so this feature has been built with this in mind.

[Trello ticket](https://trello.com/c/GeIvVkOn/598-update-create-booking-to-search-by-crn)

## Journey changes in this PR

Making a booking was previously a 2 step process: 
1. The user entered the details of the booking in a form 
2. they were then presented with confirmation of the booking.

Making a booking is now a 3 step process: 
1. The user enters a CRN
2. they are then presented with a page which shows the CRN just entered and the name related to that CRN at the top. The rest of the form is identical to the one from step 1 of the previous process.
3. The user is presented with confirmation of their booking.

## Code Changes 

I have added a new controller - the people controller - to post the CRN to our API. 
The booking controller now has two methods for the form 'find' for searching by CRN and 'new' for the rest of the booking. I'm not entirely set on these names (or any of the naming tbh) so feedback encouraged.

### Before
image

### After
image
image

### DIFF
--- a/server/controllers/manage/bookingsController.ts
+++ b/server/controllers/manage/bookingsController.ts
@@ -1,13 +1,17 @@
 import type { NewBooking } from 'approved-premises'
 import type { Request, Response, RequestHandler } from 'express'
 
-import { BookingService, PremisesService } from '../../services'
+import { BookingService, PremisesService, PersonService } from '../../services'
 import { catchValidationErrorOrPropogate, fetchErrorsAndUserInput } from '../../utils/validation'
 import { convertDateAndTimeInputsToIsoString } from '../../utils/utils'
 import paths from '../../paths/manage'
 
 export default class BookingsController {
-  constructor(private readonly bookingService: BookingService, private readonly premisesService: PremisesService) {}
+  constructor(
+    private readonly bookingService: BookingService,
+    private readonly premisesService: PremisesService,
+    private readonly personService: PersonService,
+  ) {}
 
   show(): RequestHandler {
     return async (req: Request, res: Response) => {
@@ -20,15 +24,29 @@ export default class BookingsController {
   }
 
   new(): RequestHandler {
-    return (req: Request, res: Response) => {
+    return async (req: Request, res: Response) => {
       const { premisesId } = req.params
       const { errors, errorSummary, userInput } = fetchErrorsAndUserInput(req)
 
-      return res.render(`bookings/new`, {
+      const crnArr = req.flash('crn')
+      if (crnArr.length) {
+        const person = await this.personService.findByCrn(req.user.token, crnArr[0])
+
+        return res.render(`bookings/new`, {
+          pageHeading: 'Make a booking',
+          premisesId,
+          ...person,
+          errors,
+          errorSummary,
+          ...userInput,
+        })
+      }
+
+      return res.render(`bookings/find`, {
+        pageHeading: 'Make a booking - find someone by CRN',
         premisesId,
         errors,
         errorSummary,
-        pageHeading: 'Make a booking',
         ...userInput,
       })
     }
@@ -54,14 +72,8 @@ export default class BookingsController {
           }),
         )
       } catch (err) {
-        catchValidationErrorOrPropogate(
-          req,
-          res,
-          err,
-          paths.bookings.new({
-            premisesId,
-          }),
-        )
+        req.flash('crn', booking.crn)
+        catchValidationErrorOrPropogate(req, res, err, paths.bookings.new({ premisesId }))
       }
     }
   }
--- a/server/controllers/peopleController.test.ts
+++ b/server/controllers/peopleController.test.ts
@@ -1,0 +1,79 @@
+import type { Request, Response, NextFunction } from 'express'
+import { createMock, DeepMocked } from '@golevelup/ts-jest'
+
+import PeopleController from './peopleController'
+import PersonService from '../services/personService'
+import personFactory from '../testutils/factories/person'
+import { catchValidationErrorOrPropogate } from '../utils/validation'
+
+jest.mock('../utils/validation')
+
+describe('PeopleController', () => {
+  const token = 'SOME_TOKEN'
+  const premisesId = 'premisesId'
+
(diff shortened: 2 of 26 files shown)


server/data/personClient.ts
```diff
@@ -0,0 +1,20 @@
...
+export default class PersonClient {
+  restClient: RestClient
+
+  constructor(token: string) {
+    this.restClient = new RestClient('personClient', config.apis.approvedPremises as ApiConfig, token)
+  }
+
+  async find(crn: string): Promise<Person> {
+    const response = await this.restClient.post({
+      path: `/person/search`,
```
pezholio, line 14: Yeah, I do think it should be `/people/search` actually in the spirit of RESTfulness. Will feed that back in the other PR.

server/data/personClient.ts
```diff
@@ -0,0 +1,20 @@
...
+import config, { ApiConfig } from '../config'
+
+export default class PersonClient {
+  restClient: RestClient
+
+  constructor(token: string) {
+    this.restClient = new RestClient('personClient', config.apis.approvedPremises as ApiConfig, token)
+  }
+
+  async find(crn: string): Promise<Person> {
```
pezholio, line 12: We should probably call this `search` too

server/services/personService.ts
```diff
@@ -0,0 +1,13 @@
+import type { Person } from 'approved-premises'
+import type { RestClientBuilder, PersonClient } from '../data'
+
+export default class PersonService {
+  constructor(private readonly personClientFactory: RestClientBuilder<PersonClient>) {}
+
+  async find(token: string, crn: string): Promise<Person> {
```
pezholio, line 7: I reckon this should be called `findByCrn` to better describe what it does

server/controllers/peopleController.test.ts
```diff
@@ -0,0 +1,72 @@
...
+          flash: flashSpy,
+        },
+        response,
+        next,
+      )
+
+      expect(response.redirect).toHaveBeenCalledWith('success/')
+      expect(personService.find).toHaveBeenCalledWith(token, person.crn)
+      expect(flashSpy).toHaveBeenCalledWith('crn', person.crn)
+    })
```
pezholio, line 49: Can we stick a newline here for readability?

server/controllers/peopleController.ts
```diff
@@ -0,0 +1,22 @@
...
+  constructor(private readonly personService: PersonService) {}
+
+  find(): RequestHandler {
+    return async (req: Request, res: Response) => {
+      const { crn, successUrl, failureUrl } = req.body
+
+      try {
+        const person = await this.personService.find(req.user.token, crn)
+        req.flash('crn', person.crn)
+        res.redirect(successUrl)
```
pezholio, line 16: I think it might be better to glean the `successUrl` and `failureUrl` from the referrer path, otherwise people could put any URL in the body.

server/controllers/manage/bookingsController.ts
```diff
@@ -19,16 +20,37 @@ export default class BookingsController {
     }
   }
 
-  new(): RequestHandler {
-    return (req: Request, res: Response) => {
+  find(): RequestHandler {
```
pezholio, line 23: Sorrry to be annoying, but I think this needs to be in `new`. We can then define the template based on what is in the flash.

server/views/bookings/find.njk
```diff
@@ -0,0 +1,41 @@
...
+{% extends "../partials/layout.njk" %}
+
+{% set pageTitle = applicationName + " - " + pageHeading  %}
+{% set mainClasses = "app-container govuk-body" %}
+
+{% block content %}
+
+    <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds">
+            <form action="{{ paths.people.find({ premisesId: premisesId }) }}" method="post">
```
pezholio, line 14: I feel like the heading of this page should be `Make a booking` with subheading of `Find a person by CRN` (or whatever)

server/controllers/peopleController.ts
```diff
@@ -0,0 +1,22 @@
...
+  constructor(private readonly personService: PersonService) {}
+
+  find(): RequestHandler {
+    return async (req: Request, res: Response) => {
+      const { crn } = req.body
+
+      try {
+        const person = await this.personService.findByCrn(req.user.token, crn)
+        req.flash('crn', person.crn)
+        res.redirect(req.headers.referer)
```
pezholio, line 16: Oh, this is even better! 🥇

server/controllers/manage/bookingsController.ts
```diff
@@ -20,15 +21,29 @@ export default class BookingsController {
...
 
   new(): RequestHandler {
-    return (req: Request, res: Response) => {
+    return async (req: Request, res: Response) => {
       const { premisesId } = req.params
       const { errors, errorSummary, userInput } = fetchErrorsAndUserInput(req)
 
-      return res.render(`bookings/new`, {
+      const crnArr = req.flash('crn')
+      if (crnArr && crnArr[0]) {
```
pezholio, line 29: Would `if (crnArr) {` be enough here? An empty array is falsy right?

pezholio (approved):
LGTM - nice work! 👍

server/controllers/manage/bookingsController.ts
richpjames: It would make sense if an empty array was falsy... but as an array is an object and an empty object is truthy it is unfortunately truthy.

pezholio: Ah poop. Could we do `if (crnArr?.length)`

richpjames: yes

PR #142, opened 2022-09-05
Initial attempt at application process

pezholio:
This is an initial attempt at the application process, the application service fetches a page class from the params, inside a page class are rules for validation and what page to go to next. Before I take this out of draft, I want to add some more pages, but just pushing up for initial comments

## Still to do:

- [ ] Release Date
- [ ] Oral hearing date
- [ ] Placement start

Once this is in, we can then move on to storing answers in the session, which will alter some of the logic in some of the pages.

To run locally, spin up the server and navigate to:

http://localhost:3000/applications/some-uuid/tasks/basic-information/pages/start

## Screenshots

image

image

image

image

image

image


.eslintrc.json
```diff
@@ -34,6 +34,8 @@
       ],
       "rules": {
         "@typescript-eslint/no-use-before-define": 0,
+        "no-shadow": "off",
+        "@typescript-eslint/no-shadow": "error",
```
richpjames, line 38: I'd be interested to know why we're doing this.. is it this? https://stackoverflow.com/a/64745652

server/services/applicationService.test.ts
```diff
@@ -0,0 +1,67 @@
...
+
+describe('ApplicationService', () => {
+  const service = new ApplicationService()
+
+  beforeEach(() => {
+    jest.resetAllMocks()
+  })
+
+  describe('getCurrentPage', () => {
+    const request = createMock<Request>({ params: { task: 'my-task' } })
```
richpjames, line 31: Maybe a little low level for a draft PR but could we put this in a `beforeEach`?

server/app.ts
```diff
@@ -30,6 +31,9 @@ export default function createApp(controllers: Controllers, services: Services):
   app.set('trust proxy', true)
   app.set('port', process.env.PORT || 3000)
 
+  // Add method-override to allow us to use PUT and DELETE methods
+  app.use(methodOverride('_method'))
```
richpjames, line 35: just for my understanding... why do we need to override PUT and DELETE?

server/@types/approved-premises/index.d.ts
```diff
@@ -127,6 +127,17 @@ declare module 'approved-premises' {
     userInput: Record<string, unknown>
   }
 
+  export interface TasklistPage {
```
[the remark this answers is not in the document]
pezholio: I had some weirdness trying to do this. Are you OK if we park it? We really should look at tidying up these types though!

server/utils/utils.test.ts
```diff
@@ -180,3 +181,41 @@ describe('convertDateString', () => {
     expect(() => convertDateString(date)).toThrow(new InvalidDateStringError(`Invalid Date: ${date}`))
   })
 })
+
+describe('dateAndTimeInputsAreValidDates', () => {
+  it('returns true when the date is valid', () => {
+    interface MyObjectWithADate {
```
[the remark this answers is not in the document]
pezholio: Yeah, exactly that!

.eslintrc.json
pezholio: Yeah, exactly that!

server/app.ts
pezholio: Again, this is all due to me being a REST obsessive. We can't use PUT or DELETE in browsers, so this lets us simulate those requests. This comes for free in Rails, so this is just to keep consistent with that way of doing things

server/utils/utils.test.ts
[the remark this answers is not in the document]
pezholio: I'm not sure what you mean here? Think I'm being a bit of a Typescript noob, or my brain is fried. Probably both!

server/@types/approved-premises/index.d.ts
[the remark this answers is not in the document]
richpjames: Ha, I tried it myself and had some weirdness too... 
Yeah that's fine

pezholio:
[Build out approach for Apply screens](https://trello.com/c/Rz87VFYz/706-build-out-approach-for-apply-screens)

server/services/applicationService.test.ts
richpjames: Something a little like this?
```suggestion
     describe('getCurrentPage', () => {
       let request;
       beforeEach(() => {
            request = createMock<Request>({ params: { task: 'my-task' } })
       })
      it('should return a page from a page list', () => {
        request.params.page = 'second'
   .....etc
```
something like that... So each time we use the request object it has been reset. My concern is that we're mutating the same request object that is used for each of the tests in this describe block. It works fine now but could make later changes painful.

server/app.ts
richpjames: Good to know, thanks!

pezholio:
I am starting to think about the async issue. One idea is possibly having a private constructor and adding an async `initialize` method (similiar to what we did in Alpha). Something like:

```
private constructor(private readonly body: Record<any, any>, private readonly data: Record<any, any> = {}) {}

static async initialize(body: Record<any, any>): Promise<TasklistPage> {
    data = await DoSomeAsyncStuff()

    return new TasklistPage(body, data)
}
```

But this is probably for another PR

server/utils/utils.test.ts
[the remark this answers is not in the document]
richpjames: Ignoring what I originally said I think you can just use `ObjectWithDateParts<'date'>` here?

server/utils/utils.ts
```diff
@@ -77,3 +77,20 @@ export const convertDateAndTimeInputsToIsoString = <K extends string | number>(
...
 }
+
+export const dateAndTimeInputsAreValidDates = <K extends string | number>(
+  dateInputObj: ObjectWithDateParts<K>,
+  key: K,
+): boolean => {
+  const dateString = convertDateAndTimeInputsToIsoString(dateInputObj, key)
+
+  try {
+    convertDateString(dateString[key])
```
richpjames, line 88: Not for now but it would be good to know what validation is currently in place in Delius

server/form-pages/apply/basic-information/placementDate.test.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  describe('errors', () => {
+    it('should return an empty array if the release date is the same as the start date', () => {
+      const page = new PlacementDate({
+        startDateSameAsReleaseDate: 'yes',
+      })
+      expect(page.errors()).toEqual([])
+    })
+
+    it('should return an empty array if the release date is the same as the start date and the date is specified', () => {
```
richpjames, line 17: Should this be '...if the release date is the same as the start date'. I don't think some of the other names below are accurate either. 
It might be easier to parse if we have two seperate `describe` blocks - one for when the release date is the same as the start date and one for when it isn't

server/form-pages/apply/basic-information/placementDate.ts
```diff
@@ -0,0 +1,49 @@
+import type { TasklistPage, ObjectWithDateParts } from 'approved-premises'
+import { dateAndTimeInputsAreValidDates } from '../../../utils/utils'
+
+export default class PlacementDate implements TasklistPage {
+  name = 'placement-date'
+
+  title = 'Is [release-date] the date you want the placement to start?'
```
richpjames, line 7: I guess this will be hooked up later?

richpjames (approved):

server/form-pages/apply/basic-information/placementDate.test.ts
pezholio: Good plan. I've fixed it for all the other date-shaped pages too. Much nicer!

server/form-pages/apply/basic-information/placementDate.ts
pezholio: Yeah, we need access to the session here

server/utils/utils.test.ts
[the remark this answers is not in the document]
pezholio: Yeah, we can do that everywhere in that test file - much neater!

PR #147, opened 2022-09-06
(608) Show capacity for premises

richpjames:
# Context

[Trello card](https://trello.com/c/YsjKdi7A/608-show-capacity-in-the-ui)

# Changes in this PR


server/utils/utils.ts
```diff
@@ -29,6 +29,7 @@ export const initialiseName = (fullName?: string): string | null => {
   return `${array[0][0]}. ${array.reverse()[0]}`
 }
 
+// returns a date like 'Tuesday 6 September 2022'
```
pezholio, line 32: ❤️

server/data/premisesClient.test.ts
```diff
@@ -3,6 +3,7 @@ import nock from 'nock'
 import premisesFactory from '../testutils/factories/premises'
 import PremisesClient from './premisesClient'
 import config from '../config'
+import paths from '../paths/manage'
```
pezholio, line 6: This is an ace idea, but I feel like we should seperate the application paths from the API paths

server/@types/approved-premises/index.d.ts
```diff
@@ -9,6 +9,7 @@ declare module 'approved-premises' {
   export type Cancellation = schemas['Cancellation']
   export type BookingExtension = schemas['BookingExtension']
   export type KeyWorker = schemas['KeyWorker']
+  export type DateCapacity = schemas['DateCapacity']
```
pezholio, line 12: I think this might be better expressed as `PremisesCapacityItem` and then `PremisesCapacity` can be an `Array<PremisesCapacityItem>`

server/services/premisesService.ts
```diff
@@ -30,6 +33,42 @@ export default class PremisesService {
...
+    const premisesClient = this.premisesClientFactory(token)
+    const premisesDateCapacities = await premisesClient.capacity(premisesId)
+    const overcapacityDateRanges = getDateRangesWithNegativeBeds(premisesDateCapacities)
+
+    const overcapacityMessage = this.generateOvercapacityMessage(overcapacityDateRanges)
+    return overcapacityMessage ? [overcapacityMessage] : ''
+  }
+
+  /* istanbul ignore next */
+  private generateOvercapacityMessage(overcapacityDateRanges: NegativeDateRange[]) {
```
pezholio, line 48: Love this! 👍

server/services/premisesService.ts
```diff
@@ -30,6 +33,42 @@ export default class PremisesService {
...
+  async getOvercapacityMessage(token: string, premisesId: string): Promise<string[] | string> {
+    const premisesClient = this.premisesClientFactory(token)
+    const premisesDateCapacities = await premisesClient.capacity(premisesId)
+    const overcapacityDateRanges = getDateRangesWithNegativeBeds(premisesDateCapacities)
+
+    const overcapacityMessage = this.generateOvercapacityMessage(overcapacityDateRanges)
+    return overcapacityMessage ? [overcapacityMessage] : ''
+  }
+
+  /* istanbul ignore next */
```
pezholio, line 47: Why are we ignoring coverage for this?

server/services/premisesService.test.ts
```diff
@@ -142,4 +143,94 @@ describe('PremisesService', () => {
...
+      expect(result).toBe('')
+    })
+
+    it('returns the correct string if passed a single date', async () => {
+      premisesClient.capacity.mockResolvedValue([
+        dateCapacityFactory.build({
+          date: new Date(2022, 0, 1).toISOString(),
+          availableBeds: -1,
+        }),
+      ])
```
pezholio, line 175: I think we should mock the resolved value of `getDateRangesWithNegativeBeds` here too, as that is what we're acting on. Maybe we mock the value of the `capacity` endpoint with stub, mock what we get from `getDateRangesWithNegativeBeds` and then check `getDateRangesWithNegativeBeds` was called with our `capacity` stub. 

Some psuedocode here:

```typescript

const capacityStub = [
  dateCapacityFactory.build(),
]

premisesClient.capacity.mockResolvedValue(capacityStub)

getDateRangesWithNegativeBeds.mockResolvedValue({start: new Date(2022, 0, 1).toISOString()})
```

Then that way we're clearer about what we're expecting back from that method, rather than relying on us passing stuff through the chain.

Does that make sense?

integration_tests/e2e/booking.cy.ts
```diff
@@ -37,10 +68,16 @@ context('Booking', () => {
     page.completeForm(booking)
     page.clickSubmit()
 
-    // Then I should be redirected to the confirmation page and the booking should be created in the API
+    // Then I should be redirected to the confirmation page
+    // And the booking should be created in the API
+    // And I should see the overcapacity message
```
pezholio, line 73: Can we have the comments above the relevant bits of code?

pezholio (approved):
This is absolutely banging, thanks! Just some of my usual anal-retentive naming things, and a suggestion about how to make the tests clearer

server/services/premisesService.ts
richpjames: I thought it was causing the test coverage to fail but it seems to be fine now

server/services/premisesService.test.ts
richpjames: done - although I opted to use the dates from the capacityStub in the return value from getDateRangesWithNegative beds as I think it highlights the link between the two

PR #153, opened 2022-09-09
(709) Save an in progress application in the session

pezholio:
This updates the application process to save answers to the session, as well as showing any previous answers when a user naviagates back to a previously answered question. I've also added some constructors to each Page, to ensure that only parameters that we know about are passed to the Page. This ensures a degree of confidence, without having to be completely type safe.

### DIFF
--- a/server/@types/approved-premises/index.d.ts
+++ b/server/@types/approved-premises/index.d.ts
@@ -142,6 +142,8 @@ declare module 'approved-premises' {
 
   export type TaskListErrors = Array<{ propertyName: string; errorType: string }>
 
+  export type YesOrNo = 'yes' | 'no'
+
   export interface schemas {
     Premises: {
       id: string
--- a/server/@types/approved-premises/index.d.ts
+++ b/server/@types/approved-premises/index.d.ts
@@ -134,6 +134,7 @@ declare module 'approved-premises' {
     name: string
     title: string
 
+    body?: Record<string, unknown>
     previous?(): string
     next?(): string
     errors?(): TaskListErrors
--- a/server/form-pages/apply/basic-information/oralHearing.ts
+++ b/server/form-pages/apply/basic-information/oralHearing.ts
@@ -6,7 +6,7 @@ export default class OralHearing implements TasklistPage {
 
   title = 'Do you know Robert Brown’s oral hearing date?'
 
-  constructor(private readonly body: Record<string, unknown>) {}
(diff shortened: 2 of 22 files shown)


server/@types/approved-premises/index.d.ts
```diff
@@ -134,6 +134,7 @@ declare module 'approved-premises' {
     name: string
     title: string
```
richpjames, line 136: Just a thought - I think an abstract class would more closely match what this interface is representing. 
It might be too much work for now but I think it would save us from having to make changes like this in lots of different locations

server/form-pages/apply/basic-information/oralHearing.ts
```diff
@@ -1,12 +1,23 @@
...
   name = 'oral-hearing'
 
   title = 'Do you know Robert Brown’s oral hearing date?'
 
-  constructor(readonly body: Record<string, unknown>) {}
+  body: ObjectWithDateParts<'oralHearingDate'> & {
+    knowOralHearingDate: YesOrNo
+  }
+
+  constructor(body: Record<string, unknown>) {
```
richpjames, line 13: I don't understand the advantage of typing the properties on this object to be`unknown` if we're then going to immediately say they're strings/YesOrNo. Unless I'm missing something I think it would be more readable to have `Record<string, string | YesOrNo>` here

richpjames (approved):
This is great! 
I feel like it would be a good point to add a really simple E2E test (in a subsequent PR if you feel like). Maybe just entering a CRN and then going back and checking it's still there.

pezholio:
[Save an in-progress application in the session](https://trello.com/c/LZ0UTeak/709-save-an-in-progress-application-in-the-session)

server/form-pages/apply/basic-information/oralHearing.ts
pezholio: I think my thinking was we literally don't know what is being passed into the constructor, so it makes sense to be flexible

richpjames: Yeah after having some time to reflect I think it does make sense to me. 
We're saying 'We don't know what the properties object are but if it has *x* property then we know it's *y* type'

server/@types/approved-premises/index.d.ts
pezholio: Yeah, I think let's leave it like this for now, and see what shakes out

PR #157, opened 2022-09-12
(707) Fetch person detail from Delius community api during the apply process

pezholio:
This adds a new optional async `setup` method to our page classes which calls any external APIs, and also accepts any other service classes. If we get a 404 from the API, we bubble this up to the controller and display an error.

I've also taken a bit of time to tidy up some of the e2e tests - let me know if you'd rather this went in a serparate PR!


pezholio:
[Fetch person detail from Delius community API during the Apply process](https://trello.com/c/c1VTHKOG/707-fetch-person-detail-from-delius-community-api-during-the-apply-process)

server/form-pages/apply/basic-information/confirmDetails.ts
```diff
@@ -1,9 +1,22 @@
...
 
 export default class ConfirmDetails implements TasklistPage {
   name = 'confirm-details'
 
-  title = `Confirm ${this.details().names}'s details`
+  title: string
+
+  details: Person
+
+  async setup(request: Request, dataServices: DataServices): Promise<void> {
```
richpjames, line 14: I think it would make sense to add an optional `setup` method to the TasklistPage interface if multiple pages have them - but that can be done in a future PR as needed

richpjames (approved):
This all looks great. 
A couple of things: 
1. cc1a8ee3cb5eaa682e367dd6af1f1453c6cdb968 is empty 
2. Can we move all the E2E changes to follow one antoher? I agree that it makes sense to namespace them and I don't mind it being included in this PR but currently it happens after 8068817cdb20446aaa5466c2a287ff52ab964cdd which is updating the view, then after they've been namespaced we got into 0aeafd82f1e8e12a05757d1083f2986a5b1efdaf before later in 3fa14efa2e1bb430c578c86d70cf7c15218846fc adding more E2E tests.

server/form-pages/apply/basic-information/confirmDetails.ts
pezholio: Yeah, I think I'll do it in another PR. Might make more sense to make it an abstract class, as you suggested before

PR #158, opened 2022-09-12
Disable typechecking in jest

pezholio:
This enables isolatedModules which has the effect of [disabling typechecking](https://kulshekhar.github.io/ts-jest/docs/getting-started/options/isolatedModules/).

Running this locally, this cuts down the test running time from 58 seconds to 13 seconds!

See https://github.com/ministryofjustice/hmpps-template-typescript/pull/116 for more context

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,11 @@
         "functions": 100
       }
     },
+    "globals": {
+      "ts-jest": {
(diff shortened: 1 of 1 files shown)


richpjames (approved):
Lets go *fast*

PR #169, opened 2022-09-14
(721) Use session data to populate question titles

richpjames:
# Context
Some parts of the questionaire are dependent on data stored in the session. 
This PR builds out that functionality in a couple of ways: 
1. The release date should be shown in the title of the placement date page
2. Dependent on the answer to the question 'Which of the following best describes the sentence type [name] is on?' the following question can either be 'What type of release will the application support' or 'Which of the following options best describes the situation' and the possible responses (in the form of radio buttons) for each question is also variable.

[Trello card](https://trello.com/c/QxdnPm4F/721-use-session-data-to-populate-form)

# Changes in this PR
- First we pass the session data to the constructor of every page object in the Apply questionaire
- Then we use the session passed into to populate the form pages
- Based on the session data we calculate the potential answers to the given questinos 
- Due to this change we also need to make the how the 'Back' button functionality works (see the last commit).  As we do not know in advance which page the user has come from we need to calculate the value of the 'Back' button dynamically

### After
image
image
image
image

### DIFF
--- a/integration_tests/pages/apply/enterCrn.ts
+++ b/integration_tests/pages/apply/enterCrn.ts
@@ -11,10 +11,6 @@ export default class EnterCRNPage extends Page {
     this.getTextInputByIdAndEnterDetails('crn', crn)
   }
 
-  public clickSubmit(): void {
-    cy.get('button').click()
-  }
-
   public shouldShowErrorMessage(person: Person): void {
     cy.get('.govuk-error-summary').should('contain', `No person with an CRN of '${person.crn}' was found`)
     cy.get(`[data-cy-error-crn]`).should('contain', `No person with an CRN of '${person.crn}' was found`)
--- a/server/form-pages/apply/basic-information/placementDate.ts
+++ b/server/form-pages/apply/basic-information/placementDate.ts
@@ -1,24 +1,32 @@
 import type { ObjectWithDateParts, YesOrNo } from 'approved-premises'
 
 import TasklistPage from '../../tasklistPage'
-import { dateAndTimeInputsAreValidDates } from '../../../utils/utils'
+import {
+  dateAndTimeInputsAreValidDates,
+  formatDateString,
+  retrieveQuestionResponseFromSession,
+} from '../../../utils/utils'
 
 export default class PlacementDate implements TasklistPage {
   name = 'placement-date'
 
-  title = 'Is [release-date] the date you want the placement to start?'
+  title: string
 
   body: ObjectWithDateParts<'startDate'> & {
     startDateSameAsReleaseDate: YesOrNo
   }
 
-  constructor(body: Record<string, unknown>) {
+  constructor(body: Record<string, unknown>, session: Record<string, unknown>) {
     this.body = {
       'startDate-year': body['startDate-year'] as string,
       'startDate-month': body['startDate-month'] as string,
       'startDate-day': body['startDate-day'] as string,
       startDateSameAsReleaseDate: body.startDateSameAsReleaseDate as YesOrNo,
     }
+
+    const formattedReleaseDate = formatDateString(retrieveQuestionResponseFromSession(session, 'releaseDate'))
+
+    this.title = `Is ${formattedReleaseDate} the date you want the placement to start?`
   }
 
   next() {
--- a/server/form-pages/apply/basic-information/situation.ts
+++ b/server/form-pages/apply/basic-information/situation.ts
@@ -1,24 +1,35 @@
+import { SessionDataError } from '../../../utils/errors'
+import { retrieveQuestionResponseFromSession } from '../../../utils/utils'
 import TasklistPage from '../../tasklistPage'
+import { SentenceTypesT } from './sentenceType'
 
-// TODO: The options that appear will depend on the response to the previous question.
-// For now we'll return all options
-export enum Situations {
-  riskManagement = 'Referral for risk management',
-  residencyManagenent = 'Residency management',
-  bailAssessment = 'Bail assessment for community penalty',
-  bailSentence = 'Bail sentence',
-}
+const situations = {
+  riskManagement: 'Referral for risk management',
+  residencyManagement: 'Residency management',
+  bailAssessment: 'Bail assessment for community penalty',
+  bailSentence: 'Bail sentence',
+} as const
+
+type CommunityOrderSituations = Pick<typeof situations, 'riskManagement' | 'residencyManagement'>
+type BailPlacementSituations = Pick<typeof situations, 'bailAssessment' | 'bailSentence'>
+type SentenceType = Extract<SentenceTypesT, 'communityOrder' | 'bailPlacement'>
 
 export default class Situation implements TasklistPage {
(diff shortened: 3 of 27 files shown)


server/services/applicationService.ts
```diff
@@ -81,13 +82,17 @@ export default class ApplicationService {
...
-  private getSessionData(request: Request) {
+  private getAllSessionData(request: Request) {
     const data = this.fetchOrInitializeSessionData(request.session, request.params.task, request.params.id)
 
-    return data.application[request.params.id][request.params.task][request.params.page] || {}
+    return data.application[request.params.id] || {}
+  }
+
+  private getSessionDataForPage(request: Request) {
+    return this.getAllSessionData(request)[request.params.task][request.params.page] || {}
```
pezholio, line 95: Probably not for now, but just flagging that we did discuss flattening the structure of the session data. I think it might still be worth keeping the task, but maybe not the page.

server/utils/utils.test.ts
```diff
@@ -177,3 +179,18 @@ describe('dateAndTimeInputsAreValidDates', () => {
     expect(result).toEqual(false)
   })
 })
+
+describe('retrieveQuestionResponseFromSession', () => {
+  it("throws a SessionDataError if the property doesn't exist", () => {
+    expect(() => retrieveQuestionResponseFromSession({}, '')).toThrow(SessionDataError)
+  })
+  it('returns the property if it does existion', () => {
```
pezholio, line 187: Can we have a linebreak here?

server/utils/utils.ts
```diff
@@ -94,3 +105,16 @@ export const dateAndTimeInputsAreValidDates = <K extends string | number>(
 
   return true
 }
+/**
+ * Retrieves response for a given question from the session object.
+ * @param sessionData the session data for an application.
+ * @param question the question that we need the response for in camelCase.
+ * @returns name converted to proper case.
+ */
+export const retrieveQuestionResponseFromSession = <T>(sessionData: Record<string, unknown>, question: string) => {
```
pezholio, line 114: What happens if the question name is different from the page name?

server/utils/utils.ts
```diff
@@ -29,11 +28,23 @@ export const initialiseName = (fullName?: string): string | null => {
...
+/**
+ * Converts a string from any case to kebab-case
+ * @param string string to be converted.
+ * @returns name converted to kebab-case.
+ */
+const kebabCase = (string: string) =>
+  string
+    .replace(/([a-z])([A-Z])/g, '$1-$2')
+    .replace(/[\s_]+/g, '-')
+    .toLowerCase()
```
pezholio, line 40: This will come in veeeeeery handy!

server/utils/utils.ts
```diff
@@ -94,3 +105,16 @@ export const dateAndTimeInputsAreValidDates = <K extends string | number>(
 
   return true
 }
+/**
+ * Retrieves response for a given question from the session object.
+ * @param sessionData the session data for an application.
+ * @param question the question that we need the response for in camelCase.
+ * @returns name converted to proper case.
+ */
```
pezholio, line 113: Thanks for keeping us honest with the JSDoc. I was getting lazy!

server/utils/utils.ts
```diff
@@ -94,3 +105,16 @@ export const dateAndTimeInputsAreValidDates = <K extends string | number>(
...
+export const retrieveQuestionResponseFromSession = <T>(sessionData: Record<string, unknown>, question: string) => {
+  try {
+    return sessionData['basic-information'][kebabCase(question)][question] as T
+  } catch (e) {
+    throw new SessionDataError()
```
pezholio, line 118: Very much a fan of throwing specific errors here 👍 Would it be worth adding a message here? Something like `Question ${question} was not found in the session`.

server/form-pages/apply/basic-information/placementDate.ts
```diff
@@ -1,24 +1,32 @@
...
-  constructor(body: Record<string, unknown>) {
+  constructor(body: Record<string, unknown>, session: Record<string, unknown>) {
     this.body = {
       'startDate-year': body['startDate-year'] as string,
       'startDate-month': body['startDate-month'] as string,
       'startDate-day': body['startDate-day'] as string,
       startDateSameAsReleaseDate: body.startDateSameAsReleaseDate as YesOrNo,
     }
+
+    const formattedReleaseDate = formatDateString(retrieveQuestionResponseFromSession(session, 'releaseDate'))
```
pezholio, line 27: This is :chefskiss: - thanks for doing the thinking on this!

server/form-pages/apply/basic-information/situation.ts
```diff
@@ -44,12 +57,22 @@ export default class Situation implements TasklistPage {
...
       return {
         value: key,
-        text: Situations[key],
+        text: this.situations[key],
         checked: this.body.situation === key,
       }
     })
   }
+
+  getSituationsForSentenceType(sessionSentenceType: SentenceType): CommunityOrderSituations | BailPlacementSituations {
```
pezholio, line 69: Nice work! 👍

server/form-pages/apply/basic-information/situation.ts
```diff
@@ -1,24 +1,37 @@
...
+  riskManagement: 'Referral for risk management',
+  residencyManagement: 'Residency management',
+} as const
+type CommunityOrderSituations = typeof communityOrderSituations
+
+const bailPlacementSituations = {
+  bailAssessment: 'Bail assessment for community penalty',
+  bailSentence: 'Bail sentence',
+} as const
+type BailPlacementSituations = typeof bailPlacementSituations
```
pezholio, line 16: I wonder here if if might be clearer to have `const situations` and then use the `Pick` utility type to pick the ones you want for each situation (as you have in the next commit)

integration_tests/pages/apply/enterCrn.ts
```diff
@@ -11,10 +11,6 @@ export default class EnterCRNPage extends Page {
     this.getTextInputByIdAndEnterDetails('crn', crn)
   }
 
-  public clickSubmit(): void {
-    cy.get('button').click()
-  }
-
```
pezholio, line 17: Yeah! This has bugged me for ages!

pezholio (approved):
Looking good! Don't be put off by the amount of comments, most of these are just me riffing 😆

server/services/applicationService.ts
richpjames: Yeah I did wonder if it would be best to consolidate the two objects that we pass into the Page constructor as the session is just a larger version of the body but I think it's best we wait until we have the flattened structure.

server/utils/utils.ts
richpjames: Currently we can't handle that scenario but we could add an optional third parameter if it's needed

pezholio: It's probably not worth it TBH, as we're going to change the structure

pezholio (approved):
LGTM 👍

PR #173, opened 2022-09-15
(613) Contract test between frontend and backend

pezholio:
This adds the general setup for a contract (e2e) test to create a booking in the Dev environment. This runs after Dev has been deployed. I'm using Cucumber for the specs here, and moved the page objects to a shared folder that can be used by the integration and e2e tests. We could migrate the integration tests to Cucumber further down the line if we want / need to.


pezholio:
[Contract test between frontend and backend](https://trello.com/c/NF4QRAAl/613-contract-test-between-frontend-and-backend)

richpjames (approved):
Looks great. I think it would be worth trying to move our current tests over to Cucumber syntax as we touch them

pezholio:
Yeah, I deffo think it's a sensible way forward. We've made it a bit easier for ourselves by adding some cucumber style comments.

PR #174, opened 2022-09-15
(627) Type of AP required screen

pezholio:
This adds the Type of AP Required screen. We still need to fetch the offender's to populate the title, but this will come later:

## Screenshots

image

image

### DIFF
--- a/server/utils/formUtils.ts
+++ b/server/utils/formUtils.ts
@@ -36,3 +36,13 @@ export const convertObjectsToRadioItems = (
     }
   })
 }
+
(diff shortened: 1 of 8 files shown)


server/utils/formUtils.ts
```diff
@@ -36,3 +36,13 @@ export const convertObjectsToRadioItems = (
     }
   })
 }
+
+export function convertKeyValuePairToRadioItems<T>(object: T, checkedItem: string): Array<RadioItems> {
```
richpjames, line 40: handy!

richpjames (approved):

PR #179, opened 2022-09-20
(721) -  fix - Use session to store previous page value

richpjames:
#169 broke the apply journey when reaching the 'release date' question as we weren't passing a value to the 'Back' button.
The value we were passing as 'previousPage' was not being persisted in the flash.
Instead I am suggesting we store the 'previousPage' in the session instead of the flash as this will be more robust.

The tests didn't fail on this issue as we weren't verifying that we had reached the 'release date' page so I corrected this error before implementing the solution.


pezholio (approved):
Cool - LGTM 👍

PR #182, opened 2022-09-20
(724) Move crn search to start of apply journey

richpjames:
# Context
Because the CRN is a unique identifier - we should have the person search sitting seperately from the application process. 
This means that we have to remove the CRN lookup and confirmation from the regular 'Apply' form flow and have it follow the pattern of the BookingController 'new' method.

[Trello ticket](https://trello.com/c/fTV8f4cL/724-move-crn-search-to-the-start-of-the-journey)


server/controllers/apply/applicationsController.test.ts
```diff
@@ -39,12 +51,95 @@ describe('applicationsController', () => {
...
         pageHeading: 'Apply for an Approved Premises (AP) placement',
       })
     })
   })
 
+  describe('new', () => {
+    describe('If there is a CRN in the flash', () => {
+      it('it should render the start of the application form', async () => {
+        const person = personFactory.build()
+        personService.findByCrn.mockResolvedValue(person)
```
pezholio, line 64: Might make sense to have this in a `beforeEach` step, as we have in the booking test now

server/controllers/apply/applicationsController.ts
```diff
@@ -13,17 +16,47 @@ export default class ApplicationsController {
...
+        errors,
+        errorSummary,
+        ...userInput,
+      })
     }
   }
 
   create(): RequestHandler {
     return async (req: Request, res: Response) => {
       const uuid = await this.applicationService.createApplication(req.user.token)
```
pezholio, line 57: I think we'll want to pass the user to the application here too, but that can maybe come later

pezholio (approved):
LGTM 👍

server/controllers/apply/applicationsController.test.ts
richpjames: Absolutely!

pezholio (approved):
LGTM 👍

PR #184, opened 2022-09-21
Update booking type to match API response

pezholio:
This updates the Booking type to better match what comes back from the API. At the moment, dev is broken.

### DIFF
--- a/integration_tests/pages/manage/booking/show.ts
+++ b/integration_tests/pages/manage/booking/show.ts
@@ -15,9 +15,9 @@ export default class BookingShowPage extends Page {
   }
 
   shouldShowBookingDetails(booking: Booking): void {
-    cy.get('dl[data-cy-expected-dates]').within(() => {
-      this.assertDefinition('Arrival date', formatDateString(booking.expectedArrivalDate))
-      this.assertDefinition('Departure date', formatDateString(booking.expectedDepartureDate))
+    cy.get('dl[data-cy-dates]').within(() => {
+      this.assertDefinition('Arrival date', formatDateString(booking.arrivalDate))
+      this.assertDefinition('Departure date', formatDateString(booking.departureDate))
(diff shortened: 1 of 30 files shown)


integration_tests/pages/manage/booking/show.ts
```diff
@@ -16,8 +16,8 @@ export default class BookingShowPage extends Page {
 
   shouldShowBookingDetails(booking: Booking): void {
     cy.get('dl[data-cy-expected-dates]').within(() => {
```
richpjames, line 18: should we update this attribute too?

server/testutils/factories/booking.ts
```diff
@@ -86,8 +87,8 @@ class BookingFactory extends Factory<Booking> {
 
 export default BookingFactory.define(() => ({
   person: personFactory.build(),
-  arrivalDate: faker.date.soon().toISOString(),
-  departureDate: faker.date.future().toISOString(),
+  arrivalDate: formatISO(faker.date.soon(), { representation: 'date' }),
+  departureDate: formatISO(faker.date.future(), { representation: 'date' }),
```
richpjames, line 91: I'm happy to implement it but shouldn't we use the same date format across all our factories?

richpjames (approved):
Looks good, left a couple of comments

server/testutils/factories/booking.ts
pezholio: Yeah, I kinda want to pick these things up as we encounter them, but I think we'll need to.

integration_tests/pages/manage/booking/show.ts
pezholio: I think it's fine, as far as the user is concerned they're expected dates. We just need to make the internal naming consistent

pezholio: Actually, as this data attribute is an internal thing, it probably does make sense to be consistent here too! 😆

PR #185, opened 2022-09-21
(714) Save in-progress application

richpjames:
# Changes in this PR

We need to be able to save applications to our API.
This PR enables that by calling the [specified API endpoint](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/main/src/main/resources/static/api.yml#L711) each time we save the application to the session.
This is our first PUT request so there was a little setup with the client. I think the same format as a POST should work...

Also in this PR: 
- I've started a DateFormat class that I think should be the place where all our date formatting lives in order for us to be able to change them more easily in future.
- I fixed a bug where the Booking confirmation page wasn't working

[Trello ticket](https://trello.com/c/kKJe8xHw/714-save-in-progress-application)

### DIFF
--- a/server/@types/approved-premises/index.d.ts
+++ b/server/@types/approved-premises/index.d.ts
@@ -13,6 +13,7 @@ declare module 'approved-premises' {
   export type PremisesCapacityItem = schemas['PremisesCapacityItem']
   export type PremisesCapacity = Array<PremisesCapacityItem>
   export type ApplicationSummary = schemas['ApplicationSummary']
+  export type Application = schemas['Application']
 
   // A utility type that allows us to define an object with a date attribute split into
   // date, month, year (and optionally, time) attributes. Designed for use with the GOV.UK
@@ -248,6 +249,15 @@ declare module 'approved-premises' {
       arrivalDate: string
       daysSinceApplicationRecieved: number
       status: 'In progress' | 'Submitted' | 'Information Requested' | 'Rejected'
+    }
+    Application: {
+      id: string
+      crn: string
+      createdByProbationOfficerId: string
+      schemaVersion: string
+      createdAt: string
+      submittedAt?: string
+      data: Record<string, unknown>
     }
   }
 }
--- a/server/@types/approved-premises/index.d.ts
+++ b/server/@types/approved-premises/index.d.ts
@@ -57,6 +57,10 @@ declare module 'approved-premises' {
 
   export type BookingStatus = 'arrived' | 'awaiting-arrival' | 'not-arrived' | 'departed' | 'cancelled'
 
+  export type TaskNames = 'basic-information' | 'type-of-ap'
+
+  export type ApplicationData = Record<TaskNames, unknown>
+
   export interface HtmlAttributes {
     [key: string]: string
   }
--- a/server/data/restClient.ts
+++ b/server/data/restClient.ts
@@ -26,6 +26,8 @@ interface PostRequest {
   raw?: boolean
 }
 
+interface PutRequest extends PostRequest {}
+
 interface StreamRequest {
   path?: string
   headers?: Record<string, string>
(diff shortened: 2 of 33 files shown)


server/utils/dateFormats.ts
```diff
@@ -0,0 +1,8 @@
+import formatISO from 'date-fns/formatISO'
+
+// eslint-disable-next-line
+export class DateFormats {
```
pezholio, line 4: Good call on this! 👍

server/data/restClient.ts
```diff
@@ -103,6 +103,37 @@ export default class RestClient {
     }
   }
 
+  async put({
+    path = null,
+    headers = {},
+    responseType = '',
+    data = {},
+    raw = false,
+  }: PostRequest = {}): Promise<unknown> {
```
pezholio, line 112: Should this be a `PutRequest`?

server/@types/approved-premises/index.d.ts
```diff
@@ -245,5 +246,14 @@ declare module 'approved-premises' {
...
     }
+    Application: {
+      id: string
+      crn: string
+      createdByProbationOfficerId: string
+      schemaVersion: string
+      createdAt: string
+      submittedAt?: string
+      data: string
+    }
```
pezholio, line 257: We also have an `Application` type now in the form-pages. It might be worth changing that to `ApplicationData` and giving the `data` a type of `ApplicationData`

server/services/applicationService.ts
```diff
@@ -56,18 +56,29 @@ export default class ApplicationService {
...
     }
   }
 
+  private saveToSession(page: TasklistPage, request: Request) {
+    request.session = this.fetchOrInitializeSessionData(request.session, request.params.task, request.params.id)
+    request.session.application[request.params.id][request.params.task][request.params.page] = page.body
+    request.session.previousPage = page.name
+  }
+
+  private async saveToApi(request: Request) {
```
pezholio, line 76: I think we only want to save to the API at the end of a task, rather than update each time

server/services/applicationService.ts
```diff
@@ -56,18 +56,29 @@ export default class ApplicationService {
...
+    const client = this.applicationClientFactory(request.user.token)
+
+    await client.update(request.session.application[request.params.id] as Application, request.params.id)
```
pezholio, line 79: Should this be updating only the `data` attribute of the application here?

pezholio (commented):
Couple of comments, think you're broadly on the right track tho 👍

server/utils/dateFormats.ts
richpjames: Thanks, I'll add the other date formats in a seperate PR tomorrow!

server/data/restClient.ts
richpjames: So we basically want a type that has the same properties as `PostRequest` but is called `PutRequest` but the linter says we can't extend a type without changing the properties... I think it's probably best to just disable the linter though

pezholio: Aaah - maybe give the type a more generic name like `PostOrPutRequest`?

server/services/applicationService.ts
richpjames: Can I ask why?
We'd have to build in some extra functionality to only save to the API after the end of a task whereas I can't see an issue saving after every response - but I could be missing something!

pezholio: Thanks for asking! I guess I was thinking it was to prevent any partially saved applications from being sent to the API, but actually, that's probably not the end of the world!

richpjames: I think we would be able to retrieve a partial application and repopulate the form. If it breaks anything we can add the 'only save to API at end of task' functionality later?

server/services/applicationService.ts
richpjames: Yes - good spot

server/data/restClient.ts
richpjames: I've opted to turn the lint rule off - I don't think it's useful

server/services/applicationService.ts
pezholio: 👍🏻

pezholio (approved):

PR #186, opened 2022-09-22
Add basic tasklist

pezholio:
This adds a basic tasklist that we get redirected to at the end of a step on the journey. There's still a bit more to do to get a finished journey, but this is a start.

# Screenshot

image

### DIFF
--- a/assets/sass/components/_task-list.scss
+++ b/assets/sass/components/_task-list.scss
@@ -1,0 +1,65 @@
+// Task list pattern
+
+.app-task-list {
+  list-style-type: none;
+  padding-left: 0;
+  margin-top: 0;
+  margin-bottom: 0;
+  @include govuk-media-query($from: tablet) {
+    min-width: 550px;
+  }
+}
+
+.app-task-list__section {
+  display: table;
+  @include govuk-font($size:24, $weight: bold);
+}
+
+.app-task-list__section-number {
+  display: table-cell;
+
+  @include govuk-media-query($from: tablet) {
+    min-width: govuk-spacing(6);
+    padding-right: 0;
+  }
+}
+
+.app-task-list__items {
+  @include govuk-font($size: 19);
+  @include govuk-responsive-margin(9, "bottom");
+  list-style: none;
+  padding-left: 0;
+  @include govuk-media-query($from: tablet) {
+    padding-left: govuk-spacing(6);
+  }
+}
+
+.app-task-list__item {
+  border-bottom: 1px solid $govuk-border-colour;
+  margin-bottom: 0 !important;
+  padding-top: govuk-spacing(2);
+  padding-bottom: govuk-spacing(2);
+  @include govuk-clearfix;
+}
+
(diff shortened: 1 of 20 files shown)


server/controllers/apply/applicationsController.test.ts
```diff
@@ -57,6 +60,38 @@ describe('applicationsController', () => {
...
+  describe('show', () => {
+    it('renders the task list if an application exists', () => {
+      const requestHandler = applicationsController.show()
+
+      const application = createMock<Application>()
+      const id = 'some-uuid'
+
+      request = createMock<Request>({
+        params: { id },
+        session: { application: { [`${id}`]: application } },
```
richpjames, line 72: you should be able to just do `[id]` here

server/controllers/apply/applications/pagesController.test.ts
```diff
@@ -148,6 +148,20 @@ describe('pagesController', () => {
...
     })
 
+    it('redirects to the tasklist if there is no next page', async () => {
+      page.next.mockReturnValue(undefined)
+
+      applicationService.save.mockReturnValue()
+
+      const requestHandler = pagesController.update()
+
+      await requestHandler({ ...request }, response)
```
richpjames, line 158: I think we could just use `await requestHandler(request, response)` here

server/controllers/apply/applications/pagesController.test.ts
```diff
@@ -148,6 +148,20 @@ describe('pagesController', () => {
       )
     })
 
+    it('redirects to the tasklist if there is no next page', async () => {
+      page.next.mockReturnValue(undefined)
```
richpjames, line 152: Incredibly minor but would setting `page.next` to `undefined` more accurately describe the behaviour? 
EG `page.next = undefined`
What we have here is fine if the types don't like it

assets/sass/components/_task-list.scss
```diff
@@ -0,0 +1,69 @@
...
+  @include govuk-media-query($from: 450px) {
+    float: left;
+  }
+}
+
+// The `app-task-list__task-completed` class was previously used on the task
+// list for the completed tag (changed in 86c90ec) – it's still included here to
+// avoid breaking task lists in existing prototypes.
+.app-task-list__tag,
+.app-task-list__task-completed {
```
richpjames, line 60: I don't think we're using `.app-task-list__task-completed` in the template currently... but maybe we will in future?

richpjames (approved):
Yeah looks goods. I've left some minor comments around syntax which you can feel free to ignore if you're busy :)

assets/sass/components/_task-list.scss
pezholio: Ah, probably not. Will ding it

server/controllers/apply/applications/pagesController.test.ts
pezholio: Yeah, it didn't like it. Will keep it as is

PR #189, opened 2022-09-22
Remove `And` from step definitions

pezholio:
See https://github.com/badeball/cypress-cucumber-preprocessor/issues/821

This wasn't caught in CI, because they're only used in the e2e tests at the moment


richpjames (approved):
LGTM

PR #199, opened 2022-09-27
(751) Complete person search

pezholio:
This PR updates the Apply confirmation screen to show all of the details of a prisoner that come back from the API

# Screenshot

image

### DIFF
--- a/cypress_shared/pages/apply/confirmDetails.ts
+++ b/cypress_shared/pages/apply/confirmDetails.ts
@@ -1,4 +1,5 @@
 import type { Person } from 'approved-premises'
+import { DateFormats } from '../../../server/utils/dateUtils'
 
 import Page from '../page'
 
@@ -11,6 +12,14 @@ export default class ConfirmDetailsPage extends Page {
     cy.get('dl').within(() => {
       this.assertDefinition('Name', this.person.name)
       this.assertDefinition('CRN', this.person.crn)
+      this.assertDefinition('Date of Birth', DateFormats.isoDateToUIDate(this.person.dateOfBirth, { format: 'short' }))
+      this.assertDefinition('NOMS Number', this.person.nomsNumber)
+      this.assertDefinition('Nationality', this.person.nationality)
+      this.assertDefinition('Religion or belief', this.person.religionOrBelief)
+      this.assertDefinition('Sex', this.person.sex)
+
+      cy.get(`[data-cy-status]`).should('have.attr', 'data-cy-status').and('equal', this.person.status)
+      this.assertDefinition('Prison', this.person.prisonName)
     })
   }
 }
--- a/server/controllers/apply/applicationsController.test.ts
+++ b/server/controllers/apply/applicationsController.test.ts
@@ -115,7 +115,8 @@ describe('applicationsController', () => {
         expect(response.render).toHaveBeenCalledWith('applications/confirm', {
           pageHeading: `Confirm ${person.name}'s details`,
           ...person,
-          dateOfBirth: DateFormats.isoDateToUIDate(person.dateOfBirth),
+          date: DateFormats.dateObjtoUIDate(new Date()),
+          dateOfBirth: DateFormats.isoDateToUIDate(person.dateOfBirth, { format: 'short' }),
           errors: {},
           errorSummary: [],
         })
@@ -135,7 +136,8 @@ describe('applicationsController', () => {
         expect(response.render).toHaveBeenCalledWith('applications/confirm', {
           pageHeading: `Confirm ${person.name}'s details`,
           ...person,
-          dateOfBirth: DateFormats.isoDateToUIDate(person.dateOfBirth),
+          date: DateFormats.dateObjtoUIDate(new Date()),
+          dateOfBirth: DateFormats.isoDateToUIDate(person.dateOfBirth, { format: 'short' }),
           errors: errorsAndUserInput.errors,
           errorSummary: errorsAndUserInput.errorSummary,
           ...errorsAndUserInput.userInput,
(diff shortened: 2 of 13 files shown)


pezholio:
[Complete person search](https://trello.com/c/zHTUS6JS/751-complete-person-search)

server/utils/utils.test.ts
```diff
@@ -45,3 +51,45 @@ describe('retrieveQuestionResponseFromSession', () => {
...
+
+    expect(removeBlankSummaryListItems(items)).toEqual([
+      {
+        key: {
+          text: 'Names',
+        },
+        value: {
+          text: 'Name',
+        },
+      },
```
edavey, line 92: nice utility! should we also have some `html:` values to describe how those values (like the `text:` values) are either retained or discarded depending on whether they are present?

pezholio: Oh, good call 👍

server/controllers/apply/applicationsController.test.ts
```diff
@@ -135,6 +136,7 @@ describe('applicationsController', () => {
         expect(response.render).toHaveBeenCalledWith('applications/confirm', {
           pageHeading: `Confirm ${person.name}'s details`,
           ...person,
+          date: DateFormats.dateObjtoUIDate(new Date()),
           dateOfBirth: DateFormats.isoDateToUIDate(person.dateOfBirth),
```
edavey, line 140: Just a point about this ` dateOfBirth: DateFormats.isoDateToUIDate(person.dateOfBirth),` which I realise isn't part of this PR. Probably:

> Thursday, 20 December 1982

is not how we want to express a DoB? The day of the week is probably a distraction? 

and something like:

> 20 December 1982

or

> 20-Dec-1982

would be clearer and more useful?

server/testutils/factories/person.ts
```diff
@@ -9,6 +9,9 @@ export default Factory.define<Person>(() => ({
   name: faker.name.fullName(),
   dateOfBirth: DateFormats.formatApiDate(faker.date.past()),
   sex: faker.helpers.arrayElement(['Male', 'Female', 'Other', 'Prefer not to say']),
+  status: faker.helpers.arrayElement(['InCustody', 'InCommunity']),
+  nomsNumber: `NOMS${faker.datatype.number({ min: 100, max: 999 })}`,
   nationality: faker.address.country(),
-  religion: faker.random.word(),
+  religionOrBelief: faker.helpers.arrayElement(['Christian', 'Muslim', 'Jewish', 'Hindu', 'Buddhist', 'Sikh', 'None']),
+  prisonName: `HMP ${faker.address.street()}`,
```
edavey, line 16: Small point, but in the interests of keeping the faked/fixture data vaguely realistic maybe we ought to link `status` and `prisonName` as it only makes sense to show a prison name if the person is in custody. 

Again, perhaps hardcoding would be easier?

server/testutils/factories/person.ts
```diff
@@ -9,6 +9,9 @@ export default Factory.define<Person>(() => ({
   name: faker.name.fullName(),
   dateOfBirth: DateFormats.formatApiDate(faker.date.past()),
```
edavey, line 10: This is from another PR, but just to point out that to be usefully realistic perhaps we should ensure that `dateOfBirth` is at least 18 years in the past?

On a general point I wonder how useful it is to have these person fixtures built dynamically with randomised properties. Would it be better to have the attributes hardcoded? And then to develop new specialised fixtures for different people as we need them eg:

- man in community
- trans woman in prison

cypress_shared/pages/apply/confirmDetails.ts
```diff
@@ -11,6 +12,14 @@ export default class ConfirmDetailsPage extends Page {
     cy.get('dl').within(() => {
       this.assertDefinition('Name', this.person.name)
       this.assertDefinition('CRN', this.person.crn)
+      this.assertDefinition('Date of Birth', DateFormats.isoDateToUIDate(this.person.dateOfBirth))
+      this.assertDefinition('NOMS Number', this.person.nomsNumber)
+      this.assertDefinition('Nationality', this.person.nationality)
+      this.assertDefinition('Religion or belief', this.person.religionOrBelief)
+      this.assertDefinition('Sex', this.person.sex)
```
edavey, line 19: quite likely I'm not understanding properly how these tests are wired up, but are we missing the demonstration that the `removeBlankSummaryListItems` function is being used to suppress empty lines? And could we do that with a negative assertion here? 

I'm not clear where/how the `person` picture is injected into the test but could we:

1. declare that `person.genderIdentity` is `null`
2. assert `this.assertNoDefinition('Gender identity')`

edavey (approved):
great stuff, really easy to follow, thanks.

server/controllers/apply/applicationsController.test.ts
pezholio: Yeah, that's a good point. We default to that format as it's the GDS standard, but it makes less sense for a date of birth

pezholio: I've gone with dd/mm/yyyy, which mirrors what is in NOMIS

cypress_shared/pages/apply/confirmDetails.ts
pezholio: This is a tricky one, we're trying not to dig down too much into individual sub-scenarios and focus on the happy paths, as Cypress tests tend to add a fair amount of overhead in terms of test run size. 

What I have done in the past when there's logic in views is test the unit test the views themselves  (see https://github.com/UKGovernmentBEIS/regulated-professions-register/blob/main/views/professions/show.njk.spec.ts), but again, I'm not sure how valuable this is.

Cypress has just introduced [Component Testing](https://docs.cypress.io/guides/component-testing/writing-your-first-component-test), which seems like a good candidate, but this seems more focussed on client-side apps.

server/testutils/factories/person.ts
pezholio: Yeah, we've not really thought too much about realistic fixture data, as the fixtures as a means rather than an end, but this might be a good idea for another PR 👍

edavey (approved):
👍

PR #203, opened 2022-09-27
(737) Contract tests for creating a lost bed

pezholio:
This adds a contract test for creating a lost bed, as well as the unhappy path for any errors. I've also refactored the specs a bit to make them more lightweight and lean on the page objects a bit more.

### DIFF
--- a/e2e/tests/manage.feature
+++ b/e2e/tests/manage.feature
@@ -12,5 +12,13 @@ Feature: Manage an Approved Premises
         Scenario: Creating a lost bed
                 When I access the premises homepage
                 And I choose a premises
+                And I navigate to the lost beds create page
                 And I create a lost bed
                 Then I should see a notification that the lost bed has been created
+
+        Scenario: Showing lost bed errors
+                When I access the premises homepage
+                And I choose a premises
+                And I navigate to the lost beds create page
+                And I attempt to create a lost bed without the necessary information
+                Then I should see a list of the problems encountered creating the lost bed
--- a/e2e/tests/manage.feature
+++ b/e2e/tests/manage.feature
@@ -1,7 +1,16 @@
 Feature: Manage an Approved Premises
-        Scenario: Creating a booking
+        Background:
                 Given I am logged in
+
+        Scenario: Creating a booking
+                When I access the premises homepage
                 And I see a list of premises
                 And I choose a premises
                 And I create a booking
-                Then I should see a confirmation screen
+                Then I should see a confirmation screen for my booking
+
+        Scenario: Creating a lost bed
+                When I access the premises homepage
+                And I choose a premises
+                And I create a lost bed
+                Then I should see a notification that the lost bed has been created
--- a/e2e/tests/manage.feature
+++ b/e2e/tests/manage.feature
@@ -3,22 +3,16 @@ Feature: Manage an Approved Premises
                 Given I am logged in
 
         Scenario: Creating a booking
-                When I access the premises homepage
-                And I see a list of premises
-                And I choose a premises
+                Given I'm managing a premises
                 And I create a booking
                 Then I should see a confirmation screen for my booking
 
         Scenario: Creating a lost bed
-                When I access the premises homepage
-                And I choose a premises
-                And I navigate to the lost beds create page
+                Given I'm managing a premises
                 And I create a lost bed
                 Then I should see a notification that the lost bed has been created
 
         Scenario: Showing lost bed errors
-                When I access the premises homepage
-                And I choose a premises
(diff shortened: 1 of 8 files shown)


e2e/tests/manage.feature
```diff
@@ -1,7 +1,16 @@
...
                 And I choose a premises
                 And I create a booking
-                Then I should see a confirmation screen
+                Then I should see a confirmation screen for my booking
+
+        Scenario: Creating a lost bed
+                When I access the premises homepage
+                And I choose a premises
+                And I create a lost bed
+                Then I should see a notification that the lost bed has been created
```
edavey, line 17: Is it worth also have a basic unhappy path scenario, e.g.

```
Scenario: Fail to create a lost bed
  Given I'm managing a premises
  And I attempt to create a lost bed without the necessary information
  Then I should see that the lost bed has not been created
  And I should see a list of the problems encountered creating the lost bed
```

This might be helpful, as validation, error messaging is a big area of coupling...

edavey (approved):
Nice! 

I've made a suggestion about adding an unhappy scenario...

e2e/tests/stepDefinitions/lostBeds.ts
```diff
@@ -4,23 +4,35 @@ import { LostBedCreatePage } from '../../../cypress_shared/pages/manage'
...
+  lostBedCreatePage.clickSubmit()
+})
+
 Then('I should see a notification that the lost bed has been created', () => {
   cy.get('.govuk-notification-banner').should('contain', 'Lost bed logged')
 })
+
+Then('I should see a list of the problems encountered creating the lost bed', () => {
+  const lostBedCreatePage = new LostBedCreatePage()
+  lostBedCreatePage.shouldShowErrorMessagesForFields(['startDate', 'endDate', 'numberOfBeds', 'reason'])
```
edavey: I don't think the matcher

```
shouldShowErrorMessagesForFields(fields: Array<string>): void {
    fields.forEach(field => {
      cy.get('.govuk-error-summary').should('contain', errorLookups[field]?.empty)
      cy.get(`[data-cy-error-${field}]`).should('contain', errorLookups[field]?.empty)
    })
  }
```

works to verify that none of the "reason" radios have been selected. Cypress is telling me:

```
AssertionError
Timed out retrying after 4000ms: 
Expected to find element: input[name="lostBed[reason]"][value="undefined"], 
but never found it.
```

pezholio: Ah, sorry, that's down to some missing env vars

e2e/tests/stepDefinitions/booking.ts
```diff
@@ -0,0 +1,38 @@
...
+  BookingFindPage,
+  PremisesShowPage,
+} from '../../../cypress_shared/pages/manage'
+
+import bookingFactory from '../../../server/testutils/factories/booking'
+import personFactory from '../../../server/testutils/factories/person'
+import keyWorkerFactory from '../../../server/testutils/factories/keyWorker'
+
+const keyWorker = keyWorkerFactory.build({ name: Cypress.env('keyworker_name') })
+const person = personFactory.build({ name: Cypress.env('offender_name'), crn: Cypress.env('offender_crn') })
```
edavey, line 15: I'm thinking maybe we can be more helpful if these expected environment variables aren't present, and raise a clear error early?

e.g.

```ts
const keyWorkerName = Cypress.env('keyworker_name')
expect(keyWorkerName, 'keyWorkerName was set').to.be.a('string').and.not.be.empty
```

I've made a PR (https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/208) for doing this to existing Cypress variables in `e2e/tests/stepDefinitions/manage.ts`

e2e/tests/stepDefinitions/lostBeds.ts
```diff
@@ -0,0 +1,45 @@
...
+  cy.get('@premisesShowPage').then((premisesShowPage: PremisesShowPage) => {
+    premisesShowPage.clickLostBedsOption()
+  })
+
+  const lostBedCreatePage = new LostBedCreatePage()
+
+  const lostBed = lostBedFactory.build({
+    startDate: new Date(2022, 1, 11, 0, 0).toISOString(),
+    endDate: new Date(2022, 2, 11, 0, 0).toISOString(),
+    reason: referenceDataFactory.build({ id: Cypress.env('lost_bed_reason_id') }),
```
edavey, line 17: see above, regarding avoiding mysterious missing env var related gotchas

edavey (approved):
Great stuff, just a suggestion about failing fast if a Cypress env is missing

PR #204, opened 2022-09-27
Tweak error messages

pezholio:
This adds and tweaks some error messages to better match what comes from the API, as well as tweaking our date formatter to return an `undefined` value, as oppossed to an empty string.


pezholio:
Yeah, there's definitely something there. I wonder if we should look at something like Pact, or at least check our Wiremock stubs match up with the OpenAPI spec

PR #207, opened 2022-09-28
Filter out blank strings in PUT and POST requests

pezholio:
We will very rarely want to send literal blank strings to the API, and sending blank strings can result in confusing error messages (for example if the API expects a number but gets an empty string, the error will be that the value should be a number), so it’s better to filter these out before sending them to the API.

### DIFF
--- a/server/data/restClient.ts
+++ b/server/data/restClient.ts
@@ -122,7 +122,7 @@ export default class RestClient {
         method === 'post' ? superagent.post(`${this.apiUrl()}${path}`) : superagent.put(`${this.apiUrl()}${path}`)
 
       const result = await request
-        .send(data)
+        .send(this.filterBlanksFromData(data))
         .agent(this.agent)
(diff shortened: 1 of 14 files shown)


server/data/restClient.ts
```diff
@@ -165,4 +111,40 @@ export default class RestClient {
         })
     })
   }
+
+  private async postOrPut(
```
edavey, line 115: might be easier to follow if this refactoring were in a separate commit, before the change in behaviour was introduced?

edavey (dismissed):
👍 nice work. Small nit-pick suggesting doing the refactoring as a separate commit (in future?)...

server/data/restClient.ts
pezholio: That's a good shout 👍

edavey (approved):
nice, super-clear

PR #208, opened 2022-09-28
E2E tests: ensure presence of env vars and hide sensitive vars

edavey:
This is related to my review of [(737) Contract tests for creating a lost bed](https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/203/)

It attempts two things:

### 1. ensure that missing environment variable fail clearly

Uses a utility function to throw an informative error:

```ts
const throwMissingError = (field: string) => {
  throw new Error(`Missing Cypress env variable for '${field}'`)
}
```

e.g.

```ts
const username = Cypress.env('username') || throwMissingError('username')
```

### 2. hide sensitive env vars from Cypress videos and logs

Set `log: false`

when filling in sensitive fields, eg.

```ts
cy.get('input[name="password"]').type(password, { log: false })
```

I followed guidance on:

https://glebbahmutov.com/blog/keep-passwords-secret-in-e2e-tests/

### NB: Continuous integration
I haven't checked the CircleCI setup properly but it doesn't look like
we are using the [security context](https://circleci.com/docs/2.0/contexts/)
system the above article suggests. Quite likely we have it covered in
some other way?

### DIFF
--- a/e2e/tests/stepDefinitions/manage.ts
+++ b/e2e/tests/stepDefinitions/manage.ts
@@ -1,10 +1,7 @@
 import { Given } from '@badeball/cypress-cucumber-preprocessor'
+import throwMissingError from './utils'
 import premisesFactory from '../../../server/testutils/factories/premises'
 import { PremisesShowPage } from '../../../cypress_shared/pages/manage'
-
-const throwMissingError = (field: string) => {
-  throw new Error(`Missing Cypress env variable for '${field}'`)
-}
 
 Given('I am logged in', () => {
   const username = Cypress.env('username') || throwMissingError('username')
--- a/e2e/tests/stepDefinitions/manage.ts
+++ b/e2e/tests/stepDefinitions/manage.ts
@@ -1,11 +1,11 @@
 import { Given } from '@badeball/cypress-cucumber-preprocessor'
-import throwMissingError from './utils'
+import throwMissingCypressEnvError from './utils'
 import premisesFactory from '../../../server/testutils/factories/premises'
 import { PremisesShowPage } from '../../../cypress_shared/pages/manage'
 
 Given('I am logged in', () => {
-  const username = Cypress.env('username') || throwMissingError('username')
-  const password = Cypress.env('password') || throwMissingError('password')
+  const username = Cypress.env('username') || throwMissingCypressEnvError('username')
+  const password = Cypress.env('password') || throwMissingCypressEnvError('password')
 
   cy.visit('/')
   cy.get('input[name="username"]').type(username)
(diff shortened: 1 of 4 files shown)


e2e/tests/stepDefinitions/manage.ts
```diff
@@ -6,14 +6,36 @@ import bookingFactory from '../../../server/testutils/factories/booking'
...
 import keyWorkerFactory from '../../../server/testutils/factories/keyWorker'
 
-const keyWorker = keyWorkerFactory.build({ name: Cypress.env('keyworker_name') })
-const person = personFactory.build({ name: Cypress.env('offender_name'), crn: Cypress.env('offender_crn') })
+const keyWorkerName = Cypress.env('keyworker_name')
+const offenderName = Cypress.env('offender_name')
+const offenderCrn = Cypress.env('offender_crn')
+
+// eslint-disable-next-line no-unused-expressions
+expect(keyWorkerName, 'keyWorkerName was set').to.be.a('string').and.not.be.empty
```
pezholio, line 14: This might be more readable if we had a function to throw an error like:

```typescript
const throwMissingError(field: string) {
  throw new Error(`Missing Cypress env variable for '${field}'`)
}
```

And then we could assign and throw if nil in one line like so:

```typescript
const keyWorkerName = Cypress.env('keyworker_name') || throwMissingError('keyworker_name')
```

pezholio: (Note: we can't do `const keyWorkerName = Cypress.env('keyworker_name') || throw new Error(`Missing Cypress env variable for 'keyworker_name'`)` because `throw` is a statement, and can't (yet) be used as an operand to an operator.

e2e/tests/stepDefinitions/manage.ts
```diff
@@ -6,14 +6,36 @@ import bookingFactory from '../../../server/testutils/factories/booking'
...
+  // but the password value should not be shown
+  if (typeof password !== 'string' || !password) {
+    throw new Error('Missing password value, set using CYPRESS_password=...')
+  }
+
   cy.visit('/')
-  cy.get('input[name="username"]').type(Cypress.env('username'))
-  cy.get('input[name="password"]').type(Cypress.env('password'))
+  cy.get('input[name="username"]').type(username)
+  cy.get('input[name="password"]').type(password, { log: false })
```
pezholio, line 38: This is a good catch!

pezholio:
As far as the security context stuff goes, I'm not entirely sure I see the value, the secrets are set the the single repo, so it shouldn't be visible elsewhere. The only value I can see is if we're sharing secrets across an organisation. I could be missing something though!

pezholio:
@edavey There's still some conflicts here. Also an odd CircleCI error

pezholio (approved):

pezholio:
LGTM. Gonna merge this in and see if the CircleCI issue goes away on its own!

PR #209, opened 2022-09-29
Document step of copying environment variables for development

tahb:
# Context

# Changes in this PR

Without remembering to copy this file across as is the sign in step fails with a hmpps-auth error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,12 @@ If you're coming back to the application after a certain amount of time, you can
 script/bootstrap
 ```
 
+Copy the example environment variables across into a new .env file:
+
+```bash
+cp .env.example .env
+```
+
 ## Running the application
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,6 +23,12 @@ If you're coming back to the application after a certain amount of time, you can
 script/bootstrap
 ```
 
+Copy the example environment variables across into a new .env file:
+
+```bash
+cp .env.example .env
+```
+
```
pezholio, line 31: Could we add this to the `script/setup` process? Something like:

```bash
if [ ! -f .env ]; then
  echo "==> Creating env file"
  cp .env.example .env
fi
```

tahb: Ah it is already in the setup script, though not bootstrap as I expected 🤦‍♂️ I'll close.

PR #213, opened 2022-10-03
Temporary Accommodation "Add New Premises" spike

bryangaledxw:
# Context

* [Trello ticket](https://trello.com/c/oC1y8aIe/395-spike-adding-a-property-to-the-frontend)
* This is a draft PR being written up quickly so my absence today isn't blocking discussion on how to approach codebase sharing. It's not quite as tidy as I would like, particularly commit messages and where new wiremocking lives
* This PR adds several new environment variables. See `example.env`, and as a starting point, copy these new variables into your `.env` file
* This PR looks a lot scarier than it is. Most of the file changes are files being moved and the subsequent changes to import paths. The commits doing these moves are self-contained
* The value of this PR is to highlight the practical issues we run into using a shared codebase for AP (Approved Premises) and TA (Temporary Accommodation), what our decision points on how we share code are, and where we need to have agreement on the approach to take
* I'm very open to the conclusion being that this is the wrong approach
* There's possibly another question that I don't address in this PR - "what does TA look like to begin with, before any feature work?". Is it an empty service, and feature work is adding new functionality/pulling in existing AP functionality? Or is it a replica of AP, and feature work is adding new functionality/modifying or removing existing AP functionality?

# Changes in this PR

* All existing AP functionality is now accessibly locally at `http://localhost:3000/approved-premises`.
* There is a new page for adding a TA premises the you'll be redirected to by visiting `http://localhost:3000/temporary-accommodation`

# Walkthrough

I'd begun this spike with the intention of making the smallest change possible - functionality for determining which service is being accessed, some new paths and routes, and some small additions to existing controllers, services, and clients. But for each area, namespacing (generally splitting things into `approved-premises` and `temporary-accommodation` folders, then having TA-specific versions of a Premesis controller/service/client) rather than mingling became the tidier-feeling way to go.

## Paths

Since the path objects are fixed at compile-time, we either mingle our AP and TA paths, and lose any safety around which paths are for which service, or we namespace our paths. Having two distinct sets of paths also makes it easier to differ user-facing vocabulary where needed - the AP world uses "premises" but the TA world uses "properties"

## Routes

Currently, the router object is created on startup. Since at this point we don't know which service is being accessed, we need to create a router for AP and a router for TA, and then choose which we use in `server/app.ts` depending on the service being accessed.

Since we're creating these two router objects upfront, it becomes natural to namespace them. (Though an alternative would be to implement a `createRouter(serviceName)` method, depending on how much overlap we expect)

## @Types

We expect AP and TA schemas to differ - already our `Premises` differ. The alternative would be to share a `Premises` type, but have a lot of the fields be optional, which loses us a lot of type safety

## Controllers

I think from here down is where it gets more contentious whether we should be mingling or namespacing. Having done an initial implementation that just added a couple of methods to the existing Premises controller, it felt very *off*, to have additional methods that you just had to know only worked in the TA world but not the AP world.

If we continued that approach, I could see us needing a lot of commenting to distinguish what works for what service, and we'd end up needing a lot of `if (getService(req) === 'approved-premesies) ... else`, which I think would make testing awkward, having a lot of tests where we need to mock `getService`, and I can see things getting messy in cases where we have similar but diverging user flows, which I can easily see happening once we start getting user feedback in.

Assuming we're having distinct types, we'd also need to do things like: 

```
import type { Premises as TAPremises } from 'temporary-accommodation'
import type { Premises as APPremises } from 'approved-premises'
```

which feels very awkward

## Services

Much of the above also applies to services, particularly since they have some HTML-generating responsibility - that would be one of the places where I could see us needing to do `if (service === 'approved-premesies)` when we get a request for a UI tweak

## Clients

Again, some of the above also applies to Clients, particularly on type imports, and having to just know/comment what works for which service


richpjames:
Thanks for doing this Brian, it's really useful for me to see how the codebase might look with both AP/TA involved. 
Some high level thoughts: 
- I agree that we'd want to avoid wherever possible checks for which service is running in controllers & services. I also agree that we don't want to be using comments or knowledge-in-the-head to distinguish between which services are usable. I wonder if an approach is possible using inheritance but it's frankly beyond my capabilities and the additional complexity may outweigh the benefits that it would bring. 
- What benefits does combining these repos give us? if you take away what we get from the HMPPS TS template and namespace almost everything all I believe we're left with is some shared approaches and node modules. Therefore I think it's best to think of an alternative way of sharing these approaches/packages than having them in the same repo.
- I wonder if there is a way we could have the API unify the two services *just enough* for the two UIs to be able to share controller methods, services etc - obviously out of scope for this PR but potentially a worthwhile approach.

bryangaledxw:
Thanks for your feedback on this.

Yeah, on the one hand, this is where I ended up by following the changes I needed to make to eliminate some discomforts from my initial approach. On the other, with this approach, it's not much different to TA just starting a new repo with AP as the template.

I feel like this hinges on whether we end up sharing the Premises schema/endpoint on the backend - if we do, then I can see us using the same service/client classes (with maybe the odd flag if we need to differ in our HTML styling). I'd still lean towards keeping controllers distinct though (plus integration tests and views would also make sense to separate I think)

Basically I think the further away something gets from the user, the more sense it makes to share repos. If we're not even sharing at the API/schema level, then the sharing idea kinda falls apart. If we are sharing at the API/schema level, then we can share at least up to the client and service level

cypress_shared/pages/apply/list.ts
```diff
@@ -17,7 +17,7 @@ export default class ListPage extends Page {
   shouldShowApplicationSummaries(applicationSummaries: Array<ApplicationSummary>): void {
     applicationSummaries.forEach(summary => {
       cy.contains(summary.person.name)
-        .should('have.attr', 'href', `/applications/${summary.id}`)
+        .should('have.attr', 'href', paths.applications.show({ id: summary.id }))
```
pezholio, line 20: Great catch! 👍 Would you be OK putting this in a seperate PR?

pezholio:
Yeah, I agree 100% with what you're saying @bryangaledxw - the utility for this entirely depends on if we're going to be sharing the same data schema for a lot of this. If there are too many divergences, then the whole idea kinda falls apart a bit. I do wonder if maybe another approach could be maybe sharing clients, services and controllers maybe via an npm package? Then either of us are free to diverge as we see fit, and implement our own views etc. But I'm just spitballing here...

cypress_shared/pages/apply/list.ts
bryangaledxw: On it

bryangaledxw:
Thanks for the feedback, closing while I continue working on this

PR #221, opened 2022-10-04
(763) Improve application session handling

pezholio:
This started out as an attempt to add person details to an application, but as the work progressed, I realised there was a bunch of groundwork that needed to be put in first. This changes the flow of Apply to now add an Application object to the session, this touches a whole bunch of parts of the code, so there was a fair bit to do! 

I could have added the person details to this PR, but I thought it was better to get the groundwork in place so a) I don't block any more work and b) the PR doesn't get too big!

Hopefully it's easy enough to understand - let me know if there's anything I can do to make it more digestable!

## Caveats / things to note

At the moment, running Apply locally is a bit broken, because it's tricky to maintain state using Wiremock. I'm not entirely sure how to handle this, apart from running the whole stack locally. Thoughts welcome!

### DIFF
--- a/server/data/applicationClient.test.ts
+++ b/server/data/applicationClient.test.ts
@@ -43,6 +43,22 @@ describe('ApplicationClient', () => {
     })
   })
 
+  describe('find', () => {
+    it('should return an application', async () => {
+      const application = applicationFactory.build()
+
+      fakeApprovedPremisesApi
+        .get(paths.applications.show({ id: application.id }))
+        .matchHeader('authorization', `Bearer ${token}`)
+        .reply(200, application)
+
+      const result = await applicationClient.find(application.id)
+
+      expect(result).toEqual(application)
+      expect(nock.isDone()).toBeTruthy()
+    })
+  })
+
   describe('update', () => {
     it('should return an application when a PUT request is made', async () => {
       const application = applicationFactory.build()
--- a/server/data/applicationClient.test.ts
+++ b/server/data/applicationClient.test.ts
@@ -68,7 +68,7 @@ describe('ApplicationClient', () => {
         .matchHeader('authorization', `Bearer ${token}`)
         .reply(200, application)
 
-      const result = await applicationClient.update(application, application.id)
+      const result = await applicationClient.update(application)
 
       expect(result).toEqual(application)
       expect(nock.isDone()).toBeTruthy()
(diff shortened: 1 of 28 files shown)


server/services/applicationService.ts
```diff
@@ -75,9 +74,22 @@ export default class ApplicationService {
...
   }
 
-  private saveToSession(page: TasklistPage, request: Request) {
-    request.session = this.fetchOrInitializeSessionData(request.session, request.params.task, request.params.id)
-    request.session.application[request.params.id][request.params.task][request.params.page] = page.body
+  private getApplicationFromSessionOrAPI(request: Request): Promise<Application> {
+    const { application } = request.session
+
+    if (application && application.id === request.params.id) {
+      return application as unknown as Promise<Application>
```
richpjames, line 86: I feel uneasy about casting this to a `Promise` when it will be just an `Application` - I think it would be better to have a return type of `Promise<Application> | `Application`.
Also could we improve the type in the session `@types/express`?

server/services/applicationService.ts
```diff
@@ -75,9 +74,22 @@ export default class ApplicationService {
...
+    }
+    return this.findApplication(request.user.token, request.params.id)
```
richpjames, line 88: I think we should be awaiting this

server/data/applicationClient.test.ts
```diff
@@ -64,7 +64,7 @@ describe('ApplicationClient', () => {
       const application = applicationFactory.build()
 
       fakeApprovedPremisesApi
-        .put(paths.applications.update({ id: application.id }))
+        .put(paths.applications.update({ id: application.id }), JSON.stringify({ data: application.data }))
```
richpjames, line 67: Why do we have to stringify this and not any of our other client request bodies in the tests?

richpjames (dismissed):
Yeah this looks good on the whole. Have left some comments

server/services/applicationService.ts
pezholio: This gives me `Redundant use of 'await' on a return value`

server/services/applicationService.ts
pezholio: Good call 👍

pezholio: Fixed in https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/221/commits/61759942dc88db09739047adea3b73302245ddb3

server/services/applicationService.ts
richpjames: Oh yeah, I think you'd have to store is as a const first and then return the const (for reasons I don't understand). As long as we're awaiting the value returns from the function it will be fine

richpjames (approved):

PR #222, opened 2022-10-04
Align node versions between docker image and .node-version file

richpjames:
I noticed in the [CircleCI logs](https://app.circleci.com/pipelines/github/ministryofjustice/hmpps-approved-premises-ui/777/workflows/5524518d-9aa1-43be-8b82-8e7749d18725/jobs/1498/parallel-runs/0/steps/0-111) (line 35) we were getting errors that our Node version was incompatible with the one specified in `.node-version` so I have updated it.

18.0.0 wasn't available in docker so I've opted to move us to 18.10.0. This way we can use the same version in docker and other envs


pezholio (approved):
LGTM - thanks for spotting this!

PR #223, opened 2022-10-04
(763) Add person details to an application

pezholio:
Now we have a full application object stored in the session and passed to our pages, we can use the name of the person we are applying for in the page titles.

## Screenshots

Apply -- shows the details of a person from their CRN

Apply -- shows a tasklist


pezholio:
[Add person details to an application](https://trello.com/c/jeNUOIIM/763-add-person-details-to-an-application)

richpjames (approved):

PR #226, opened 2022-10-05
Add Renovate config

pezholio:
We currently use Dependabot for managing updates, but it seems that a lot of teams are migrating to Renovate now, so I thought we’d give it a go. We get a lot more control over what dependencies we can update, and can be a lot smarter about rebasing etc.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,7 +1,0 @@
-version: 2
-updates:
-  - package-ecosystem: 'npm'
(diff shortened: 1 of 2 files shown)


richpjames:
We should probably remove dependabot.yml at the same time as adding this

pezholio:
Tru dat. Lemme sort it

richpjames (approved):

PR #228, opened 2022-10-05
(767) Move key worker selection from booking to arrival form

pezholio:
The selection of a keyworker currently takes place on the booking form. In Delius this is done on the Arrival form, so in mini-manage we ought to be replicating this behaviour. The API has already changed this behaviour, so the contract tests are currently broken. We also now fetch the key workers from the appropriate endpoint, which returns staff for a specific premises.

## Screenshots

image

image

### DIFF
--- a/wiremock/bookingStubs.ts
+++ b/wiremock/bookingStubs.ts
@@ -14,10 +14,10 @@ bookingStubs.push({
         matchesJsonPath: "$.[?(@.crn != '')]",
       },
       {
-        matchesJsonPath: "$.[?(@.expectedArrivalDate != '')]",
+        matchesJsonPath: "$.[?(@.arrivalDate != '')]",
       },
       {
-        matchesJsonPath: "$.[?(@.expectedDepartureDate != '')]",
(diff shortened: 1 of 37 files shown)


wiremock/bookingStubs.ts
```diff
@@ -48,7 +45,7 @@ bookingStubs.push({
   },
 })
 
-const requiredFields = getCombinations(['crn', 'name', 'expectedArrivalDate', 'expectedDepartureDate', 'keyWorkerId'])
+const requiredFields = getCombinations(['crn', 'name', 'expectedArrivalDate', 'expectedDepartureDate'])
```
richpjames, line 48: Can we remove 'name' too as we get it from the CRN now?

richpjames (approved):
LGTM

wiremock/bookingStubs.ts
richpjames: Actually - whilst you're here - I think 'expectedArrivalDate' and 'expectedDepartureDate' should just be 'arrivalDate' and 'departureDate' to much the UI fields

richpjames (approved):

PR #229, opened 2022-10-05
Temporary Accommodation "Add New Premises" spike

bryangaledxw:
# Context

[This PR](https://trello.com/c/oC1y8aIe/395-spike-adding-a-property-to-the-frontend) demonstrates how we might expand the existing codebase to accommodation the Temporary Accommodation service, side-by-side with the Approved Premises service, sharing code where possible. This is a follow up to #213, and makes changes in response to feedback given.

Again, the PR looks scarier than it is - most of the files changes are in "Namespace paths", and most of those changes are simply moving files and fixing import paths

This PR assumes we want to share service and client classes between AP and TA, but not paths and controllers.

While still a spike and primarily to gather feedback on an approach, this is a reasonably sincere PR, i.e., I am submitting it with few known blockers to it actually being merged in. For the "real" PR, I would:
* Move the AP-specific contents of `server/controllers` to `server/controllers/approved-premises`
* Same as above for `server/views`, `cypress_shared/pages`, and `integration_tests/e2e`
* Move `server/types/approved-premises` to a service-agnostic location
* Split this into two PRs, the first ending at "Choose start page depending on service being accessed"

Again this PR adds several new environment variables. See `example.env`, and as a starting point, copy these new variables into your `.env` file

# Changes in this PR

* All existing AP functionality is now accessibly locally at `http://localhost:3000/approved-premises`
* There is a new page for adding a TA premises that you'll be redirected to by visiting `http://localhost:3000/temporary-accommodation`

## Screenshots of UI changes

localhost_3000_temporary-accommodation_properties_new


pezholio:
I think this is a much neater approach, so thanks for working on this! I do wonder if there's scope for having a standard set of controllers and then diverging where we need to?

bryangaledxw:
Closing to start on the "real" PR.

@pezholio My feeling on shared controllers is that we'll know what those are as we're doing feature work. So it might be that step one of a TA feature is to pull a controller from AP-world to shared-world

PR #230, opened 2022-10-05
Add sentry.io config

pezholio:
This will allow us to keep a handle on any errors in the application. I’ve also added an `ENVIRONMENT` variable, so we can track the environment the error has ocurred in. I’ve also added the `SENTRY_DSN` secret to the Kubernetes secrets.

### DIFF
--- a/server/config.ts
+++ b/server/config.ts
@@ -36,6 +36,10 @@ export interface ApiConfig {
 export default {
   https: production,
   staticResourceCacheDuration: 20,
+  environment: process.env.ENVIRONMENT || 'local',
+  sentry: {
+    dsn: get('SENTRY_DSN', null, requiredInProduction),
+  },
   redis: {
     host: get('REDIS_HOST', 'localhost', requiredInProduction),
     port: parseInt(process.env.REDIS_PORT, 10) || 6379,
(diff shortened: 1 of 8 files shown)


server/config.ts
```diff
@@ -36,6 +36,10 @@ export interface ApiConfig {
 export default {
   https: production,
   staticResourceCacheDuration: 20,
+  environment: process.env.ENVIRONMENT || 'local',
```
richpjames, line 39: I'm guessing this is a HMPPS/MoJ standard? It seems strange to use `ENVIRONMENT` when we use `NODE_ENV` elsewhere but happy to bow to superior knowledge

richpjames (approved):

server/config.ts
pezholio: Yeah, `NODE_ENV` will usually be `production` for production like envs, and `development` otherwise. This just allows us to differentiate between `dev`, `prod` and `pre-prod`

richpjames (approved):
Sorry, thought I'd approved this!

pezholio:
You did, but I needed to force push because Github Actions was having a moment

PR #231, opened 2022-10-06
Fix blank over capacity message on booking confirmation page

richpjames:
# Changes in this PR
When a user visited the Booking Confirmation page previously there was an empty info flash. 
This fixes the issue so if there is overcapacity messaging it shows correctly and if there isn't the flash doesn't show.

### Before
image

### After
image


pezholio (approved):
Ah, nice catch! 👍

PR #232, opened 2022-10-06
(740) Booking extension contract test

richpjames:
This PR adds integration tests for extending a booking following the prescedent set by #203 #173 #203 etc. 

In doing this I realised a couple of errors we had made:
- we were assuming the incorrect type being returned from the extension endpoint
- we were redirecting the booking extension id after creating a booking extension when we should have been redirecting to the *booking* id


pezholio (approved):
LGTM 👍

PR #234, opened 2022-10-10
Update Renovate config

richpjames:
I was getting concerned that we haven't had any Renovate updates since we enabled it (and disabled dependabot #226) so I took a second look at the config. 
It seems like we want to *enable* the first rule in the rule set as we want `npm` updates.
Also I think we should be updating minor versions as this should give us less headaches when the time comes to update to major versions.


pezholio:
Ooops - well spotted!

pezholio (approved):

PR #236, opened 2022-10-10
Allow Temporary Accommodation to be developed on top of Approved Premises

bryangaledxw:
# Context

[This PR](https://trello.com/c/oC1y8aIe/395-spike-adding-a-property-to-the-frontend) allows us to develop Temporary Accommodation service, side-by-side with the Approved Premises service, sharing code where possible.

This PR doesn't go as far as #229 with namespacing parts of the code where we want TA and AP to diverge, since we expect `temporary-accommodation-main` to be a long-running branch, and moving a lot of files would be make later rebases painful

This PR adds several new environment variables. See `example.env`, and as a starting point, copy these new variables into your `.env` file

# Changes in this PR

* All existing AP functionality is now accessibly locally at `http://localhost:3000/approved-premises`
* There is a new stub TA page at `http://localhost:3000/temporary-accommodation`

## Screenshots of UI changes

localhost_3000_


server/utils/applicationUtils.test.ts
```diff
@@ -32,4 +37,71 @@ describe('applicationUtils', () => {
       )
     })
   })
+
+  describe('getService', () => {
+    beforeAll(() => {
+      config.approvedPremisesRootPath = 'approved-premises-root-path'
+      config.temporaryAccommodationSubdomain = 'approved-premises-root-path'
```
tahb, line 44: This line could possibly be removed in favour of the same option set a few lines below: 

```
config.temporaryAccommodationSubdomain = 'temporary-accommodation-subdomain'
```

Which looks to have the right value assigned to it too.

server/utils/applicationUtils.ts
```diff
@@ -20,4 +22,33 @@ const taskLink = (task: TaskNames, id: string): string => {
   })}" aria-describedby="eligibility-${task}" data-cy-task-name="${task}">${taskLookup[task]}</a>`
 }
 
-export { getTaskStatus, taskLink }
+const getService = (req: Request): 'approved-premises' | 'temporary-accommodation' => {
```
tahb, line 25: Non-blocking suggestion: There's a fair bit of conditional logic going on in one method, there might be an opportunity to refactor this into some private methods. That said it's readable and unlikely to get changed much more. This approach may be more common in the Typescript world in which case I'll get back in my hole :)

server/utils/applicationUtils.test.ts
```diff
@@ -32,4 +37,71 @@ describe('applicationUtils', () => {
       )
     })
   })
+
+  describe('getService', () => {
```
tahb, line 41: Excellent to see the code and tests going into the same commit, love to see it 👍

server/routes/temporary-accommodation/index.ts
```diff
@@ -0,0 +1,18 @@
+/* istanbul ignore file */
```
tahb, line 1: Today I learned. An instruction to exclude this file from test coverage?

tahb (approved):
I've read through as best I can though lots is unfamiliar to me. I note that Stu has already had a look at the earlier draft ([229](https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/229)) you pointed towards and there wasn't any lower level feedback. I wonder if that's because it's all good or because Stu didn't have the time to give it more than a high level review? If it's the latter we might want to get a review from someone more useful than me. Given the convergence work here that affects AP I suggest it'd be worth confirming that with Stu as we want this to be good enough to be merged into `main` when ready?

In general though, as I said, I love seeing lots of test coverage and small commits that help me follow along. 

I'll approve from my end and let you have that final check with Stu before merging.

server/paths/api.ts
```diff
@@ -1,5 +1,32 @@
-import managePaths from './manage'
-import applyPaths from './apply'
+import { path } from 'static-path'
```
pezholio, line 1: I think this is a really good idea - combining application and API paths always make me a bit uneasy

server/controllers/index.ts
```diff
@@ -1,8 +1,8 @@
 /* istanbul ignore file */
 
 import ApplicationController from './applicationController'
-import { controllers as manageControllers } from './manage'
-import { controllers as applyControllers } from './apply'
+import { controllers as apManageControllers } from './manage'
+import { controllers as apApplyControllers } from './apply'
```
pezholio, line 5: I wondere if there's a more elegant way of namespacing the controllers? What I'd like to see is a single set of "standard" controllers in the base path, and then a way of overriding the standard controllers by adding a controller to a namespaced path, and the same with views. That said, I don't think this is a blocker given you're working on a feature branch, but something worth thinking about.

server/config.ts
```diff
@@ -83,4 +83,17 @@ export default {
     },
   },
   domain: get('INGRESS_URL', 'http://localhost:3000', requiredInProduction),
+  serviceSignifier: get<'path' | 'domain' | 'approved-premises-only' | 'temporary-accommodation-only'>(
+    'SERVICE_SIGNIFIER',
+    'path',
+    requiredInProduction,
+  ) as 'path' | 'domain' | 'approved-premises-only' | 'temporary-accommodation-only',
```
pezholio, line 90: I wonder if we should have default service signifiers, so locally we're using the path and then in 'prod-like' envs we're using the domain

server/routes/temporary-accommodation/index.ts
bryangaledxw: Seems to be - this is following the convention of the AP routes

server/utils/nunjucksSetup.ts
```diff
@@ -22,7 +22,18 @@ export default function nunjucksSetup(app: express.Express, path: pathModule.Pla
...
-  app.locals.applicationName = 'Approved Premises Ui'
+
+  app.use((req, _res, next) => {
+    const service = getService(req)
+
+    if (service === 'approved-premises') {
+      app.locals.applicationName = 'Approved Premises Ui'
+    } else {
+      app.locals.applicationName = 'Temporary Accommodation Ui'
+    }
```
pezholio, line 40: We should probably call the apps just `Approved Premises` or `Temporary Accommodation`. The UI distinction is irrelevant. I've been meaning to change this for ages, was just a default name that got put in when we generated the repo from the template

pezholio (approved):
I think this is all fine, nice and easy to follow. No blockers given this is going into a feature branch, but a few suggestions / things to think about

server/utils/applicationUtils.test.ts
bryangaledxw: Ah, this should be

```
config.temporaryAccommodation*RootPath* = 'temporary-accommodation-root-path'
```

Will fix

PR #237, opened 2022-10-11
Fix/docker container errors in dev

tahb:
# Context

# Changes in this PR

These changes are the ones I found were needed to get both apps running using docker-compose. This is a proposal that goes in tandem with a similar [pull request in the API](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/142).

We've got:

- removing container names when not needed to remove name clashes with the frontend
- an addition to the script/server to shut any old containers down when starting up. The script can leave things hanging when we term the process. We found maing changes to the docker-compose and restarting the server was clashing with existing containers running without this

With these changes and a similar set on the frontend both apps work together when running script/server which is nice. I'm aware you're experimenting with Tilt right now however I think these changes are still useful.

### DIFF
--- a/script/server
+++ b/script/server
@@ -7,6 +7,12 @@ set -e
 
 cd "$(dirname "$0")/.."
 
+cleanup() {
+  echo "==> Tearing down any old containers..."
+  docker-compose down
+}
+trap cleanup EXIT
+
 script/utils/launch-docker.sh
 
 script/utils/start-backing-services.sh
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,7 +5,6 @@ services:
     image: 'bitnami/redis:5.0'
     networks:
       - hmpps
-    container_name: redis
     environment:
       - ALLOW_EMPTY_PASSWORD=yes
     ports:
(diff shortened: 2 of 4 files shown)


script/server
```diff
@@ -7,6 +7,9 @@ set -e
 
 cd "$(dirname "$0")/.."
 
+echo "==> Tearing down any old containers..."
+docker-compose down
```
pezholio, line 11: Is it worth having this in a teardown function like so:

```bash
cleanup() {
  echo "==> Tearing down any old containers..."
  docker-compose down
}
```

Which we can then use when the script exits:

```bash
trap cleanup EXIT
```

pezholio (approved):
LGTM - just a teeny bash suggestion

script/server
tahb: Oo much better 👍

tahb:
@pezholio thanks for the suggestion i've made that change but it's invalidated your review, please can I have another one!

pezholio (dismissed):
LGTM 👍

tahb:
We've have added a new commit after noticing a problem when running with wiremock on and now need to distract you again for another review 😞

PR #241, opened 2022-10-11
Remove matching the @types/node package in Renovate config

richpjames:
We want to recieve updates for all our packages not just this specific one


pezholio:
Ooops! Nice catch!

pezholio (approved):

PR #242, opened 2022-10-13
Fix arrival types

pezholio:
This updates the Arrival form to send the expected values, before the `arrivalDate` was `date`  and the `keyWorkerStaffId` was not present. I've also updated the DateTime converter to only include the timestamp if a time is provided, The API only accepts dates in `YYYY-MM-DD` format, except for in the case of departures.


richpjames (approved):
Looks good, how did we only spot this now?

pezholio:
No contract tests! Have had it on my list to sort for ages!

PR #245, opened 2022-10-13
Update RestClient query type to allow a Record<string, string>

bryangaledxw:
This allows us to do:
```
restClient.get({ path: somePath, query: { foo: 'bar' } })
```

rather than:
```
restClient.get({ path: somePath, query: 'foo:bar' })
```


pezholio (approved):
LGTM 👍

PR #246, opened 2022-10-13
Add a clickBack helper method to our base Page class

bryangaledxw:
A back link is used in many of our templates, so worth including a helper in our base Page class


pezholio (approved):
LGTM - Thanks for committing this kind of stuff back 👍

PR #247, opened 2022-10-13
(750) Frontend for risk widgets

richpjames:
# Context
Applicants need to be able to see the risk widgets on the page
[Trello ticket](https://trello.com/c/2QdcWtlh/750-frontend-for-risk-widgets)

Helpfully these widgets already exist and we just have to pull them in with minor tweaks: https://github.com/ministryofjustice/hmpps-risk-ui-elements/

## Changes in this PR

As the risks relate to a Person we use the Person service and client to make the call to the API from the Applications controller.

### Before
image

### After
image

### DIFF
--- a/server/controllers/apply/applicationsController.ts
+++ b/server/controllers/apply/applicationsController.ts
@@ -27,14 +27,16 @@ export default class ApplicationsController {
   }
 
   show(): RequestHandler {
-    return (req: Request, res: Response, next: NextFunction) => {
+    return async (req: Request, res: Response, next: NextFunction) => {
       const { application } = req.session
 
-      if (!application) {
+      if (application) {
+        const risks = await this.personService.getPersonRisks(req.user.token, application.person.crn)
+
+        res.render('applications/show', { application, risks })
+      } else {
         next(createError(404, 'Not found'))
       }
-
-      res.render('applications/show', { application })
     }
   }
 
--- a/server/testutils/factories/risks.ts
+++ b/server/testutils/factories/risks.ts
@@ -1,0 +1,69 @@
+import { Factory } from 'fishery'
+import { faker } from '@faker-js/faker/locale/en_GB'
+
+import type {
+  PersonRisks,
+  RiskLevel,
+  RiskEnvelopeStatus,
+  TierLetter,
+  TierNumber,
+  RiskTierLevel,
+} from 'approved-premises'
+import { DateFormats } from '../../utils/dateUtils'
+
+const riskLevels: RiskLevel[] = ['Low', 'Medium', 'High', 'Very High']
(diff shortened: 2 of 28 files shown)


server/paths/apply.ts
```diff
@@ -10,6 +10,7 @@ const paths = {
     new: applicationsPath.path('new'),
     start: applicationsPath.path('start'),
     people: applicationsPath.path('people').path('find'),
+    personRisks: applicationsPath.path('people').path(':crn').path('risks'),
```
pezholio, line 13: Is this path used anywhere?

server/testutils/factories/risks.ts
```diff
@@ -0,0 +1,69 @@
...
+  TierLetter,
+  TierNumber,
+  RiskTierLevel,
+} from 'approved-premises'
+import { DateFormats } from '../../utils/dateUtils'
+
+const riskLevels: RiskLevel[] = ['Low', 'Medium', 'High', 'Very High']
+const riskEnvelopeStatuses: RiskEnvelopeStatus[] = ['retrieved', 'not_found', 'error']
+
+export default Factory.define<PersonRisks>(() => ({
```
pezholio, line 17: I like this a lot - nice way of getting realistic looking data 👍

server/utils/utils.ts
```diff
@@ -72,3 +73,23 @@ export const removeBlankSummaryListItems = (items: Array<SummaryListItem>): Arra
     return false
   })
 }
+
+export const mapApiPersonRiskForUi = (risks: PersonRisks): PersonRisksUI => {
```
pezholio, line 77: I see why you've not tested this, but it might be worth doing it, just so we're documenting the functionality

server/controllers/apply/applicationsController.ts
```diff
@@ -27,14 +27,16 @@ export default class ApplicationsController {
...
-    return (req: Request, res: Response, next: NextFunction) => {
+    return async (req: Request, res: Response, next: NextFunction) => {
       const { application } = req.session
 
       if (!application) {
         next(createError(404, 'Not found'))
       }
 
-      res.render('applications/show', { application })
+      const risks = await this.personService.getPersonRisks(req.user.token, application?.person.crn)
```
pezholio, line 37: Could we do:

```
if (application) {
  const risks = await this.personService.getPersonRisks(req.user.token, application?.person.crn)
  
  res.render('applications/show', { application, risks })
} else {
  next(createError(404, 'Not found'))
}
```

pezholio (approved):
Looks great! Really thorough. Just a couple of minor comments 👍

server/paths/apply.ts
richpjames: Good eyes! It was used in the test but not in the code.

server/utils/utils.ts
richpjames: Yeah no probs!

PR #251, opened 2022-10-17
Use extension of ARN Renovate config

richpjames:
[Renovate](https://github.com/renovatebot/renovate) has yet to issue any updates in the ~3 weeks we've had this config in place.
Various tweaks (#234 #241) haven't seem to have any effect so instead I propose extending our colleagues in the ARNS team's config in order to get some PRs coming through. If there are too many (or too few) updates we can tweak our settings from there but this should get us moving for now.


pezholio (approved):
Sounds like a plan 👍

PR #253, opened 2022-10-17
Allow convertObjectsToSelectOptions to take a prompt string

bryangaledxw:
Previously this was hardcoded to 'Select a keyworker', but we want to use this function in other contexts, so allow this string to be specified by the caller


pezholio (approved):
LGTM - ta!

PR #256, opened 2022-10-18
(629) Type of AP required screen: ESAP

pezholio:
This adds various screens for if a user chooses an ESAP AP. A couple of things have come to light here, including the fact that these are the first multi-question pages, so I've added a `questions` attribute to make sure there is one source of truth for question names. This is also the first time we're using the answers to previous questions for branching, so I've made an improvement to the `retrieveQuestionResponseFromApplication` function.

## Screenshots

image
image
image

### DIFF
--- a/server/@types/approved-premises/index.d.ts
+++ b/server/@types/approved-premises/index.d.ts
@@ -98,13 +98,19 @@ declare module 'approved-premises' {
     rows: Array<SummaryListItem>
   }
 
-  export interface RadioItems {
+  export interface RadioItem {
     text: string
     value: string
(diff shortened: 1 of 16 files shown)


pezholio:
[Type of AP required screen: ESAP](https://trello.com/c/dx0okaqb/629-type-of-ap-required-screen-esap)

richpjames (approved):
LGTM - is it worth extending the integration tests through these questions?

pezholio:
@richpjames I think so, but let's wait until we have a straight-through happy path

PR #259, opened 2022-10-18
[TA] (447) Creating a premises includes location data

bryangaledxw:
# Changes in this PR

* Frontend now makes a call to a `local-authorities` endpoint, rather than using a hardcoded list of local authorities and database IDs

### DIFF
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -12,6 +12,7 @@ import cancellation from './integration_tests/mockApis/cancellation'
 import lostBed from './integration_tests/mockApis/lostBed'
 import person from './integration_tests/mockApis/person'
 import applications from './integration_tests/mockApis/applications'
+import localAuthority from './integration_tests/mockApis/localAuthority'
 
 export default defineConfig({
   chromeWebSecurity: false,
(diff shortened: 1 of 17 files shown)


bryangaledxw:
[Creating a premise includes location data](https://trello.com/c/kAqCU7zb/447-creating-a-premise-includes-location-data)

tahb (commented):
Looks good.

The API half of this isn't ready and local e2e test seemingly fail with:

```
[Node]     text: '{"timestamp":"2022-10-18 16:56:19","status":404,"error":"Not Found","message":"No message available","path":"/local-authorities"}',
```

With Wiremock the form input pops up with local authorities. I was surprised to see some duplicate entries, is this because of existing API seeding or is there an issue with my environment?:

Screenshot 2022-10-18 at 18 00 11

bryangaledxw:
- The local authorities are just [faker](https://www.npmjs.com/package/@faker-js/faker) randomly choosing counties. I think duplicates are expected, Birthday Paradox and all that, and the sorting just makes the duplicates really obvious
- Yep, have been neglecting E2E tests - let's discuss in our catch-up today?
- Just pushed an update to the test name

tahb (approved):

PR #261, opened 2022-10-18
(801) Generate types from api spec

richpjames:
# Context
We want to be sure that we're sending the data to the API that it expects.
Copying what is in the [api spec](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/main/src/main/resources/static/api.yml) by eye is ineffective as it is prone to human error and the api spec can change.
A solution we have come up with is to generate the Typescript types that we use in the UI repo from OpenApi Spec that is used in the API repo. We can pull the latest version of these types whenever we run tests, this way we should always be in lock step with the API when we make changes in this repo. 

[Trello ticket](https://trello.com/c/VdgBBusD/801-implement-openapi-type-generation-in-ui-repo)

# Changes in this PR
- First we create a script to generate the typescript types from the openApi spec
- We then add the types we've generated to the repo's two tsconfigs
- We then add all the models/types to the repo that are generated by the script
- We then correct all the imports to use the new types
- Then we correct all the issues raised by consolidating the contract, most of these are minor

However one thing stands out:
The assumptions we made about the ApplicationSummary were way out, the view and controller will have to have significant changes made. In order to get things moving (and as these changes touch almost every file in the repo we need to move fast to avoid this branch becoming stale) I've opted for a quick fix for this and created [a trello card](https://trello.com/c/HcX0l05U/799-application-dashboard-pt-2) to address these issues later.


script/generate-types
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
```
pezholio, line 1: Not for now, but we should probably have [Shellcheck](https://www.shellcheck.net/) running as part of our CI pipeline, as we're using a lot of shell scripts now

integration_tests/tsconfig.json
```diff
@@ -3,12 +3,16 @@
...
+    "types": ["cypress"],
     "esModuleInterop": true,
     "typeRoots": ["../server/@types", "../node_modules/@types"],
     "resolveJsonModule": true,
-    "baseUrl": "./",
     "moduleResolution": "node",
+    "baseUrl": ".",
+    "paths": {
+      "@approved-premises-ui": ["../server/@types/ui/index.d.ts"],
+      "@approved-premises-shared": ["../server/@types/shared/index.d.ts"]
```
pezholio, line 14: Could we have `@approved-premises/ui` and `@approved-premises/api` or something?

pezholio: I deffo think there's scope for breaking this out into a module of its own at some point, especially as CAS3 will be forking at some point

pezholio: Will also mean that we can generate new types and schedule releases automatically. That way if there are any major changes to the API, we're alerted via Renovate

server/@types/shared/models/Booking.ts
```diff
@@ -0,0 +1,20 @@
...
+
+import type { Arrival } from './Arrival';
+import type { BookingBody } from './BookingBody';
+import type { Cancellation } from './Cancellation';
+import type { Departure } from './Departure';
+import type { Extension } from './Extension';
+import type { Nonarrival } from './Nonarrival';
+
+export type Booking = (BookingBody & {
+    status: 'arrived' | 'awaiting-arrival' | 'not-arrived' | 'departed' | 'cancelled';
```
pezholio, line 13: I'm really chuffed it generates union types from enums 👍

server/@types/ui/index.d.ts
```diff
@@ -0,0 +1,130 @@
...
+  checked?: boolean
+}
+
+export interface CheckBoxItem {
+  text: string
+  value: string
+  checked?: boolean
+}
+
+export interface SelectOption {
```
pezholio, line 45: Again, not for now, but it would be ace if we could break some of this out into a types library for GOV.UK frontend that we could publish via `DefinitelyTyped`. It's wild that this doesn't exist already

server/@types/ui/index.d.ts
```diff
@@ -0,0 +1,130 @@
...
+  text: string
+}
+
+export type RiskLevel = 'Low' | 'Medium' | 'High' | 'Very High'
+
+export type RiskEnvelopeStatus = 'retrieved' | 'not_found' | 'error'
+
+export type TierNumber = '1' | '2' | '3' | '4'
+export type TierLetter = 'A' | 'B' | 'C' | 'D'
+export type RiskTierLevel = `${TierLetter}${TierNumber}`
```
pezholio, line 85: Should these be in the models, or is the generator not smart enough to do this? If it's the latter, do they actually get used anywhere?

pezholio (approved):
Yes, yes! I ❤️ this! Thanks so much for taking the time to do this and persevering 🎖️ 

Just a couple of comments, but none of them deal breakers at all. Let's get this merged 👍

script/generate-types
richpjames: Yep good idea, I've made a note of it

server/@types/shared/models/Booking.ts
richpjames: Yeah it was an option for the package thankfully 😌

server/@types/ui/index.d.ts
richpjames: Yeah I was thinking this, it would be easy now we're using paths in the tsconfig

server/@types/ui/index.d.ts
richpjames: The RiskTierLevel and RiskLevel are currently used in the applicationSummary factory (soon to be RIP) and the risks factory. 
RiskEnvlopeStatus is a dupe of a type in the model so I'll remove

I think that the generator is only as smart as what we tell it - I think we just have `string` for the risk tier level.

server/@types/ui/index.d.ts
yndajas: That would be nice!

PR #269, opened 2022-10-19
(800) Update error handling in Apply forms

pezholio:
Currently, when we return errors from an Apply Page, we get them back in the same way we get them from an API call, in this format:

```
{
  propertyName: '$.secretingHistory',
  errorType: 'empty',
}
```

These then refer to a translation in the errors.json file, i.e

```
{
...
  "secretingHistory": { "empty": "Some error"},
...
}
```

This causes us two problems:

- We can't inject information about the application (at the moment there are placeholders like `You must specify why xxxx requires an enhanced security placement`
- This breaks our rule of the page classes knowing everything about the question / answer text

This updates the Apply page errors so they're simple key/value records with the text defined in the classes themselves. This allows us to inject information about the application directly into the errors, as well as keep information about the question(s) in one place.

There are a lot of commits, but most of them are updating the error types to the correct format. We also encountered some errors where types were being pulled in alongside other code by the integration tests, so the typechecks were failing, so I've addressed that too!

### DIFF
--- a/server/utils/validation.ts
+++ b/server/utils/validation.ts
@@ -18,9 +18,9 @@ export const catchValidationErrorOrPropogate = (
   redirectPath: string,
 ): void => {
   if ('data' in error) {
-    const invalidParams = error.data['invalid-params'] || error.data
-
-    const errors = generateErrors(invalidParams)
+    const errors = error.data['invalid-params']
+      ? generateErrors(error.data['invalid-params'])
+      : (error.data as Record<string, string>)
 
     const errorMessages = generateErrorMessages(errors)
     const errorSummary = generateErrorSummary(errors)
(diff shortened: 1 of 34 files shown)


pezholio:
[Update error handling in Apply forms](https://trello.com/c/ysIc4DUH/800-update-error-handling-in-apply-forms)

server/utils/validation.ts
```diff
@@ -18,9 +18,14 @@ export const catchValidationErrorOrPropogate = (
   redirectPath: string,
 ): void => {
   if ('data' in error) {
-    const invalidParams = error.data['invalid-params'] || error.data
+    let errors: Record<string, string>
```
richpjames, line 21: Up to you but I think we could do:
`const errors = error.data['invalid-params'] ? generateErrors(error.data['invalid-params']) : error.data as Record<string, string>`

richpjames (approved):
Looks great, thanks for perservering through the TS pain! 1 minor ignorable comment :)

server/utils/validation.ts
pezholio: Oooh, that's better. Big fan of one-liners

PR #270, opened 2022-10-19
Add shellcheck to CI

richpjames:
We have quite a lot of shell scripts now so it makes sense to lint them to ensure we're not doing anything we don't want to.

### DIFF
--- a/script/test
+++ b/script/test
@@ -4,6 +4,12 @@
 
 set -e
 
+echo "==> Checking shell scripts"
+
+node_modules/.bin/shellcheck ./script/**/*[^utils] ./script/utils/**
+
+echo "==> Typechecking the code"
+
 script/utils/launch-docker.sh
 
 echo "==> Starting the backing services in Docker..."
--- a/package-lock.json
+++ b/package-lock.json
@@ -43,6 +43,7 @@
         "passport-oauth2": "^1.6.1",
         "prom-client": "^14.1.0",
         "redis": "^4.3.1",
+        "shellcheck": "^1.1.0",
         "static-path": "^0.0.4",
         "superagent": "^8.0.2",
         "url-value-parser": "^2.2.0",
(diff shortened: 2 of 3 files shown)


pezholio:
Could we use the shellcheck npm package (https://www.npmjs.com/package/shellcheck) and use it as part of the `test` script, so we're running it locally too?

richpjames:
@pezholio it's a good idea but I can't get it working - we need to give shellcheck an argument with the files we want it to check. The readme suggests '**.*.sh' but this doesn't work I get: 
```**/*.sh: **/*.sh: openBinaryFile: does not exist (No such file or directory)```
and if it did I don't think it would do what we want because most of our scripts don't end in '*.sh'.
I've pushed what I have in case you want to have a go at working it out

**Comment by pezholio** [comment]
@richpjames After a bit of trial and error I found adding:

```
"shellcheck": "shellcheck ./script/**/*[^utils] ./script/utils/**"
```

To the package.json scripts and then running `npm run shellcheck` seems to work

(We need to add the `[^utils]` bit to the glob, because `shellcheck` doesn't automatically exclude directories)

**Comment by richpjames** [comment]
> @richpjames After a bit of trial and error I found adding:
> 
> ```
> "shellcheck": "shellcheck ./script/**/*[^utils] ./script/utils/**"
> ```
> 
> To the package.json scripts and then running `npm run shellcheck` seems to work
> 
> (We need to add the `[^utils]` bit to the glob, because `shellcheck` doesn't automatically exclude directories)

Ahh, thanks Stu! I was struggling to do 'include the contents of utils but not utils itself', will update the script

File: script/test [hunk 1]
```diff
@@ -4,6 +4,12 @@
 
 set -e
 
+echo "==> Checking shell scripts"
+
+shellcheck ./script/**/*[^utils] ./script/utils/**
```
pezholio, line 9: Does the npm package make `shellcheck` available globally then? I'd have thought we'd have to run it as an npm command?

richpjames: Given the description in [the usage section of the README](https://www.npmjs.com/package/shellcheck) I think it just makes it available globally but I've added the path to the binary just to be sure

pezholio (approved):

PR #271, opened 2022-10-19
[TA] (423) A user can view a list of properties (part 2)

bryangaledxw:
# Changes in this PR

* This brings the premises listing page in line with the wireframe designs

### Before
localhost_3000_temporary-accommodation_properties (3)

### After
localhost_3000_temporary-accommodation_properties (2)


bryangaledxw:
[A user can view a list of properties](https://trello.com/c/P6TEGW6m/423-a-user-can-view-a-list-of-properties)

pezholio:
[Update error handling in Apply forms](https://trello.com/c/ysIc4DUH/800-update-error-handling-in-apply-forms)

server/@types/ui/index.d.ts
```diff
@@ -164,3 +164,7 @@ export type LocalAuthority = {
   id: string
   name: string
 }
+
+export type Premises = ApiPremises & {
```
richpjames, line 163: I know it's a bit more work but I wonder if this should live in a CAS3-specific type file with its own namespace in the `paths` object in the `tsconfig.json`? 
It seems a bit awkward to have to know that if you're referring to a CAS3 Premises it's in the UI types and a CAS1 Premises it's the api types

server/views/temporary-accommodation/premises/index.njk
```diff
@@ -2,16 +2,16 @@
 {% from "govuk/components/table/macro.njk" import govukTable %}
 {% extends "../../partials/layout.njk" %}
 
-{% set pageTitle = applicationName + " - Properties" %}
+{% set pageTitle = applicationName + " - List of properties" %}
```
richpjames, line 6: Completely up to you but we have been passing through the pageHeading from the controller as a variable for methods we've written recently.
I find it helps me work out which view we're rendering without having to look in the template

richpjames (approved):
Looks good, I've left a couple of comments

server/@types/ui/index.d.ts
bryangaledxw: Ah, so this is an artefact of UI being ahead of API, which now that we've got a second API dev starting I'm hoping will fix itself. Basically as soon as `address` makes its way into API main, this goes away!

server/views/temporary-accommodation/premises/index.njk
bryangaledxw: Cool, cheers for the heads up

server/@types/ui/index.d.ts
richpjames: makes sense - thanks

PR #273, opened 2022-10-19
(776) Add 'Planned FM works required' option

richpjames:
We need to add an radio option for **Planned FM works required** on the lost bed radio list
[Trello card](https://trello.com/c/4fGUP8HV/776-add-an-extra-radio-button-to-the-lost-bed-radio-list)
This change is made in the DB with [this PR](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/164)

### DIFF
--- a/wiremock/stubs/lost-bed-reasons.json
+++ b/wiremock/stubs/lost-bed-reasons.json
@@ -18,5 +18,10 @@
     "id": "7f3eb6cd-b3fc-4a8d-b9ce-5959918ffbe8",
     "name": "Staff shortage",
     "isActive": true
+  },
+  {
+    "id": "1f2eb6cd-b3fc-4a8d-b9ce-5959918ffbe8",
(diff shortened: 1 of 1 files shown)


wiremock/stubs/lost-bed-reasons.json
```diff
@@ -18,5 +18,10 @@
     "id": "7f3eb6cd-b3fc-4a8d-b9ce-5959918ffbe8",
     "name": "Staff shortage",
     "isActive": true
+  },
+  {
+    "id": "1f2eb6cd-b3fc-4a8d-b9ce-5959918ffbe8",
+    "name": "Planned FM works required",
+    "isActive": true
```
pezholio, line 25: This should be a database change right? This will only update the stubs

richpjames: You're right, Joe will be making the DB change today

PR #276, opened 2022-10-20
(746) Generate task list automatically

pezholio:
# Context

At the moment, the task list has the layout copied from the prototype with the tasks we have added manually. As the project develops, we'll need to add more tasks and sections, and we need to make sure we can do this in a manageable way.

# Changes in this PR

This adds a `sections` object, which spells out what tasks are contained in what section, allowing us to check if a section is complete, as well as automatically output the section and tasks within in the Nunjucks template.

### Before

image

### After

image


pezholio:
[Task List](https://trello.com/c/wV3ywlJB/746-task-list)

richpjames (approved):
Looks good!

PR #277, opened 2022-10-20
Fix e2e

pezholio:
Currently e2e tests are failing because https://github.com/gunar/shellcheck/issues/26


richpjames (approved):
thank you!

PR #282, opened 2022-10-24
[TA] (417) A user can view information about a property

bryangaledxw:
# Changes in this PR

* Adds a page for viewing property details
* When creating a property, the user is now redirected to that property's details page

## Screenshots of UI changes

localhost_3000_temporary-accommodation_properties_3d9f4ad2-97f8-464d-ab79-e197e0f9e0aa

### DIFF
--- a/server/services/premisesService.test.ts
+++ b/server/services/premisesService.test.ts
@@ -222,7 +222,7 @@ describe('PremisesService', () => {
     })
   })
 
-  describe('getPremisesDetails', () => {
+  describe('getApprovedPremisesPremisesDetails', () => {
     it('returns a title and a summary list for a given Premises ID', async () => {
       const premises = premisesFactory.build({
         name: 'Test',
@@ -233,7 +233,7 @@ describe('PremisesService', () => {
       })
       premisesClient.find.mockResolvedValue(premises)
 
-      const result = await service.getPremisesDetails(token, premises.id)
+      const result = await service.getApprovedPremisesPremisesDetails(token, premises.id)
 
       expect(result).toEqual({
         name: 'Test',
--- a/server/services/premisesService.test.ts
+++ b/server/services/premisesService.test.ts
@@ -1,15 +1,18 @@
 import PremisesService from './premisesService'
 import PremisesClient from '../data/premisesClient'
 import premisesFactory from '../testutils/factories/premises'
+import localAuthorityFactory from '../testutils/factories/localAuthority'
 import dateCapacityFactory from '../testutils/factories/dateCapacity'
 import staffMemberFactory from '../testutils/factories/staffMember'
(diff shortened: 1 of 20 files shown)


server/utils/viewUtils.ts
```diff
@@ -0,0 +1,7 @@
+import * as nunjucks from 'nunjucks'
+
+export const escape = (text: string): string => {
```
richpjames, line 3: Is this different from using the [safe flag](https://mozilla.github.io/nunjucks/templating.html#safe)?

server/services/premisesService.test.ts
```diff
@@ -266,6 +269,73 @@ describe('PremisesService', () => {
...
+              value: { html: 'Some notes' },
+            },
+          ],
+        },
+      })
+
+      expect(premisesClient.find).toHaveBeenCalledWith(premises.id)
+
+      expect(premisesClientFactory).toHaveBeenCalledWith(token)
+      expect(premisesClient.find).toHaveBeenCalledWith(premises.id)
```
richpjames, line 331: this is a dupe of line 328

server/services/premisesService.ts
```diff
@@ -131,6 +143,41 @@ export default class PremisesService {
...
+  private async temporaryAccommodationSummaryListForPremises(premises: Premises): Promise<SummaryList> {
+    return {
+      rows: [
+        {
+          key: this.textValue('Property name'),
+          value: this.textValue(premises.name),
+        },
+        {
+          key: this.textValue('Address'),
+          value: this.htmlValue(`${escape(premises.address)}<br />${escape(premises.postcode)}`),
```
richpjames, line 155: I may be missing something here but couldn't we do `value.html: this.htmlValue(...` and avoid having to add the `escape` function? https://design-system.service.gov.uk/components/summary-list/#options-summary-list-example--rows

richpjames (approved):
Looks good, I've left a couple of comments

server/utils/viewUtils.ts
bryangaledxw: Yep, this is kind of the mirror of the safe flag. Before we can flag something as safe, we need to actually make it safe

My mental model is that in cases where you pass html into GDS Nunjucks components, that HTML is rendered with the safe flag, so that it doesn't get escaped, so we need to make sure we *do* escape anything we don't want rendered as raw HTML

server/services/premisesService.ts
bryangaledxw: We want the `<br />` interpreted as HTML, but we don't want anything inside `address` or `postcode` interpretted as HTTML, hence we need to escape the latter but not the former

server/utils/viewUtils.ts
richpjames: Thanks - that makes sense

PR #285, opened 2022-10-25
Send X-SERVICE-NAME header

bryangaledxw:
# Context

The Temporary Accommodation UI service has multiple places where it needs to indicate to the API that it is TA making the request rather than Approved Premises. We had previously done this with a query parameter that we'd use on a call-by-call basis, but we'd like a more generic way of doing this. Following a team discussion, we agreed to send the service name as an HTTP header, with the value coming from an environment variable

As we are using an environment variable, this PR assumes that Temporary Accommodation UI will soon move to its own environment

AP developers should set the environment variable `COMMUNITY_ACCOMMODATION_API_SERVICE_NAME=approved-premises`. TA developers should set the environment variable `COMMUNITY_ACCOMMODATION_API_SERVICE_NAME=temporary-accommodation`


server/config.ts
```diff
@@ -80,6 +81,7 @@ export default {
         deadline: 10000,
       },
       agent: new AgentConfig(10000),
+      serviceName: get('APPROVED_PREMISES_API_SERVICE_NAME', 'approved-premises', requiredInProduction),
```
tahb, line 84: Does this set a default of `approved-premises` if the environment variable is undefined? If so, line 34 might not ever be undefined in which case we could remove the optionality.

edavey:
Would `COMMUNITY_ACCOMODATION_API_SERVICE_NAME=approved-premises|temporary-accommodation` be clearer? 

Given our remit is to build an API which serves the needs of accommodation services in general? CAS1 and CAS3 now and potentially CAS2 later?

Or, to put it another way, "CAS" is not desirable because it's an acronym but the term "Community Accommodation" describes the overall problem space clearly and accurately?

server/data/restClient.test.ts
```diff
@@ -24,15 +31,41 @@ describe('restClient', () => {
...
+      fakeApprovedPremisesApi.get(`/some/path`).reply(200, { some: 'data' })
+
+      const result = await restClient.get({ path: '/some/path' })
+
+      expect(result).toEqual({ some: 'data' })
+      expect(nock.isDone()).toBeTruthy()
+    })
+
+    it('should omit the X-SERVICE-NAME header when the configuration does not include a service name', async () => {
+      config.apis.approvedPremises.serviceName = undefined
```
tahb, line 45: When we stub our boundary here we lose a small bit coverage where the system extracts the environment variable and any potential default logic. 

In Rails we have a [gem](https://github.com/thoughtbot/climate_control) that lets us stub the environment variables. I'm not sure if that's possible or conventional in Typescript but just an optional idea incase it is.

server/data/restClient.test.ts
```diff
@@ -11,7 +11,14 @@ describe('restClient', () => {
 
   beforeEach(() => {
     config.apis.approvedPremises.url = 'http://localhost:8080'
-    fakeApprovedPremisesApi = nock(config.apis.approvedPremises.url)
+    config.apis.approvedPremises.serviceName = 'approved-premises'
+
+    fakeApprovedPremisesApi = nock(config.apis.approvedPremises.url, {
```
tahb, line 16: Is `nock` doing some magic here so that if the bearer token is not set, it'll go bang and let us know? That'd make me confident in the tidying up we do to lines 68 and 82.

tahb (approved):

server/config.ts
bryangaledxw: There's multiple APIs - approved-premises, hmppsAuth, and tokenVerification, and we only define `serviceName` for approved-premises, hence it being optional

bryangaledxw:
@edavey Right now I'm going for consistency (e.g., the config for this API is under `approvedPremises`, we have the `APPROVED_PREMISES_API_URL` environment var already). My feeling is that there's a PR at some point to change all the places we're using "Approved Premises" to refer to the more general service, but that isn't this PR

server/data/restClient.test.ts
bryangaledxw: Yep - the new `reqheaders` config at line 17

server/data/restClient.test.ts
bryangaledxw: This actually pushed me in the other direction of removing the dependency on environment variables entirely from this test!

Particularly as it was modifying env vars, it felt like it was breaking the self-containedness of our test, and it isn't `restClient.test.ts`'s role to be testing env var parsing

bryangaledxw:
@tahb @edavey I've updated this PR (it's now using the ServiceName type) and replied to comments. Is this fine to go in? My inclination is still not to change the naming in this PR, right now the API is still called the Approved Premises API, but Ed I'm not sure how strongly you feel on that

richpjames (approved):

bryangaledxw:
@edavey Discussed with Tom, renaming and accepting inconsistency

PR #287, opened 2022-10-25
[TA] Remove "service" from premises creation and listing API calls

bryangaledxw:
# Context

See #285 

As we will be sending the service name as part of the header once #285 is merged in, we no longer need to send the `service` field as part of premises creation or the `service` query parameter when getting the list of premises

This PR is opened as a draft, to be finalised when #285 is merged in, and `service` is removed from the `NewPremises` type. This will allow us to remove `service: undefined` from the `NewPremises` factory in `newPremises.ts`, which is something of a bodge

Note this PR does not remove `SERVICE_SIGNIFIER` etc. configuration, which can safely be removed once TA is in its own fork and running in its own environment

### DIFF
--- a/integration_tests/mockApis/premises.ts
+++ b/integration_tests/mockApis/premises.ts
@@ -1,29 +1,23 @@
 import type { Response, SuperAgentRequest } from 'superagent'
 
 import type { Premises, Booking, DateCapacity, StaffMember } from '@approved-premises/api'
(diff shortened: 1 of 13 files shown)


bryangaledxw:
[Creation of a property stores the service name from a header](https://trello.com/c/36NijLkN/485-creation-of-a-property-stores-the-service-name-from-a-header)

PR #294, opened 2022-10-27
Change all instances of Premises to ApprovedPremises

richpjames:
As of [this PR](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/183) The API splits the Premises response body for AP/TA into two different types. This means what we previously referred to as ‘Premises’ are now ‘ApprovedPremises’ (this enables TA to use ‘TemporaryAccomodationPremises’)

I have pointed the type generator to the PR version of the api.spec so this PR is a draft for now. I will revert the change to the typegen script before merging.


brownbearsoftware (approved):
LGTM

PR #297, opened 2022-10-31
Add API service name env var

richpjames:
Due to #285 we need to add the following env variable to the dev environment


brownbearsoftware (approved):
LGTM

PR #290, opened 2022-10-26
(632) - Apply - Risk management features

richpjames:
# Context
Another page for the Apply form similar to #281 
[Trello ticket](https://trello.com/c/iDR0FgPE/632-risk-management-features)

### After
Apply -- shows a tasklist (1)

PR #281, opened 2022-10-24
(630) Add Purpose of placement application page

richpjames:
# Context
This adds a 'Purpose of AP placement' screen to the Apply form in the pattern establish in #142 

[Trello card](https://trello.com/c/ketHjqUU/630-purpose-of-ap-placement)
# Changes in this PR

### Before
N/A
### After
image

PR #299, opened 2022-10-31
Fix - Apply - Errors bugs

richpjames:
I spotted a bug in the applicationService with how we were checking for errors. 

Previously we were checking for the `length` method to be truthy on the errors object. This would always be falsy as we don’t define a `length` on the errors object.
I assume that this was an array but at some point but it changed to being an object. TS didn’t highlight this error because the check for ‘.length’ on a `Record<string, string>` type is valid.

Correcting this made the error handling on Apply work properly which highlighted some issues with #290 #281 which I have fixed in this PR.


pezholio (approved):
LGTM - good catch! 👍

PR #307, opened 2022-11-01
(637) - Apply - Rehabilitative interventions

richpjames:
# Context
This PR adds a page to the Apply form for Rehabilitative Interventions. 
This is one of the more interesting Apply pages for a couple of reasons.
[Trello ticket](https://trello.com/c/edxP0Guq/637-purposeful-activities-now-called-rehabilitative-intervention-screen)

# Changes in this PR
1. I believe this is the final page of the 'Risk Management Features' task. It is the point at which the paths that split at the 'ConvictedOffences' page rejoin. All users will see this page but the previous page will be different depending on the different paths that have been followed. For this reason I have taken some time to correct the user flows through this task.
2. Removing 'Other' from `interventions`: We want to keep as much of the copy as possible in the rehabilitativeInterventions.ts file so it can be unit tested in order for changes to break tests so that we remember to update copy when things change. However the ‘other’ checkbox needs to trigger a textbox to show which means it needs to use a variable which can only be in nunjucks land. We need the translation to be in TS though as that’s where our translations live so I have opted to have 'Other' in the object on line 7 of `rehabilitativeInterventions.ts` so we can use it for the Interventions type and responses translations but we need to remove it in the `items` method. It’s not ideal but hopefully it’s clear enough what’s going on. 
An alternative option is to add in 'Other' when doing the translation - let me know if this would be preferable.
3. Error handling: This page is notable as we currently believe the user does not have to select any intervention to be able to proceed so the `errors` method simply returns an empty object if no interventions are selected. However we do want to check that an ‘other’ intervention is specified if the user chooses that option so we have to test whether the input is ‘other’ or is an array including ‘other’.

## Screenshots of UI changes
Apply -- shows a tasklist

### DIFF
--- a/server/form-pages/apply/risk-management-features/rehabilitativeInterventions.ts
+++ b/server/form-pages/apply/risk-management-features/rehabilitativeInterventions.ts
@@ -62,6 +62,14 @@ export default class RehabilitativeInterventions implements TasklistPage {
     return { ...response, 'Other intervention': this.body.otherIntervention }
   }
 
+  errors() {
+    const errors: TaskListErrors<this> = {}
+    if (this.otherInterventionDetailIsNeeded(this.body.interventions) && !this.body.otherIntervention) {
+      errors.otherIntervention = 'You must specify the other intervention'
+    }
+    return errors
+  }
+
   items() {
     const { other, ...uiInterventions } = interventionsTranslations
     return convertKeyValuePairToCheckBoxItems(uiInterventions, this.body.interventions)
--- a/server/form-pages/apply/risk-management-features/rehabilitativeInterventions.ts
+++ b/server/form-pages/apply/risk-management-features/rehabilitativeInterventions.ts
@@ -1,0 +1,73 @@
+import type { TaskListErrors } from '@approved-premises/ui'
+import { Application } from '../../../@types/shared'
+import { convertKeyValuePairToCheckBoxItems } from '../../../utils/formUtils'
+
+import TasklistPage from '../../tasklistPage'
+
+export const interventionsTranslations = {
+  accomodation: 'Accomodation',
+  drugsAndAlcohol: 'Drugs and alcohol',
+  childrenAndFamilies: 'Children and families',
+  health: 'Health',
+  educationTrainingAndEmployment: 'Education, training and employment',
+  financeBenefitsAndDebt: 'Finance, benefits and debt',
+  attitudesAndBehaviour: 'Attitudes, thinking and behaviour',
+  abuse: 'Abuse',
+  other: 'Other',
+} as const
+
+type Interventions = Array<keyof typeof interventionsTranslations>
+
+export default class RehabilitativeInterventions implements TasklistPage {
+  name = 'rehabilitative-interventions'
+
+  title = `Which rehabilitative interventions will support the person's Approved Premises (AP) placement?`
+
+  body: { interventions: Interventions; otherIntervention?: string }
+
+  constructor(
+    body: Record<string, unknown>,
+    private readonly _application: Application,
+    private readonly previousPage: string,
+  ) {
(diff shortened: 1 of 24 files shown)


server/form-pages/apply/risk-management-features/rehabilitativeInterventions.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  next() {
+    return ''
+  }
+
+  response() {
+    return {
+      [this.title]: {
+        Interventions: Array.isArray(this.body.interventions)
```
pezholio, line 54: I'm struggling to see here where `this.body.interventions` will ever not be an array? If there's a condition where it will be, would it perhaps be better if we put `body.interventions` into and array and flattened it, to ensure we always have an array, instead of having a complex conditional?

server/form-pages/apply/risk-management-features/rehabilitativeInterventions.ts
```diff
@@ -0,0 +1,76 @@
...
+    return this.previousPage
+  }
+
+  next() {
+    return ''
+  }
+
+  response() {
+    return {
+      [this.title]: {
```
pezholio, line 53: Should we be including `otherIntervention` in this response too?

pezholio (commented):
Just a couple of questions. I think we're missing `otherIntervention` from the response too. Though I could be missing something!

server/form-pages/apply/risk-management-features/rehabilitativeInterventions.ts
richpjames: If the user ticks one checkbox the response will be just that string. But good shout: we should put that into an array to save messiness elsewhere

server/form-pages/apply/risk-management-features/rehabilitativeInterventions.ts
richpjames: Yep my bad

server/form-pages/apply/risk-management-features/rehabilitativeInterventions.ts
```diff
@@ -0,0 +1,82 @@
...
+  body: { interventions: Interventions; otherIntervention?: string }
+
+  constructor(
+    body: Record<string, unknown>,
+    private readonly _application: Application,
+    private readonly previousPage: string,
+  ) {
+    const interventions = Array.isArray(body.interventions)
+      ? (body.interventions as Interventions)
+      : ([body.interventions] as Interventions)
```
pezholio, line 35: Could we not do this:

```suggestion
    const interventions = [body.interventions].flat()
```

server/form-pages/apply/risk-management-features/rehabilitativeInterventions.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  next() {
+    return ''
+  }
+
+  response() {
+    return this.otherInterventionDetailIsNeeded(this.body.interventions)
+      ? {
+          [this.title]: {
+            Interventions: this.body.interventions
```
pezholio, line 59: `Interventions` here feels redundant. These should probably sit under `Which rehabilitative interventions will support the person's Approved Premises (AP) placement?`

pezholio (approved):
LGTM - just a couple of suggestions

server/form-pages/apply/risk-management-features/rehabilitativeInterventions.ts
richpjames: Yes, much better!

server/form-pages/apply/risk-management-features/rehabilitativeInterventions.ts
richpjames: Yeah good point

PR #315, opened 2022-11-04
Add generator for form pages

pezholio:
This add a generator task that will allow us to generate all the necessary scaffolding for creating a new form page for Apply. It can be run like so:

```bash
npm run form-page:generate apply/basic-information/page
```

It's very scrappy and procedural, but will do the job for us. It generates a skeleton page class, a test file, a view, and a page object for use in Cypress and adds `TODO` comments for all the things that need adding.

The only thing it doesn’t do is add a new section to the imports, but this requires manual work anyway. It does add a `TODO` as a reminder to do it manually.

### DIFF
--- a/server/utils/utils.ts
+++ b/server/utils/utils.ts
@@ -44,6 +44,25 @@ const kebabCase = (string: string) =>
     .toLowerCase()
 
 /**
+ * Converts a string from any case to camelCase
+ * @param string string to be converted.
+ * @returns name converted to camelCase.
+ */
+export const camelCase = (string: string) =>
+  string
+    .toLowerCase()
+    .replace(/[-_]+/g, ' ')
+    .replace(/[^\w\s]/g, '')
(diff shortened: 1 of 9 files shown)


server/utils/utils.ts
```diff
@@ -43,6 +43,25 @@ const kebabCase = (string: string) =>
...
+    .replace(/[-_]+/g, ' ')
+    .replace(/[^\w\s]/g, '')
+    .replace(/\s+(.)(\w*)/g, (_$1, $2, $3) => `${$2.toUpperCase() + $3}`)
+
+/**
+ * Converts a string from any case to PascalCase
+ * @param string string to be converted.
+ * @returns name converted to PascalCase.
+ */
+export const pascalCase = (string: string) => camelCase(string).replace(/\w/, s => s.toUpperCase())
```
richpjames, line 63: I would've used the `case` package for these in order to avoid unit testing them but there's a good argument for not using it too

richpjames (approved):

PR #321, opened 2022-11-07
Update the docker-compose ports for test

pezholio:
This allows us to keep the dev application running and run the tests too without having to take any containers down


brownbearsoftware (approved):
LGTM

PR #310, opened 2022-11-03
(638) Location factors screen

pezholio:
This adds the screen to cover location factors. We've agreed with Dan that we'll ask for postcode areas initially, as these are easier to geocode. If someone answers "Yes" to the question about other areas, we ask what radius they would be willing to accept. This might need some fine tuning, but is OK for now.

## Screenshot

image

PR #323, opened 2022-11-07
(639) - Apply - PDU Transfer screen

richpjames:
This PR adds a screen asking the user whether a PDU transfer has been agreed. This only shows if the user submits yes to the 'different PDU' question on the Location Factors screen #310 
In the pattern established in https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/142
[Trello card](https://trello.com/b/ps9bUN4V/approved-premises-team-board)

## Screenshots of UI changes

Apply -- shows a tasklist

### DIFF
--- a/integration_tests/tests/apply/apply.cy.ts
+++ b/integration_tests/tests/apply/apply.cy.ts
@@ -12,6 +12,7 @@ import {
 } from '../../../cypress_shared/pages/apply'
 import ConvictedOffences from '../../../cypress_shared/pages/apply/convictedOffences'
 import DateOfOffence from '../../../cypress_shared/pages/apply/dateOfOffence'
+import PduTransferPage from '../../../cypress_shared/pages/apply/pduTransfer'
 import PlacementPurposePage from '../../../cypress_shared/pages/apply/placementPurpose'
 import RehabilitativeInterventions from '../../../cypress_shared/pages/apply/rehabilitativeInterventions'
 import RiskManagementFeatures from '../../../cypress_shared/pages/apply/riskManagementFeatures'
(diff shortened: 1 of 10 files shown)


server/form-pages/apply/location-factors/pduTransfer.ts
```diff
@@ -0,0 +1,54 @@
+import type { TaskListErrors } from '@approved-premises/ui'
+import { Application } from '../../../@types/shared'
+
+import TasklistPage from '../../tasklistPage'
+
+type Response = 'yes' | 'noNeedToMakeArrangements' | 'noProbationPractitioner'
```
pezholio, line 6: To be more DRY and to match how we've done it elsewhere, it might be worth structuring the questions and responses like so:

```suggestion
const questions = {
    yes: 'Yes',
    noNeedToMakeArrangements: 'No, I still need to make arrangements',
    noProbationPractitioner: 'No, management and supervision will be maintained by the existing Probation Practitioner',
} as const

const responses = Object.keys(questions)

type Response = keyof typeof responses
```

integration_tests/tests/apply/apply.cy.ts
```diff
@@ -12,6 +12,7 @@ import {
 } from '../../../cypress_shared/pages/apply'
```
pezholio, line 12: Not related to this file, but a `.DS_Store` has snuck into this commit

server/views/applications/pages/location-factors/pdu-transfer.njk
```diff
@@ -0,0 +1,48 @@
...
+            text: page.title,
+            isPageHeading: true,
+            classes: "govuk-fieldset__legend--l"
+          }
+        },
+        items: [
+          {
+            value: "yes",
+            text: "Yes",
+            checked: false,
```
pezholio, line 31: Should this be here?

pezholio (approved):
LGTM - just a couple of tiny things

PR #332, opened 2022-11-09
(645) Apply - Further Considerations: Vulnerability screen

pezholio:
This adds the Vulnerability screen to the Further Considerations step. There's a bunch of commonalities from the previous screen, so I've pulled some of the stuff around Yes/No answers that ask for more detail.

## Screenshot

image

### DIFF
--- a/server/form-pages/apply/further-considerations/roomSharing.test.ts
+++ b/server/form-pages/apply/further-considerations/roomSharing.test.ts
@@ -24,7 +24,7 @@ describe('RoomSharing', () => {
     })
   })
 
-  itShouldHaveNextValue(new RoomSharing({}), '')
+  itShouldHaveNextValue(new RoomSharing({}), 'vulnerability')
   itShouldHavePreviousValue(new RoomSharing({}), '')
 
   describe('errors', () => {
--- a/server/views/applications/pages/further-considerations/vulnerability.njk
+++ b/server/views/applications/pages/further-considerations/vulnerability.njk
@@ -1,0 +1,15 @@
+{% extends "../layout.njk" %}
+
+{% block questions %}
+  <h1 class="govuk-heading-l">
+    <span class="govuk-caption-l">{{ page.title }}</span>
+    Further placement considerations
+  </h1>
+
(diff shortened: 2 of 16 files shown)


pezholio:
[Vulnerability screen](https://trello.com/c/p3ScW8Ea/645-vulnerability-screen)

server/form-pages/apply/further-considerations/roomSharing.test.ts
```diff
@@ -24,7 +24,7 @@ describe('RoomSharing', () => {
     })
   })
 
-  itShouldHaveNextValue(new RoomSharing({}), '')
+  itShouldHaveNextValue(new RoomSharing({}), 'vulnerability')
```
richpjames, line 27: I think this has snuck in from another commit

server/views/applications/pages/further-considerations/vulnerability.njk
```diff
@@ -0,0 +1,15 @@
+{% extends "../layout.njk" %}
+
+{% block questions %}
+  <h1 class="govuk-heading-l">
```
richpjames, line 4: I think semantically this would make more sense as a `<h1>` and a `<h2>`. 
To me they're two levels of heading but happy to bow to your wisdom if you think otherwise 🙇‍♂️

richpjames (approved):
LGTM

server/views/applications/pages/further-considerations/vulnerability.njk
pezholio: I kinda agree with you, but this is an [established pattern](https://design-system.service.gov.uk/styles/typography/#headings-with-captions), so I'd rather not deviate from it. We can always tweak later.

server/form-pages/apply/further-considerations/roomSharing.test.ts
pezholio: Good catch!

server/views/applications/pages/further-considerations/vulnerability.njk
richpjames: Well, every day is a school day!

PR #347, opened 2022-11-10
(650) Apply: Relocation region screen

richpjames:
# Changes in this PR
Another Apply screen in the pattern established in #142 
This one is just a postcode input

## Screenshots of UI changes
Apply -- allows completion of the form (1)

### DIFF
--- a/cypress_shared/pages/apply/relocationRegion.ts
+++ b/cypress_shared/pages/apply/relocationRegion.ts
@@ -1,0 +1,13 @@
+import { Person } from '../../../server/@types/shared'
+import Page from '../page'
+
+export default class RelocationRegionPage extends Page {
+  constructor(person: Person) {
+    super('Placement duration and move on')
+    cy.get('.govuk-form-group').contains(`Where is ${person.name} most likely to live when they move on from the AP?`)
+  }
+
+  completeForm() {
+    this.getTextInputByIdAndEnterDetails('postcodeArea', 'XX1')
+  }
+}
--- a/server/form-pages/apply/move-on/placementDuration.test.ts
+++ b/server/form-pages/apply/move-on/placementDuration.test.ts
@@ -109,7 +109,7 @@ describe('PlacementDuration', () => {
     }
     application = applicationFactory.build({ data })
 
-    itShouldHaveNextValue(new PlacementDuration({}, application), '')
+    itShouldHaveNextValue(new PlacementDuration({}, application), 'relocation-region')
     itShouldHavePreviousValue(new PlacementDuration({}, application), '')
   })
 
--- a/server/utils/formUtils.test.ts
+++ b/server/utils/formUtils.test.ts
@@ -7,6 +7,7 @@ import {
   convertKeyValuePairToRadioItems,
   convertObjectsToSelectOptions,
   convertKeyValuePairToCheckBoxItems,
+  validPostcodeArea,
 } from './formUtils'
 
 describe('formUtils', () => {
@@ -308,6 +309,15 @@ describe('formUtils', () => {
           selected: false,
         },
       ])
+    })
+  })
+  describe('validPostcodeArea', () => {
+    it('when passed a postcode area it returns true', () => {
+      expect(validPostcodeArea('HR1')).toBe(true)
+    })
+
+    it('when passed a non-postcode string returns false', () => {
+      expect(validPostcodeArea('foo')).toBe(false)
     })
   })
 })
--- a/server/utils/formUtils.ts
+++ b/server/utils/formUtils.ts
@@ -86,3 +86,12 @@ export function convertKeyValuePairToCheckBoxItems<T>(
     }
   })
 }
+
+/**
+ * Performs validation on the area of a postcode (IE the first three characters)
+ * @param string string to be validated.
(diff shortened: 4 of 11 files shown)


server/utils/formUtils.ts
```diff
@@ -86,3 +86,7 @@ export function convertKeyValuePairToCheckBoxItems<T>(
     }
   })
 }
+
+export function validPostcode(potentialPostcode: string) {
```
pezholio, line 90: This should just be for the postcode area (i.e. SW1, WS13 etc etc)

server/form-pages/apply/move-on/pduRegion.ts
```diff
@@ -0,0 +1,46 @@
...
+    return {
+      [this.question]: this.body.postcode,
+    }
+  }
+
+  errors() {
+    const errors: TaskListErrors<this> = {}
+
+    if (!this.body.postcode) {
+      errors.postcode = 'You must enter a postcode for the PDU location'
```
pezholio, line 39: I was thinking we would just ask for the first part of the postcode here, as exact postcodes might be too granular. Also, I was under the impression that we'd be asking for the the postcode area of where they're relocating to, not necessarily the PDU region

server/form-pages/apply/move-on/pduRegion.ts
```diff
@@ -0,0 +1,46 @@
...
+  body: {
+    postcode: string
+  }
+
+  constructor(body: Record<string, unknown>) {
+    this.body = { postcode: body.postcode as string }
+  }
+
+  previous() {
+    return ''
```
pezholio, line 22: This previous value will probs be `placement-duration` now

server/utils/formUtils.ts
richpjames: Ok I'll update the function

server/form-pages/apply/move-on/pduRegion.ts
richpjames: Ok yeah that makes sense, I think I misread the question.
I've asked Selina for some content input on this because I think we could do with it and we can come back to it to improve later on but I will clarify the error for now

server/utils/formUtils.test.ts
```diff
@@ -310,4 +311,13 @@ describe('formUtils', () => {
       ])
     })
   })
+  describe('validPostcode', () => {
+    it('when passed a postcode it returns true', () => {
```
pezholio, line 315: Can we update this to be just a postcode area?

server/form-pages/apply/move-on/pduRegion.ts
```diff
@@ -0,0 +1,46 @@
+import type { TaskListErrors } from '@approved-premises/ui'
+import { validPostcodeArea } from '../../../utils/formUtils'
+
+import TasklistPage from '../../tasklistPage'
+
+export default class PduRegion implements TasklistPage {
```
pezholio, line 6: Should this be `RelocationArea` or similar now?

server/form-pages/apply/move-on/pduRegion.ts
```diff
@@ -0,0 +1,46 @@
...
+  name = 'pdu-region'
+
+  title = 'Placement duration and move on'
+
+  question = 'Probation delivery unit (PDU) postcode region'
+
+  body: {
+    postcode: string
```
pezholio, line 14: Should probably rename this to `postcodeArea` I think

server/form-pages/apply/move-on/pduRegion.ts
```diff
@@ -0,0 +1,46 @@
...
+import { validPostcodeArea } from '../../../utils/formUtils'
+
+import TasklistPage from '../../tasklistPage'
+
+export default class PduRegion implements TasklistPage {
+  name = 'pdu-region'
+
+  title = 'Placement duration and move on'
+
+  question = 'Probation delivery unit (PDU) postcode region'
```
pezholio, line 11: Are we sure this is the postcode region of the PDU? I was under the impression this would be the region the person would be relocating to?

pezholio (commented):
Just a couple of naming things. I'm still unsure what the postcode region refers to though?

server/form-pages/apply/move-on/placementDuration.test.ts
```diff
@@ -109,7 +109,7 @@ describe('PlacementDuration', () => {
     }
     application = applicationFactory.build({ data })
 
-    itShouldHaveNextValue(new PlacementDuration({}, application), '')
+    itShouldHaveNextValue(new PlacementDuration({}, application), 'relocation-region')
```
pezholio, line 112: I think this has snuck in from another commit

cypress_shared/pages/apply/relocationRegion.ts
```diff
@@ -0,0 +1,11 @@
+import Page from '../page'
+
+export default class RelocationRegionPage extends Page {
+  constructor() {
+    super('Placement duration and move on')
```
pezholio, line 5: Should we check for the question title with the person's name here?

pezholio (approved):
LGTM - just a couple of minor things. Thanks for bearing with me!

PR #325, opened 2022-11-08
(640) - Apply - Access needs screen

richpjames:
This PR adds an Apply questionaire page for the 'Access needs' screen in the pattern established in #142.

The eagle eyed reviewer may notice that this question asks questions unrelated to the PoP's access needs. I have raised this with our content and design team, they are working on it and the title can be updated once I've heard back.

[Trello ticket](https://trello.com/c/l14emuWP/640-access-needs-screen)

## Screenshots of UI changes
Apply -- allows completion of the form

PR #348, opened 2022-11-10
Disable other access fields when none selected

pezholio:
This follows on from #325 to disable and clear the access needs checkboxes when `none` is selected. We handle this in the backend, but I thought it would be good to add this for usablity (and I has my frontend JS hat on 😆 )

### DIFF
--- a/cypress_shared/pages/page.ts
+++ b/cypress_shared/pages/page.ts
@@ -62,6 +62,10 @@ export default abstract class Page {
     cy.get(`input[name="${name}"][value="${option}"]`).check()
   }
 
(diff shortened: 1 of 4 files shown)


cypress_shared/pages/page.ts
```diff
@@ -62,6 +62,10 @@ export default abstract class Page {
     cy.get(`input[name="${name}"][value="${option}"]`).check()
   }
 
+  unCheckCheckboxByNameAndValue(name: string, option: string): void {
```
richpjames, line 65: I think it should be `uncheckCheckboxbyNameAndValue`

richpjames (approved):

PR #349, opened 2022-11-10
(651) Apply: Move on - plans in place screen

richpjames:
This adds an Apply screen to the 'Move on' section in the pattern established in #142 
This one is a simple yes/no answer. 


pezholio (approved):
LGTM 👍

PR #352, opened 2022-11-11
(653) Apply: Move on - Foreign national screen

richpjames:
# Changes in this PR
This PR adds a screen asking whether the Home Office have been informed that accomodation will be required. 
This page is only shown if the user responds with 'foreignNational' to the accommodationType question. 
If the user answers 'Yes' they are asked to supply a date of notification.

[Trello card](https://trello.com/c/MamQ7L5x/653-foreign-national-screen)

## Screenshots of UI changes


server/views/applications/pages/move-on/foreign-national.njk
```diff
@@ -0,0 +1,46 @@
+{% extends "../layout.njk" %}
+
+{% block questions %}
+
+  <h1 class="govuk-heading-l">{{page.title}}</h1>
+
+  {% set fnInformed %}
+  {{ govukDateInput({
```
pezholio, line 8: We can / should use the `applyDateInput` macro here

richpjames: I didn't realise we had one, you thought of everything! Will add it in

pezholio (approved):
LGTM 👍

PR #359, opened 2022-11-16
(749) Fetch the application from the API on the tasklist

pezholio:
When loading the tasklist, and the application is not found in the session, then we should load it from the API. This will allow people to come back to an application after a period of time.

### DIFF
--- a/server/controllers/apply/applicationsController.ts
+++ b/server/controllers/apply/applicationsController.ts
@@ -1,5 +1,5 @@
-import type { Request, Response, RequestHandler, NextFunction } from 'express'
-import createError from 'http-errors'
+import type { Request, Response, RequestHandler } from 'express'
+import type { Application } from '@approved-premises/api'
 
 import ApplicationService from '../../services/applicationService'
 import { PersonService } from '../../services'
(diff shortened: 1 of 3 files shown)


pezholio:
[Fetch application from API when loading tasklist](https://trello.com/c/ITyjhWG3/749-fetch-application-from-api-when-loading-tasklist)

server/controllers/apply/applicationsController.ts
```diff
@@ -29,15 +29,17 @@ export default class ApplicationsController {
 
   show(): RequestHandler {
     return async (req: Request, res: Response, next: NextFunction) => {
-      const { application } = req.session
+      let application: Application
```
richpjames, line 32: I think you could do
`const application: Application =  req.session.application ?  req.session.application : await this.applicationService.findApplication(req.user.token, req.params.id)`

richpjames (approved):
Looks good, 
I do wonder how much benefit we get from having the Application in session but it's not a big deal

server/controllers/apply/applicationsController.ts
pezholio: Much better - love a one-liner

PR #361, opened 2022-11-17
Fix placement purpose

pezholio:
There were a couple of issues with the Placement Purpose screen with translations not returning properly, as well as errors showing when `other` was the only option provided.

### DIFF
--- a/server/@types/ui/index.d.ts
+++ b/server/@types/ui/index.d.ts
@@ -74,11 +74,15 @@ export interface RadioItem {
   checked?: boolean
 }
 
-export interface CheckBoxItem {
-  text: string
-  value: string
-  checked?: boolean
-}
+export type CheckBoxItem =
+  | {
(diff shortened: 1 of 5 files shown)


server/@types/ui/index.d.ts
```diff
@@ -74,11 +74,15 @@ export interface RadioItem {
   checked?: boolean
 }
 
-export interface CheckBoxItem {
-  text: string
-  value: string
-  checked?: boolean
-}
+export type CheckBoxItem =
+  | {
```
bryangaledxw, line 78: Does this initial `|` do anything?

bryangaledxw (approved):
LGTM

server/@types/ui/index.d.ts
pezholio: Eslint puts in in automatically for `$REASONS`

PR #371, opened 2022-11-22
(789) - Apply - Submit application

richpjames:
# Context
We need to submit completed applications to the API.
The word submission is perhaps misleading, maybe 'finalize' would be more accurate as we are only 'submitting' the Application ID as the API already holds all the data it needs. The user simply tells it the application is complete.

As part of this work I have added a confirmation screen as it was a small piece of work and we needed somewhere to redirect to.

[Submission Trello card](https://trello.com/c/Gjsg2teU/789-persist-submitted-documents-in-translated-json)
[Confirmation screen Trello card](https://trello.com/c/I8ngVZL4/657-confirmation-screen)

# Changes in this PR
We add the required methods to the client, service and controller and wire up the 'Submit application' button to confirm completion of the application to the API

## Screenshots of UI changes

### DIFF
--- a/server/controllers/apply/applicationsController.ts
+++ b/server/controllers/apply/applicationsController.ts
@@ -48,7 +48,7 @@ export default class ApplicationsController {
       if (crnArr.length) {
         const person = await this.personService.findByCrn(req.user.token, crnArr[0])
 
-        return res.render(`applications/confirm`, {
+        return res.render(`applications/people/confirm`, {
           pageHeading: `Confirm ${person.name}'s details`,
           ...person,
           date: DateFormats.dateObjtoUIDate(new Date()),
--- a/server/controllers/apply/applicationsController.ts
+++ b/server/controllers/apply/applicationsController.ts
@@ -7,6 +7,7 @@ import { fetchErrorsAndUserInput } from '../../utils/validation'
 import paths from '../../paths/apply'
 import { DateFormats } from '../../utils/dateUtils'
 import { sections } from '../../form-pages/apply'
+import { getResponses } from '../../utils/applicationUtils'
 
 export default class ApplicationsController {
   constructor(private readonly applicationService: ApplicationService, private readonly personService: PersonService) {}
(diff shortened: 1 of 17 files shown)


pezholio (commented):
Cool - nice work! I think it might be a good opportunity to confirm the data we expect is sent to the application - a similar pattern to https://github.com/ministryofjustice/hmpps-approved-premises-ui/blob/main/integration_tests/tests/manage/arrivals.cy.ts#L42

server/controllers/apply/applicationsController.ts
```diff
@@ -48,7 +48,7 @@ export default class ApplicationsController {
       if (crnArr.length) {
         const person = await this.personService.findByCrn(req.user.token, crnArr[0])
 
-        return res.render(`applications/confirm`, {
+        return res.render(`applications/personConfirmation`, {
```
pezholio, line 52: I think `applications/people/confirm` might be better naming here

server/controllers/apply/applicationsController.ts
```diff
@@ -78,4 +79,16 @@ export default class ApplicationsController {
       )
     }
   }
+
+  submit(): RequestHandler {
+    return async (req: Request, res: Response) => {
+      const application = await this.applicationService.findApplication(req.user.token, req.params.id)
+      application.document = getResponses(application)
+
+      await this.applicationService.saveToApi(application, req)
```
pezholio, line 88: Do we need to save to the API and submit here?

server/views/applications/submissionConfirmation.njk
```diff
@@ -0,0 +1,43 @@
...
+    <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds">
+            {{ govukPanel({
+                titleText: "Application confirmation"
+            }) }}
+            <p>
+          We’ve sent you a confirmation email.
+        </p>
+            <p>
+                <a href="">Download a copy of this application (PDF, 708 KB) </a>
```
pezholio, line 21: I missed this from the prototypes. Will be very cool to be able to do this. Should be fairly easy too!

server/views/applications/submissionConfirmation.njk
```diff
@@ -0,0 +1,43 @@
...
+            </p>
+
+            <h2 class="govuk-heading-m">What happens next</h2>
+
+            <p>We’ve sent your application to the Midlands central referral unit (CRU) for Approved Premises (AP).</p>
```
pezholio, line 26: Not for now, but I think we'll need to get the CRU from the API response

pezholio (approved):
LGTM - just some minor bits

server/controllers/apply/applicationsController.ts
richpjames: We probably don't have to... 
But we only submit the translated document.
By saving it we save the application with the translated text so the API then has the final version of the full Application object. 
But the API must be doing some work to attatch the translated text to the application object anyway

PR #376, opened 2022-11-23
Return user to tasklist after answering oral hearing question

richpjames:
Previously we took the user to the Placement Date page. 
The Placement Date page requires there to be a release date in the session however on the Release Date page they answered that they don’t know the the release date. There doesn’t make sense to move them on to the placement date page so instead we return them to the tasklist.

This has been confirmed by our SME [here](https://mojdt.slack.com/archives/C02SHKETZ7F/p1668418539408919?thread_ts=1668174667.005489&cid=C02SHKETZ7F)
[Sentry Issue here](https://sentry.io/organizations/ministryofjustice/issues/3760879622)

### DIFF
--- a/server/form-pages/apply/basic-information/oralHearing.test.ts
+++ b/server/form-pages/apply/basic-information/oralHearing.test.ts
@@ -38,7 +38,7 @@ describe('OralHearing', () => {
     })
   })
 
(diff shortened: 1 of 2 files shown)


pezholio (approved):
LGTM 👍

PR #382, opened 2022-11-28
Add an ADR for our use of decorators

pezholio:
This codifies the work around decorators that we have already done in #386 and #391


richpjames:
Makes sense to me. 
Given the use of decorators in Angular/Nest I don't have any reservations about this implementation. 
It looks like [they may be added to ECMAscript](https://github.com/tc39/proposal-decorators/) (eventually).

richpjames (approved):

PR #391, opened 2022-11-30
(890) Refactor form pages into sections

pezholio:
This adds decorators for Tasks and Sections, which allows us to get the name and Tasks that a section contains, as well as all the pages within a Task. This also allows us to arrange Tasks into Section folders, and tidy up how we define what the Apply Form contains - this will help us going into Apply too, and also allow us to easily define a schema from a form object.


server/form-pages/apply/access-and-healthcare/index.ts
```diff
@@ -1,11 +1,20 @@
+// eslint-disable-next-line max-classes-per-file
```
bryangaledxw, line 1: I don't think this is needed?

server/form-pages/apply/access-and-healthcare/index.ts
```diff
@@ -1,11 +1,20 @@
...
 import AccessNeeds from './accessNeeds'
 import AccessNeedsMobility from './accessNeedsMobility'
 import Covid from './covid'
 import AccessNeedsAdditionalAdjustments from './accessNeedsAdditionalAdjustments'
+import { Task } from '../../utils/decorators'
 
 export default {
```
bryangaledxw, line 8: It tripped me up that this doesn't follow the conventions of other files in this commit. Worth changing to `const pages = ....` and `export default pages`?

server/form-pages/utils/index.ts
```diff
@@ -32,3 +32,18 @@ export const getTask = <T>(task: T) => {
     pages: taskPages,
   }
 }
+
+export const getSection = <T>(section: T) => {
```
bryangaledxw, line 36: My preference would have been for this to be in its own commit, as it gets a bit lost among the other changes

bryangaledxw (approved):
Looks good. Just wondering if the file moves and import fix-up from "Organise tasks into sections" can be split from the other changes in that commit, since they get a bit lost

server/form-pages/utils/index.test.ts
```diff
@@ -1,87 +1,65 @@
 /* eslint-disable max-classes-per-file */
 import 'reflect-metadata'
-import {
-  applyYesOrNo,
-  yesOrNoResponseWithDetail,
-  yesNoOrDontKnowResponseWithDetail,
-  getTask,
-  getSection,
-} from './index'
+import * as utils from './index'
```
richpjames, line 3: I think we should stick to named imports as it's what we use elsewhere

richpjames (approved):
LGTM left a couple comments

server/form-pages/utils/index.test.ts
pezholio: I needed to do a wildcard import here to allow me to use a spy later on. Will tweak the commit message to make it clear (might even add a comment)

server/form-pages/utils/index.test.ts
```diff
@@ -1,4 +1,6 @@
-import { applyYesOrNo, yesOrNoResponseWithDetail, yesNoOrDontKnowResponseWithDetail } from './index'
+/* eslint-disable max-classes-per-file */
```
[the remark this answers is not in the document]
pezholio: Yeah, agreed

server/form-pages/apply/access-and-healthcare/index.ts
pezholio: Ah, we don't even need the page exports anymore! Will remove then

PR #159, opened 2022-09-12
(661) Previous applications dashboard

richpjames:
# Context

We need to be able to show Probation Practicioner's previous applications so they know the status of each one.
[Trello card](https://trello.com/c/AVyXDHjh/661-previous-applications-dashboard).
This neccessitates a [new endpoint in the API which I have specced out here](https://github.com/ministryofjustice/approved-premises-api/pull/65)

# Changes in this PR
I have followed the patterns we have used throughout development in this repo. The client fetches the data, the service maps this into a form that is usable by the controller and the controller renders the page. 

### After
image

PR #392, opened 2022-11-30
(799) Application dashboard iterations

richpjames:
# Context
The first iteration of the dashboard (#159) made some incorrect assumptions about where the data was coming from.
This PR updates the dashboard so we no longer have to use the ApplicationSummary object to get the data.

[Trello ticket](https://trello.com/c/HcX0l05U/799-application-dashboard-pt-2)

# Changes in this PR
- In the first 4 commits we rationalize all the names of the functions in DateFormats so that they all use the same language
- The risk tier comes from the `people/{crn}/risks` endpoint so we need to call that to retrieve the Application's person's risks
- We remove the status badge, this is on the mocks but doesn't exist in the API so we're removing it for now. We can add it back in if needs be
- We now derive the arrival date from the Application response as opposed to pulling it from the summary
- We add a 'submitted at' column
- Following this we remove all references to ApplicationSummary which no longer exists

### Before
image

### After
Applications dashboard -- shows the dashboard


e2e/tests/stepDefinitions/bookingExtensions.ts
```diff
@@ -9,7 +9,7 @@ import {
 
 import { DateFormats } from '../../../server/utils/dateUtils'
 
-const extensionDate = DateFormats.formatApiDate(new Date(2050, 0, 1))
```
pezholio, line 12: Thanks for taking the time to do this - the naming was confusing as hell!

server/controllers/apply/applicationsController.test.ts
```diff
@@ -39,7 +39,7 @@ describe('applicationsController', () => {
 
   describe('list', () => {
     it('renders the list view', async () => {
-      applicationService.tableRows.mockResolvedValue([])
+      applicationService.dashboardTableRows.mockResolvedValue([])
```
pezholio, line 42: Not for now, but I think it might be worth us moving a bunch of this kind of logic to helper functions that we can use in the views, rather than sending marshalled data via the controller. Just so we're seperating data and presentation layer concerns a bit better

pezholio: I'm quite aware that I might be breaking this rule elsewhere (and I probably introduced the patten in the first place 😜 )

server/form-pages/apply/basic-information/placementDate.test.ts
```diff
@@ -6,7 +6,7 @@ import applicationFactory from '../../../testutils/factories/application'
 
 describe('PlacementDate', () => {
   const releaseDate = new Date().toISOString()
-  const application = applicationFactory.build({ data: { 'basic-information': { 'release-date': { releaseDate } } } })
+  const application = applicationFactory.withReleaseDate(releaseDate).build()
```
pezholio, line 9: I ❤️ this

server/utils/applicationUtils.ts
```diff
@@ -86,4 +86,42 @@ const getPage = (taskName: string, pageName: string) => {
   return Page
 }
 
-export { getTaskStatus, taskLink, getCompleteSectionCount, getResponses, getResponseForPage, getPage }
+const getArrivalDate = (application: Application): string | null => {
```
pezholio, line 89: I do wonder if this is something we could/should do in the API at the point of submission, but this is OK for now

pezholio: I've added a card https://trello.com/c/U6OqS6Zb/928-get-application-arrival-date-at-point-of-submission

pezholio (approved):
LGTM - some of this logic can probably be pushed to the API, but we need to start prioritising and thinking MVP, so this gets us over the line 👍

server/utils/applicationUtils.ts
richpjames: Yeah 100% agree, my understanding was this was a workaround for the time being

PR #409, opened 2022-12-06
(945) Set index offence in frontend

pezholio:
This builds on a breaking change in https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/276 where we'll need to send the index offence through to the API when creating a new application. Rather than waiting for designs and ending up with the application in a broken state, we're just treating the first offence as the index offence. We can iterate on this as we go.


pezholio:
[Set index offence in frontend](https://trello.com/c/DstklAYs/945-set-index-offence-in-frontend)

richpjames (approved):

PR #415, opened 2022-12-07
(727) - Apply - OASys Import

richpjames:
# Context

Probation officers need to be able to select which optional sections ([supporting information](https://github.com/ministryofjustice/hmpps-approved-premises-ui/blob/6ae4cc119e543e02408953f56c4b4ed54ee9e875/server/@types/shared/models/OASysSections.ts#L20)) of OASys they would like to add to the application.
On this page (OptionalOasysSections) they can select these sections from two lists the first is needs linked to reoffending and the other is needs not linked to serious harm or reoffending (otherNeeds).
For the check your answers page and the persisted JSON document we only record the section number and title on this page but on the subsequent page (yet to be built) we will record the content of the imported sections.

In order to glean which sections of OASys are available for import we need to call the `/people/{crn}/oasys/selection` endpoint and filter into the two lists of needs.

[Trello card](https://trello.com/c/PYl2WtOG/727-oasys-risk-and-needs-import)

# Changes in this PR

## Screenshots of UI changes

image

### DIFF
--- a/server/data/personClient.ts
+++ b/server/data/personClient.ts
@@ -1,4 +1,12 @@
-import type { ActiveOffence, Adjudication, Person, PersonRisks, PrisonCaseNote } from '@approved-premises/api'
+import type {
+  ActiveOffence,
+  Adjudication,
+  Person,
+  PersonRisks,
+  PrisonCaseNote,
+  OASysSection,
+} from '@approved-premises/api'
+
 import RestClient from './restClient'
 import config, { ApiConfig } from '../config'
 import paths from '../paths/api'
@@ -42,5 +50,11 @@ export default class PersonClient {
     const response = await this.restClient.get({ path: paths.people.offences({ crn }) })
 
     return response as Array<ActiveOffence>
+  }
+
+  async oasysSelections(crn: string): Promise<Array<OASysSection>> {
+    const response = await this.restClient.get({ path: paths.people.oasys.selection({ crn }) })
+
+    return response as Array<OASysSection>
   }
 }
--- a/server/form-pages/apply/risk-and-need-factors/oasys-import/optionalOasysSections.ts
+++ b/server/form-pages/apply/risk-and-need-factors/oasys-import/optionalOasysSections.ts
@@ -1,0 +1,124 @@
+import { Page } from '../../../utils/decorators'
+
+import TasklistPage from '../../../tasklistPage'
+
+import { flattenCheckboxInput, isStringOrArrayOfStrings } from '../../../../utils/formUtils'
+import { Application, OASysSection } from '../../../../@types/shared'
+import { DataServices } from '../../../../@types/ui'
+import { sentenceCase } from '../../../../utils/utils'
+
+interface Response {
+  needsLinkedToReoffending: Array<string> | string | Array<OASysSection>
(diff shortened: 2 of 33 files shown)


server/paths/api.ts
```diff
@@ -63,5 +64,8 @@ export default {
     prisonCaseNotes: personPath.path('prison-case-notes'),
     adjudications: personPath.path('adjudications'),
     offences: personPath.path('offences'),
+    oasys: {
```
pezholio, line 67: I think this should be namespaced under `people`

server/form-pages/apply/risk-and-need-factors/oasys-import/optionalOasysSections.ts
```diff
@@ -0,0 +1,115 @@
...
+
+  otherNeedsItems() {
+    return this.checkboxes(this.allOtherNeeds, this.body.otherNeeds)
+  }
+
+  private checkboxes(fullList: OASysSection[], selectedList: string[]) {
+    return fullList.map(need => {
+      const sectionAndName = `${need.section}. ${sentenceCase(need.name)}`
+      return {
+        value: sectionAndName,
```
pezholio, line 105: Should the value be just the section number? It'll make it tricky to use in the next step if we have to extract the section number

server/data/personClient.ts
```diff
@@ -43,4 +51,10 @@ export default class PersonClient {
 
     return response as ActiveOffence[]
   }
+
+  async oasysSelections(crn: string): Promise<OASysSection[]> {
+    const response = await this.restClient.get({ path: paths.people.oasys.selection({ crn }) })
+
+    return response as OASysSection[]
```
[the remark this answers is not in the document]
richpjames: I don't really mind this tbh but if we are going to standardise we should do it with a linting rule.

server/paths/api.ts
richpjames: It already is? or do you mean remove the `oasys` object and have it as something like `oasysSelection`?

server/form-pages/apply/risk-and-need-factors/oasys-import/optionalOasysSections.ts
richpjames: I was having trouble with this as we then had to look up the section name when doing the CYA pages. It's solvable but I will need a bit of guidance

server/data/personClient.ts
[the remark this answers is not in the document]
pezholio: Totes - I actually thought we had one already, but I think I was confusing that with another project 😆

server/paths/api.ts
pezholio: D'oh! Sorry!

server/form-pages/apply/risk-and-need-factors/oasys-import/optionalOasysSections.ts
pezholio: Could we not do the lookup at the point of submission and then store the response as a key/pair value by overriding the body setter?

pezholio (approved):
LGTM - thanks for taking the time on this!

PR #417, opened 2022-12-08
(947) Allow the index offence to be selected at the start of the journey if an offender has more than one offence

pezholio:
This adds a new page to the start of the Apply journey if an offender has more than one offence associated with them. The guidance notes could use some work, but this is probably enough as a started for ten.

## Screenshot

image

### DIFF
--- a/server/controllers/apply/applicationsController.ts
+++ b/server/controllers/apply/applicationsController.ts
@@ -76,16 +76,20 @@ export default class ApplicationsController {
 
   create(): RequestHandler {
     return async (req: Request, res: Response) => {
-      const { crn } = req.body
-      const offences = await this.personService.getOffences(req.user.token, crn)
-      const indexOffence = offences.find(o => o.offenceId === req.body.offenceId)
+      const { crn, offenceId } = req.body
+      if (!offenceId) {
+        res.redirect(paths.applications.people.selectOffence({ crn }))
+      } else {
+        const offences = await this.personService.getOffences(req.user.token, crn)
+        const indexOffence = offences.find(o => o.offenceId === offenceId)
 
-      const application = await this.applicationService.createApplication(req.user.token, crn, indexOffence)
-      req.session.application = application
+        const application = await this.applicationService.createApplication(req.user.token, crn, indexOffence)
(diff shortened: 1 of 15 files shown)


pezholio:
[Allow the index offence to be selected at the start of the journey if an offender has more than one index offence](https://trello.com/c/UQHa6V4B/947-allow-the-index-offence-to-be-selected-at-the-start-of-the-journey-if-an-offender-has-more-than-one-index-offence)

server/controllers/apply/applicationsController.ts
```diff
@@ -51,17 +51,14 @@ export default class ApplicationsController {
         const person = await this.personService.findByCrn(req.user.token, crn)
         const offences = await this.personService.getOffences(req.user.token, crn)
 
-        // TODO: For now, we treat the first offence as the index offence, going
-        // forward, we'll need to design an approach to select one if there are
-        // more than one
-        const offence = offences[0]
+        const offenceId = offences.length === 1 ? offences[0].offenceId : undefined
```
richpjames, line 54: I don't feel too strongly about it but if we're setting a variable to be falsy I prefer `null` over `undefined` as it makes things easier to debug if something goes wrong.

server/views/applications/people/selectOffence.njk
```diff
@@ -0,0 +1,58 @@
...
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-full">
+      <h1>{{ pageHeading}}</h1>
+
+      <p class="govuk-body">
+        {{ person.name }} has more than one offence identified against their CRN.
+
+      <p class="govuk-body">
+        Please select the main offence that will be used to assess their sutability
+        for an Approved Premises (the "index offence") below:
```
richpjames, line 20: I think this should be `'` as opposed to `"` but, as ever, happy to be wrong!

richpjames (approved):
LGTM

server/controllers/apply/applicationsController.ts
pezholio: Yeah, good call 👍

PR #418, opened 2022-12-08
Remove the null allowlist for preprod

pezholio:
Even though we’re not mastering data in preprod yet, we are accessing preprod services, so should be using the master allowlist in the main `hmpps-approved-premises-ui/values.yaml` file


tahb (approved):
We can log in on both AP and TA so let's see what happens after removing this!

PR #420, opened 2022-12-08
Fix: Allow application dashboard to handle cases when tier value is null

richpjames:
This fixes the following [error in Sentry](https://sentry.io/organizations/ministryofjustice/issues/3783016406/events/d511cc5b34e647e5ad3aa8dbe016233b/?project=4503931667742720&referrer=alert-rule-issue-list)
Sometimes a tier's value is null and previously this would break application dashboard.


pezholio (approved):
LGTM 👍

PR #421, opened 2022-12-08
(870) Assessment tasklist

pezholio:
This adds a basic asssessment tasklist without any sections or tasks to set us on the right path towards building assess screens

## Screenshot

image


server/views/assessments/show.njk
```diff
@@ -0,0 +1,83 @@
...
+
+        {{ govukButton({
+          text: "Submit application"
+          }) }}
+      </form>
+
+      {{ govukButton({
+        text: "Save and come back later",
+        classes: "govuk-button--secondary",
+        href: "/pp-dashboard.html"
```
richpjames, line 76: I think this link snuck in from the prototype

server/controllers/assess/assessmentsController.ts
```diff
@@ -12,4 +13,12 @@ export default class AssessmentsController {
       res.render('assessments/index', { pageHeading: 'Approved Premises applications', assessments })
     }
   }
+
+  show(): RequestHandler {
+    return async (req: Request, res: Response) => {
+      const assessment = await this.assessmentService.findAssessment(req.user.token, req.params.id)
+
+      res.render('assessments/show', { assessment, sections: Assess.sections })
```
richpjames, line 21: can we pass the page title in here like we do for most other pages?

richpjames (approved):

server/views/assessments/show.njk
pezholio: Ah yeah, I think it's in the Apply one too. I'll fix it in a seperate PR 👍

PR #423, opened 2022-12-12
API model updates

brownbearsoftware:
Updating hmpps-approved-premises-api models from OpenAPI specification.  This PR was created automatically from the generate-types.yml Workflow


tahb:
Wondering how we can make these types of changes less breaking. Can we make them backwards compatible on the API side?

pezholio (approved):

PR #425, opened 2022-12-12
(654) Document sync screens

pezholio:
This adds the document sync screens for the Apply journey. Still need to do the actual download of the documents, but I'll do this in a seperate PR

## Screenshot

image

### DIFF
--- a/server/form-pages/apply/schema.json
+++ b/server/form-pages/apply/schema.json
@@ -12,6 +12,7 @@
     "access-and-healthcare",
     "further-considerations",
     "move-on",
+    "attach-required-documents",
     "check-your-answers"
   ],
   "properties": {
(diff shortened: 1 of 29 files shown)


pezholio:
[Document sync screens](https://trello.com/c/DchCcfE1/654-document-sync-screens)

server/form-pages/apply/schema.json
```diff
@@ -2,6 +2,7 @@
   "$schema": "https://json-schema.org/draft/2019-09/schema",
   "type": "object",
   "title": "Apply Schema",
+  "additionalProperties": false,
```
richpjames, line 5: I think it would be best if you saved this change without formatting. It'll be easier to see this is the only relevant change (I think?) and when we come to generate the next schema it'll be easier to see what the changes area

richpjames (approved):
LGTM, left a couple of comments

server/form-pages/apply/schema.json
pezholio: Ah, good catch. That was me just editing it in VS Code and the Prettier plugin doing its magic!

PR #429, opened 2022-12-14
(654) Download documents

pezholio:
This pipes the document downloads from the API to a stream on our frontend application. It was a bit fiddly to work out and test, but seems to work pretty well now. We also pass in the `content-disposition` header from the API, so the filename is the same as the one the API gets from Alfresco.


pezholio:
[Document sync screens](https://trello.com/c/DchCcfE1/654-document-sync-screens)

cypress_shared/pages/apply/attachDocumentsPage.ts
```diff
@@ -29,6 +30,28 @@ export default class AttachDocumentsPage extends Page {
...
+              doc.location.reload()
+            }, 300)
+          })
+          cy.get(`a[data-cy-documentId="${document.id}"]`).click()
+        })
+
+      const downloadsFolder = Cypress.config('downloadsFolder')
+      const downloadedFilename = path.join(downloadsFolder, document.fileName)
+
+      cy.readFile(downloadedFilename, 'binary', { timeout: 15000 })
```
richpjames, line 51: You may have already done this but could we start with a smaller value and increase it if it takes longer?

richpjames (approved):
LGTM

cypress_shared/pages/apply/attachDocumentsPage.ts
pezholio: Good call 👍

PR #432, opened 2022-12-14
(963) Use first class risk ratings in frontend for applications

pezholio:
Now Risks are a first class citizen of Applications, we can update the app to use the new application-specific Application, as well as use the first-class ratings object that we now get from the API. This makes listing applications much easier to do, so we're not having to make multiple API calls each time we list the applications.

I've also removed the `dashboardTableRows` fuction from the service and made this a view helper, as well as fixing the issues with the missing dates in the application dashboard.


pezholio:

richpjames (approved):
LGTM

PR #435, opened 2022-12-15
Fetch application from the API on the tasklist

pezholio:
If we’re dealing with multiple Applications, then it will automatically try and fetch the last Application we entered (which may be incomplete).

I’m not entirely sure how useful the session stuff is now, as we’re sending to the API on each screen, but we can clean this up later.


richpjames (approved):
Yeah agreed the session stuff isn't useful

richpjames (approved):

PR #433, opened 2022-12-14
(727) - Apply - OASys: Supporting information

richpjames:
# Context
This is very similar to #431 #426.
This time however we have to call the OASys sections endpoint with the options selected in the OptionalOasysImport section of the Apply journey.

[Trello card](https://trello.com/c/PYl2WtOG/727-oasys-risk-and-needs-import)

## Screenshots of UI changes
Apply -- allows completion of the form

PR #431, opened 2022-12-14
727 -  Apply -  OASys import:  Offence details

richpjames:
# Context
Similar to #426 users need to be able to add and edit sections from the Offence Details section of OASys.

[Trello card](https://trello.com/c/PYl2WtOG/727-oasys-risk-and-needs-import)

# Changes in this PR
Firstly there is some tidying of the RoshSummary class. 
Then we extract some methods that are to be used by both the RoshSummary and OffenceDetails class.
Next we add the class and view themselves, they are very similar to the RoshSummary.
After that we're tidying up: making sure RoshSummary correctly links to the OffenceDetails page and updating the schema.

## Screenshots of UI changes
Apply -- allows completion of the form

PR #426, opened 2022-12-12
(727) - Apply - OASys RoSH summary

richpjames:
# Context
Users need to able to import and edit the RoSH (Risk of Serious Harm) summary questions from OASys. This page enables them to do that.

[Trello card](https://trello.com/c/PYl2WtOG/727-oasys-risk-and-needs-import)

# Changes in this PR
Firstly there is some tidying from #415 where we had misnamed some OASys '*selections*' '*sections*'.
Then we add the neccessary methods to the PersonClient and the PersonService.
We then add the Class for RoSH Summary which pulls the sections from the API. 
We also need to show the risks on this page so we call the API for those as well.
For the view we iterate over each question/answer from OAsys and display the answer in a text box (if it exists) for the user to edit.

## Screenshots of UI changes
Apply -- allows completion of the form

PR #436, opened 2022-12-15
(727) - Apply - OASys import: Risk Management Plan and Risk To Self Pages

richpjames:
# Context
Similar to #433 #431 #426 we need to import the Risk Management Plan and Risk To Self pages.

[Trello card](https://trello.com/c/PYl2WtOG/727-oasys-risk-and-needs-import)

# Changes in this PR

These are the final two screens, I will put in a subsequent PR to tidy things up a bit

## Screenshots of UI changes

Apply -- allows completion of the form (4)

### DIFF
--- a/integration_tests/tests/apply/apply.cy.ts
+++ b/integration_tests/tests/apply/apply.cy.ts
@@ -57,12 +57,14 @@ import {
   documentsFromApplication,
   offenceDetailSummariesFromApplication,
   riskManagementPlanFromApplication,
+  riskToSelfSummariesFromApplication,
   roshSummariesFromApplication,
   supportInformationFromApplication,
 } from '../../support/helpers'
 import OffenceDetailsPage from '../../../cypress_shared/pages/apply/offenceDetails'
 import SupportingInformationPage from '../../../cypress_shared/pages/apply/supportingInformation'
 import RiskManagementPlanPage from '../../../cypress_shared/pages/apply/riskManagementPlan'
+import RiskToSelfPage from '../../../cypress_shared/pages/apply/riskToSelf'
 
 context('Apply', () => {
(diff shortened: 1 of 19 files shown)


integration_tests/tests/apply/apply.cy.ts
```diff
@@ -696,8 +705,8 @@ context('Apply', () => {
 
       // Then the application should be submitted to the API
       cy.task('verifyApplicationUpdate', application.id).then(requests => {
-        expect(requests).to.have.length(36)
-        const requestBody = JSON.parse(requests[35].body)
+        expect(requests).to.have.length(37)
```
pezholio, line 708: Idly wondering if we can do something about these magic numbers here - is there maybe a way we can count the number of pages or something?

integration_tests/tests/apply/apply.cy.ts
```diff
@@ -696,8 +705,8 @@ context('Apply', () => {
...
+        const requestBody = JSON.parse(requests[36].body)
```
pezholio, line 709: Can probably do this here:

```
const requestBody = JSON.parse(requests[requests.length - 1].body)
```

pezholio: Also maybe `applicationSubmissionBody` might be a better name, as `requestBody` is a bit generic.

pezholio (approved):
LGTM - just some tiny things, not necessarily for now. Can probably be included in the next PR if you think they're worth addressing

integration_tests/tests/apply/apply.cy.ts
richpjames: Yep, will fix in a subsequent PR

PR #439, opened 2022-12-15
(963) Use first class risk ratings in frontend for assessments

pezholio:
Now the risk information is automatically included in the Application data, we no longer need to hit the risks endpoint each time.


pezholio:

PR #440, opened 2022-12-15
Use first class risk ratings on tasklist

pezholio:
I’ve also added the `mapApiPersonRisksForUi` as a nunjucks filter, so we can use it in the view. This is the last place we're fetching risks manually, other than the OAsys stuff, which is in-flight at the mo.


pezholio:

richpjames (approved):

PR #437, opened 2022-12-15
Add CodeQL analysis action

pezholio:
This will automatically analyse our code for common security issues

PR #441, opened 2022-12-15
Make form pages safer

pezholio:
When introducing CodeQL in #437, we've had a couple of alerts  about using using user-provided values in the Apply journey:

- https://github.com/ministryofjustice/hmpps-approved-premises-ui/security/code-scanning/1
- https://github.com/ministryofjustice/hmpps-approved-premises-ui/security/code-scanning/11

Through the power of decorators and metadata, we can glean the name of the page and its associated task, thereby solving those two issues.


bryangaledxw (approved):
LGTM

PR #449, opened 2022-12-19
Add rate limiting to the authentication endpoints

pezholio:
Following [a recommendation by CodeQL](https://github.com/ministryofjustice/hmpps-approved-premises-ui/security/code-scanning/3), I’ve added some rate limiting to the endpoints that use authentication. This uses `express-rate-limit` and `rate-limit-redis` to prevent too many requests to specific endpoints/middleware. The rateLimiter function has options we can pass to set the window and max number of requests that can be made to an endpoint before hitting a rate limit.

For the authentication endpoints, I’ve set this to 10 requests per IP address per 10 minute window, and 100 requests per 1 minute windw for the authorisation endpoints which seems like a reasonable default to start with.


pezholio:
I'm going to close this as a no-op. It's going to be tricky to do rate-limiting on an IP basis, as the IP addresses that will be accessing the application will be limited to a handful of IPs from the corporate network or VPN, so we'll end up rate limiting unneccessarily. It's also low-risk as the application is purely internal.

PR #450, opened 2022-12-19
956, 957, 958, 959, 960 Bug tickets

richpjames:
# Context

This PR fixes some of the issues spotted during mock user testing. 

Tickets: 
[956](https://trello.com/c/LKh43dZf/956-accommodation-is-mispelled-in-some-places-throughout-the-app), [957](https://trello.com/c/OVGsbrTO/957-back-buttons-dont-exist-on-the-first-page-of-each-section), [958](https://trello.com/c/JtPv0SWk/958-copy-edits), [960](https://trello.com/c/KGOJihAa/960-remove-access-needs-additional-adjustments-page), [959](https://trello.com/c/5X5iCVCX/959-back-to-dashboard-link-takes-users-to-booking-page)

# Changes in this PR
- Correct mispelling 'accomodation' of 'accommodation'
- Minor copy changes
- Add back button for the first page of each task and test all the other back buttons
- Remove 'access needs additional adjustments page'
- Back to dashboard links in Manage now take users back to Premises dashboard as opposed to the booking page


server/views/applications/pages/layout.njk
```diff
@@ -21,7 +21,12 @@
 
 {% block content %}
 
-  {% if page.previous() %}
+  {% if page.previous() === 'dashboard' %}
```
pezholio, line 24: Should we default to going back to the dashboard if there's no back link rather than having `dashboard`? I can't think of any circumstance where we won't have a back link.

pezholio (approved):
LGTM - just one teeny thing

server/views/applications/pages/layout.njk
richpjames: So the exception to the rule is the first page after the CRN lookup (Sentence Type) which you can't go back on. 
Although the code isn't as nice I thought it might be clearer to people other than us where the link is going to have it explicitly say 'dashboard' (although we don't do this for the `next` method. 
I could do something different though EG have an explicit 'none' returned from the `previous` method mean we don't show a backlink, if you think that would be better?

PR #451, opened 2022-12-20
20/12 Content changes

beth-dixxon:
# Context
We've made some content changes based off the AP backlog. Tests were failing on the previous commit due to 're-release' being removed as an option on the 'Sentence type' screen.

# Changes in this PR

-  content changes to risk management plan 
-  removed 're-release' from sentence type and changed item length (which was causing issues) 
-  changed 'service user' to person on PP dashboard


richpjames (approved):
Looks good @beth-dixxon 
I've made a couple of changes to the commit history just to tidy things up a bit.

PR #452, opened 2022-12-20
Fix/E2E test fail due to 'Back to dashboard' link changing

richpjames:
The adjustment of the 'Back To Dashboard' link in #450 broke the booking extension E2E tests but it didn't show up until they ran post-merge.


pezholio (approved):
LGTM 👍

PR #453, opened 2022-12-21
(982) Fix dates in OASys apply screens

pezholio:
If `dateCompleted` is blank from an OASys response, we were getting an exception. After a discussion with the team, we decided that if the assessment was not completed, we'd use the `dateStarted` to get the date that OASys was last updated. This also uses a partial for those screens to DRY things up a bit.


pezholio:
[Fix dates in Oasys Apply screens](https://trello.com/c/dTmqzJ1y/982-fix-dates-in-oasys-apply-screens)

richpjames (approved):
LGTM

PR #455, opened 2022-12-21
(983) Fix OptionalOasysSections page

pezholio:
When we revisit the Oasys Sections page (for example, to change a response), the body contains the actual section objects, rather than the section IDs, before, we would get an error as the mapping was attempting to match the selected sections against the section ID oft he returned sections. This refactors the code to only do the mapping if the section IDs are present, otherwise pass the objects through as they are.

I've also fixed a bug in the template and moved the checkbox logic to a util method, to fit in with our latest way of doing things.


pezholio:
[Fix OptionalOasysSections page](https://trello.com/c/FvsXQHAt/983-fix-optionaloasyssections-page)

PR #465, opened 2023-01-05
(891) End to end tests for Apply

pezholio:
This adds e2e tests for the Apply journey, as part of this work, I've added a shared helper that runs through the whole Apply journey, for tidiness and ease of maintenance.


pezholio:
[End to End tests for Apply](https://trello.com/c/qWSDIqJV/891-end-to-end-tests-for-apply)

cypress_shared/pages/apply/attachDocumentsPage.ts
```diff
@@ -1,5 +1,4 @@
 import { ApprovedPremisesApplication, Document } from '@approved-premises/api'
```
bryangaledxw, line 1: I ran into the same issue, but solved it with some changes in the cypress config - https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/74/commits/5632be8f91d7feba056e5a0a9c8418ccc8b4894c

cypress.config.e2e.ts
```diff
@@ -23,7 +23,7 @@ async function setupNodeEvents(
 
 export default defineConfig({
   chromeWebSecurity: false,
-  fixturesFolder: 'e2e/fixtures',
+  fixturesFolder: 'integration_tests/fixtures',
```
bryangaledxw, line 26: Should this also be `cypress_shared`?

cypress_shared/helpers/apply.ts
```diff
@@ -0,0 +1,690 @@
...
+    attachDocumentsPage.clickSubmit()
+
+    // Then I should be taken back to the task list
+    const tasklistPage = Page.verifyOnPage(TaskListPage)
+
+    // And the Attach Documents task should show a completed status
+    tasklistPage.shouldShowTaskStatus('attach-required-documents', 'Completed')
+  }
+
+  completeCheckYourAnswersSection() {
```
bryangaledxw, line 652: Can this and `submitApplication` also be private?

bryangaledxw (commented):
Looks good, just a few minor comments

Before I hit approve, can you help me understand why those two `this.type === 'integration'` checks are needed?

pezholio:
@bryangaledxw Hmmm - that's interesting. I'm gonna look into this a bit more, because we _might_ not need these checks....

pezholio:
@bryangaledxw I've looked at this, and basically, we _can_ run these tests in e2e, but it will require a bunch more (potentially brittle) setup to ensure it matches the data in the external systems, so we skip these. I'll add a comment to make it clearer why we're doing this, and only skip the ones that are necessary.

bryangaledxw (approved):

PR #467, opened 2023-01-06
(1000) Add person names to bookings

pezholio:
This adds a person's name when viewing Bookings on the premises page. Before we didn't have access to this information via the API, but now we do, I think it makes sense to show this information upfront. I've also taken the opportunity to add some quality of life improvements to better match the patterns we're using in services and views.

## Before

before

## After

after

### DIFF
--- a/cypress_shared/pages/manage/premisesShow.ts
+++ b/cypress_shared/pages/manage/premisesShow.ts
@@ -86,12 +86,13 @@ export default class PremisesShowPage extends Page {
   private tableShouldContainBookings(bookings: Array<Booking>, type: 'arrival' | 'departure') {
     bookings.forEach((item: Booking) => {
       const date = type === 'arrival' ? item.arrivalDate : item.departureDate
-      cy.contains(item.person.crn)
+      cy.contains(item.person.name)
         .parent()
         .within(() => {
-          cy.get('td').eq(0).contains(DateFormats.isoDateToUIDate(date))
+          cy.get('td').eq(0).contains(item.person.crn)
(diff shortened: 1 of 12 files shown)


pezholio:
[Add person names to bookings](https://trello.com/c/THnPcw1E/1000-add-person-names-to-bookings)

server/utils/bookingUtils.test.ts
```diff
@@ -92,8 +92,6 @@ describe('bookingUtils', () => {
     })
   })
 
-  describe('currentResidentsToTableRows', () => {})
-
```
bryangaledxw: Think this is the mirror of the stray added line in previous commit

server/form-pages/apply/schema.json
```diff
@@ -940,19 +940,35 @@
           "type": "object",
           "properties": {
             "additionalNeeds": {
-              "type": "array",
```
bryangaledxw, line 943: Is this a tidy-up that could be in its own commit?

cypress_shared/pages/manage/premisesShow.ts
```diff
@@ -7,6 +7,7 @@ import { DateFormats } from '../../../server/utils/dateUtils'
 export default class PremisesShowPage extends Page {
   constructor(private readonly premises: ApprovedPremises) {
     super(premises.name)
+    cy.screenshot()
```
bryangaledxw, line 10: Is this leftover debugging?

bryangaledxw (approved):
Looks good, just a few minor comments

bryangaledxw (approved):

PR #472, opened 2023-01-10
Content changes

beth-dixxon:
# Context

Content changes across front-end UI. 
We've added 
- details component, including additional guidance for PP's on the type of ap screen 
- changed content on Type of AP screen radio buttons
- updated guidance information on 'Location factors' screen

No

No

### DIFF
--- a/server/views/applications/new.njk
+++ b/server/views/applications/new.njk
@@ -38,7 +38,7 @@
 
         {{ govukButton({
                     name: 'submit',
-                    text: "Submit"
+                    text: "Save and continue"
                 }) }}
       </form>
     </div>
--- a/server/views/applications/new.njk
+++ b/server/views/applications/new.njk
@@ -26,7 +26,8 @@
                 isPageHeading: true
               },
               hint: {
-                html: "<p>Enter the CRN for the person needing an AP placement.</p><p>We'll import their details into your application.</p>"
+                html: "<p>Enter the CRN for the person needing an AP placement.</p><p>We'll import their details into your application.</p>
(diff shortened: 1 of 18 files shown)


server/views/applications/pages/location-factors/describe-location-factors.njk
```diff
@@ -13,7 +13,9 @@
           classes: "govuk-label--m"
         },
         hint: {
-          text: "For example SW1."
+          html: '<p>For example SW1.</p>
+          <p>We aim to accommodate someone within 50 miles of the postcode area provided.</p>
+          <p></p>'
```
pezholio, line 18: Is there a reason for this empty paragraph here?

server/views/applications/new.njk
```diff
@@ -38,7 +38,7 @@
 
         {{ govukButton({
                     name: 'submit',
-                    text: "Submit"
+                    text: "Save and continue"
```
pezholio, line 41: This has bugged me for ages! Thanks for addressing it!

pezholio (approved):
LGTM - just one minor issue

Just as a style thing, we tend not to like long commit messages (so ones that run on - I think Github Desktop gives you a warning when you go over 50 characters), if you need to give a bit more information, summarise the changes in the subject, and then go into more detail in the description. Don't stress it for now though 👍

server/views/applications/pages/location-factors/describe-location-factors.njk
beth-dixxon: That is meant to have content in it. Oh dear. I'll add it in

PR #480, opened 2023-01-12
Use a helper when testing the Assess journey

pezholio:
This updates the Assess integration tests to use a helper, similar to Apply. 

Because Assess doesn't try to load from the session like Apply does (we should probably remove this functionality), checking what tasks have been completed as we move through the journey is harder. 

With this in mind, I've also added the ability to use the data we add to the form the update the stubs as we go, so when we get to the end of one task, we call `updateAssessmentAndStub` to get the data from the form class, update our stub session and the stub `GET` application call, so the assessment we get back is up to date.

If this works well for us in Assess, we can backfill this approach for Apply too and remove the session loading code.


server/form-pages/utils/index.ts
```diff
@@ -78,6 +78,17 @@ export const getTaskName = <T>(page: T) => {
   return Reflect.getMetadata('page:task', page)
 }
 
+export const updateData = (page: TasklistPage, assessment: ApprovedPremisesAssessment): ApprovedPremisesAssessment => {
```
richpjames, line 81: Would this be better places in the `assessmentUtils.ts` file? or does that make mocking it hard when needed. 
If we are keeping it here can we make it 'updateAssessmentData'?

richpjames (approved):
LGTM!

server/form-pages/utils/index.ts
pezholio: 👍

PR #482, opened 2023-01-12
(870) Assess - Review application - download documents

richpjames:
# Context
Reviews of applications need to be able to download the documents attached to the application. 

# Changes in this PR
We build the Application review page from the 'response' that each page object in the 'Apply' stage gives us. 
For the attached files this is simply the file name and the description. In order to download the file we need its 'ID' as well. 
We could 'cheat' and add the 'ID' to the 'response' but this would show an ugly ID in the Apply Check Your Answers page and the Assess Application review stage. So instead we have to a bit more work when setting up the Review page object. 
In the Review page class need to enrich the application with the full document details so we can add a link to the document on the review page. In order to do this we fetch all the documents for the application. We then iterate through the selected documents attached to the application and collect the full document objects for the selected documents from the documents we fetched from the API as we need the ‘id’ of each document.
We then replace the original ‘selectedDocuments’ array in the application with our newly generated ‘selectedDocuments’ array.

As well as this we add some links to the bottom of the OASys and Prison Information sections that can be used when we have built out the pages for them.

### Before
image

### After

# Release checklist

confirm:

## Pre merge checklist

    advance?

## Post merge checklist

considering it done.

fix it as soon as possible.

### DIFF
--- a/server/form-pages/assess/reviewApplication/reviewApplicationAndDocumentsPage.test.ts
+++ b/server/form-pages/assess/reviewApplication/reviewApplicationAndDocumentsPage.test.ts
@@ -4,23 +4,23 @@ import Review from './reviewApplicationAndDocumentsPage'
 
 describe('Review', () => {
   describe('title', () => {
-    expect(new Review({}).title).toBe('Review application')
+    expect(new Review({}, assessment).title).toBe('Review application')
   })
 
   describe('body', () => {
     it('should set the body', () => {
-      const page = new Review({ reviewed: 'yes' })
+      const page = new Review({ reviewed: 'yes' }, assessment)
       expect(page.body).toEqual({ reviewed: 'yes' })
     })
   })
 
   itShouldHaveNextValue(new Review({}), '')
 
-  itShouldHavePreviousValue(new Review({}), 'dashboard')
+  itShouldHavePreviousValue(new Review({}, assessment), 'dashboard')
 
   describe('errors', () => {
     it('should have an error if there is no answer', () => {
-      const page = new Review({})
+      const page = new Review({}, assessment)
 
       expect(page.errors()).toEqual({
         reviewed: 'You must review all of the application and documents provided before proceeding',
@@ -28,7 +28,7 @@ describe('Review', () => {
     })
 
     it('should have an error if the answer is "no"', () => {
-      const page = new Review({ reviewed: 'no' })
+      const page = new Review({ reviewed: 'no' }, assessment)
 
       expect(page.errors()).toEqual({
         reviewed: 'You must review all of the application and documents provided before proceeding',
@@ -38,7 +38,7 @@ describe('Review', () => {
 
   describe('response', () => {
     it('returns an empty object', () => {
-      const page = new Review({ reviewed: 'yes' })
+      const page = new Review({ reviewed: 'yes' }, assessment)
 
       expect(page.response()).toEqual({})
     })
--- a/server/form-pages/assess/reviewApplication/reviewApplicationAndDocumentsPage.test.ts
+++ b/server/form-pages/assess/reviewApplication/reviewApplicationAndDocumentsPage.test.ts
@@ -1,8 +1,16 @@
+import { createMock } from '@golevelup/ts-jest'
+import assessmentFactory from '../../../testutils/factories/assessment'
 import { itShouldHaveNextValue, itShouldHavePreviousValue } from '../../shared-examples'
 
 import Review from './reviewApplicationAndDocumentsPage'
+import { ApplicationService } from '../../../services'
+import documentFactory from '../../../testutils/factories/document'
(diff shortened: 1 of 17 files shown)


server/form-pages/assess/reviewApplication/reviewApplicationAndDocumentsPage.test.ts
```diff
@@ -14,7 +22,42 @@ describe('Review', () => {
...
+        { fileName: 'documentB' } as Document,
+      ])
+
+      const getDocumentsMock = jest.fn().mockResolvedValue(fullSelectedDocumentObjects)
+
+      const applicationService = createMock<ApplicationService>({ getDocuments: getDocumentsMock })
+
+      const page = await Review.initialize({}, assessment, token, { applicationService })
+
+      // eslint-disable-next-line
```
pezholio, line 55: What rule are we disabling here? Can we just disable the specific rule, rather than everything?

server/form-pages/assess/reviewApplication/reviewApplicationAndDocumentsPage.test.ts
```diff
@@ -14,7 +22,42 @@ describe('Review', () => {
...
+      expect(documentsFromApplication((page as any).document.application)).toEqual(fullSelectedDocumentObjects)
```
pezholio, line 56: I'd usually expect this to be the other way round unless there's a specific reason? Also, would be good to have the results of `documentsFromApplication` as a variable, so:

```suggestion
      const expectedDocuments = documentsFromApplication((page as any).document.application)
      expect(fullSelectedDocumentObjects).toEqual(expectedDocuments)
```

server/utils/assessmentUtils.ts
```diff
@@ -200,28 +200,54 @@ const getTaskResponsesAsSummaryListItems = (
...
-        value,
-      })
+
+    if (pageName === 'attach-documents') {
+      const summaryListItems = getAttatchDocumentsSummaryListItems(application)
+      return [...prev, ...summaryListItems]
+    }
+
+    const summaryListItems = getGenericSummaryListItems(response)
+    return [...prev, ...summaryListItems]
```
pezholio, line 218: Could we do a conditional assignment here?

```suggestion
      const summaryListItems = pageName === 'attach-documents' ? getAttatchDocumentsSummaryListItems(application) : getGenericSummaryListItems(response)
    
    return [...prev, ...summaryListItems]
```

pezholio (approved):
LGTM - just a couple of things

server/form-pages/assess/reviewApplication/reviewApplicationAndDocumentsPage.test.ts
richpjames: Yeah this slipped through, I'll add `no-explicit-any`

server/form-pages/assess/reviewApplication/reviewApplicationAndDocumentsPage.test.ts
richpjames: I agree with having the result of `documentsFromApplication` as a variable but what you have called `expectedDocuments` is the thing we want to assert is like the `fullSelectedDocumentObjects` so I think it should be: 
``` const actualDocuments = documentsFromApplication((page as any).document.application)
      expect(actualDocuments).toEqual(fullSelectedDocumentObjects)
```
If that makes sense to you? I'm wondering if there is something I could do to make it a bit clearer. Maybe using the `documentsFromApplication` helper confuses things here?

pezholio: How about this?:

```
    it('should add the full Document objects given the document filenames supplied on the application', async () => {
      const allDocuments = [
        documentFactory.build({ fileName: 'documentA' }),
        documentFactory.build({ fileName: 'documentB' }),
        documentFactory.build({ fileName: 'documentC' }),
      ]
      assessment.application = overwriteApplicationDocuments(assessment.application, [
        { fileName: allDocuments[0].fileName } as Document,
        { fileName: allDocuments[1].fileName } as Document,
      ])

      const getDocumentsMock = jest.fn().mockResolvedValue(allDocuments)

      const applicationService = createMock<ApplicationService>({ getDocuments: getDocumentsMock })

      const page = await Review.initialize({}, assessment, token, { applicationService })

      // eslint-disable-next-line @typescript-eslint/no-explicit-any
      const actualDocuments = documentsFromApplication((page as any).document.application)

      expect(actualDocuments).toEqual([allDocuments[0], allDocuments[1]])
    })
```

pezholio: This way you're clearer that what you want to get back is a subset of all the documents that you're mocking.

richpjames: Yep that's better

PR #484, opened 2023-01-16
Allow tasklist helpers to work with no data

pezholio:
When an asessment is first initialized, the data is in a `null` state, so we see errors when viewing the tasklist. This adds some guards to those helpers to allow us to query the data safely.


richpjames (approved):
LGTM

PR #487, opened 2023-01-17
Task list improvements

pezholio:
This adds the tasklist-specific helpers to the taskList file, as well as a couple of improvements to ensure that they work with both applications and assessments. I'm concerned we may need to be a bit smarter further down the line when we want to work out if a task is completed / in-progress, so this will help us out.

### DIFF
--- a/server/utils/taskListUtils.ts
+++ b/server/utils/taskListUtils.ts
@@ -1,18 +1,19 @@
 import type { FormSections, FormSection } from '@approved-premises/ui'
 import { ApprovedPremisesApplication as Application, ApprovedPremisesAssessment as Assessment } from '../@types/shared'
-import { JourneyType, Task } from '../@types/ui'
+import { Task } from '../@types/ui'
 import applyPaths from '../paths/apply'
 import assessPaths from '../paths/assess'
 import Apply from '../form-pages/apply'
-
-const taskIds = Object.keys(Apply.pages)
+import Assess from '../form-pages/assess'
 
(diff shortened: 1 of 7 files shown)


server/@types/shared/models/ApprovedPremises.ts
```diff
@@ -2,11 +2,13 @@
 /* tslint:disable */
 /* eslint-disable */
 
+import type { LocalAuthorityArea } from './LocalAuthorityArea';
```
richpjames, line 5: I think these type updates should be in a different commit/PR

server/utils/taskListUtils.ts
```diff
@@ -37,22 +38,25 @@ const getCompleteSectionCount = (sections: FormSections, applicationOrAssessment
   }).length
 }
 
-const taskLink = (task: Task, applicationOrAssessment: Application | Assessment, journeyType?: JourneyType): string => {
+const isAssessment = (applicationOrAssessment: Application | Assessment): boolean => {
```
richpjames, line 41: instead of `: boolean` I think you could do `: applicationOrAssessment is Assessment` as a [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates). 
I don't think there's a major advantage of doing it here as `id` is defined on both

richpjames (approved):
LGTM

server/@types/shared/models/ApprovedPremises.ts
pezholio: Yeah, I dunno why those haven't come in a PR

PR #488, opened 2023-01-17
Fix - Finding documents in assessment review

richpjames:
We hope that every document attached to an Assessment can be retrieved from the API. However we should guard against this if this is not the case.
Previously we were using `Array.prototype.find` which returns `undefined` if the documents couldn't be found. Now we instead add documents to the `selectedDocuments` array. It's less pure but a bit easier to grok and works better!


pezholio (approved):
LGTM 👍

PR #490, opened 2023-01-18
Allow full error trace in Nunjucks

pezholio:
When a global function in a template throws an error, we don’t see the full trace of the error, which makes diagnosing issues really hard.

This makes use of the [undocumented dev argument](https://github.com/mozilla/nunjucks/issues/1430) which allows us to see the cause of the error in our logs and in Sentry.

I’ve had to extend the nunjucks types to allow for this option to be set, as it’s not in the type definition in `DefinitelyTyped`. However, I’ve [opened a PR to add this](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/63968)

### DIFF
--- a/server/@types/nunjucks/index.d.ts
+++ b/server/@types/nunjucks/index.d.ts
@@ -1,0 +1,7 @@
+export default {}
+
+declare module 'nunjucks' {
+  export interface ConfigureOptions {
+    dev?: boolean | undefined
+  }
+}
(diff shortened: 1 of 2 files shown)


richpjames:
Is it worth adding this to the [template](https://github.com/ministryofjustice/hmpps-template-typescript)? or maybe we should wait for the DefinitelyTyped PR to merge

richpjames (approved):

PR #493, opened 2023-01-18
Make the review page full width

pezholio:
The review page was a bit squished, which I think might have been a regression

## Before

image

## After

image


richpjames (approved):
LGTM

PR #495, opened 2023-01-19
(873) Use assessment statuses in listing

pezholio:
Assessments now have a "status", which we can use to glean what tab they are listed in. I've also updated the code to include the "pending" assessments (i.e. assessments that have a note without a response)


pezholio:
[Confirm if there's enough information in assessment](https://trello.com/c/Jd5xAwWR/873-confirm-if-theres-enough-information-in-assessment)

richpjames (approved):
LGTM

PR #501, opened 2023-01-20
(1017) Allow a response to be added to a clarification note

pezholio:
This allows a response to be added to a clarification note. I've had to tweak the way our custom controller update actions work slightly, but this is better, as they update the page data too, as well as carrying out any associated actions. 

After a note is completed, and the user returns to the assessment, they are then prompted to answer if the response has been received. If they click "Yes", then they can add the response, and the date it was received, they are then redirected to complete the journey. 

I still need to handle what happens when they click "no", and this will require a bit of tweaking to the tasklist, but this is the next job!

I had to deviate from the design a bit, so would appreciate feedback from @beth-dixxon and @selina-dxw too 👍 

## Screenshots

Assess -- allows me to create and update a clarification note

### DIFF
--- a/server/form-pages/assess/reviewApplication/sufficientInformation/informationReceived.ts
+++ b/server/form-pages/assess/reviewApplication/sufficientInformation/informationReceived.ts
@@ -1,0 +1,94 @@
+import type { TaskListErrors, YesOrNo, ObjectWithDateParts } from '@approved-premises/ui'
+import type { User } from '@approved-premises/api'
+
+import { Page } from '../../../utils/decorators'
+import { sentenceCase } from '../../../../utils/utils'
+import { dateAndTimeInputsAreValidDates, dateIsBlank, DateFormats } from '../../../../utils/dateUtils'
+
+import TasklistPage from '../../../tasklistPage'
+
+type InformationReceivedBody = ObjectWithDateParts<'responseReceivedOn'> & {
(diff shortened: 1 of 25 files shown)


pezholio:
[Allow a response to be added to a clarification note](https://trello.com/c/qgf9IJkh/1017-allow-a-response-to-be-added-to-a-clarification-note)

selina-dxw:
This looks good, just a couple of small changes: 

- Change spelling in question to 'practi**t**ioner'
- Change textbox text to: "Provide the additional information received from the probation practitioner"

server/form-pages/assess/reviewApplication/sufficientInformation/informationReceived.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  user: User
+
+  constructor(private _body: Partial<InformationReceivedBody>) {}
+
+  public set body(value: Partial<InformationReceivedBody>) {
+    this._body = {
+      informationReceived: value.informationReceived as YesOrNo,
+    }
+    if (value.informationReceived === 'yes') {
```
richpjames, line 40: Why do we need to do this for this page?

richpjames (approved):
LGTM. 
Have left a question - I don't think there's an issue with the code I'm just trying to understand better :)

richpjames (approved):

PR #503, opened 2023-01-24
(873) Reject assessment if no information

pezholio:
This allows a user to continue on the application journey if the additional information requested has not been recieved. This also means that users won't have to complete the "Assess Application" section, so we add some helpers to check 
if the answer to the InformationReceived question was "no". If it is, we can remove the assess question from the flow.

I've also taken the liberty of tidying up the Assess integration tests, as they were getting a little unweildy. I'm still not 100% happy with them, but they are better. I wonder if going forward we should think about using the Cucumber preprocessor for integration tests too?

## Screenshots

Assess -- should allow me to reject an application where I have not received the correct information (3)

Assess -- should allow me to reject an application where I have not received the correct information (4)

### DIFF
--- a/server/controllers/assess/assessmentsController.ts
+++ b/server/controllers/assess/assessmentsController.ts
@@ -1,8 +1,7 @@
 import type { Request, Response, RequestHandler } from 'express'
 
 import { AssessmentService } from '../../services'
-import Assess from '../../form-pages/assess'
-import { informationSetAsNotReceived } from '../../utils/assessmentUtils'
+import { getSections, informationSetAsNotReceived } from '../../utils/assessmentUtils'
 
 import paths from '../../paths/assess'
 
@@ -34,7 +33,7 @@ export default class AssessmentsController {
         res.render('assessments/show', {
           assessment,
           pageHeading: 'Assess an Approved Premises (AP) application',
-          sections: Assess.sections,
+          sections: getSections(assessment),
         })
       }
     }
--- a/server/controllers/assess/assessmentsController.ts
+++ b/server/controllers/assess/assessmentsController.ts
@@ -2,6 +2,7 @@ import type { Request, Response, RequestHandler } from 'express'
 
 import { AssessmentService } from '../../services'
 import Assess from '../../form-pages/assess'
+import { informationSetAsNotReceived } from '../../utils/assessmentUtils'
 
 import paths from '../../paths/assess'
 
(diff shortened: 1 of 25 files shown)


pezholio:
[Confirm if there's enough information in assessment](https://trello.com/c/Jd5xAwWR/873-confirm-if-theres-enough-information-in-assessment)

server/controllers/assess/assessmentsController.ts
```diff
@@ -20,7 +21,7 @@ export default class AssessmentsController {
     return async (req: Request, res: Response) => {
       const assessment = await this.assessmentService.findAssessment(req.user.token, req.params.id)
 
-      if (assessment.status === 'pending') {
+      if (assessment.status === 'pending' && !informationSetAsNotReceived(assessment)) {
```
bryangaledxw, line 23: This is making me wonder if it'd be better to have `informationReceived`, rather than needing the double-negative

server/utils/assessmentUtils.ts
```diff
@@ -114,6 +114,16 @@ const informationSetAsNotReceived = (assessment: Assessment): boolean => {
   return false
 }
 
+const getSections = (assessment: Assessment) => {
+  let { sections } = Assess
+
+  if (informationSetAsNotReceived(assessment)) {
+    sections = sections.filter(section => section.name !== 'AssessApplication')
```
bryangaledxw, line 121: Would changing `'AssessApplication'` to `AssessApplication.name` work/be safer?

server/utils/assessmentUtils.test.ts
```diff
@@ -417,4 +428,28 @@ describe('assessmentUtils', () => {
...
+      assessment.data = { 'sufficient-information': { 'information-received': { informationReceived: 'yes' } } }
+
+      const sections = getSections(assessment)
+      const sectionNames = sections.map(s => s.name)
+
+      expect(sections.length).toEqual(Assess.sections.length)
+      expect(sectionNames).toContain('AssessApplication')
+    })
+
+    it('returns all sections if informationSetAsNotReceived is true', () => {
```
bryangaledxw, line 445: Is this description correct? We're checking that it *doesn't* return all sections, right?

And to match the description, should these two tests be mocking `informationSetAsNotReceived` rather than setting the assessment data?

bryangaledxw (approved):
LGTM, couple of minor comments

server/controllers/assess/assessmentsController.ts
pezholio: Yeah, I struggled with the naming here - we want to be sure it's actually set, and it also pushes an (arguably more confusing) double negative into the `getSections` method. Maybe I'll set this as a variable, so it's clearer as to what's going on.

server/utils/assessmentUtils.ts
pezholio: Great idea! I'll try that 👍

server/utils/assessmentUtils.test.ts
pezholio: Yeah, good call. To avoid having to mock the same module, I've changed the description too, and added a test for when the answer has not been completed.

PR #506, opened 2023-01-25
Add a script to create an application/assessment

pezholio:
Creating local data for applications and assessments can be a pain sometimes, so I’ve generated a script that gets application data from the Wiremock repo and creates an application in the local API database for the dev user (JIMSNOWLDAP) and an associated assessment, again associated with the dev user.


richpjames (approved):
🚀🚀🚀

PR #507, opened 2023-01-25
(1057) Allow us to skip oasys integration with a feature flag

pezholio:
This prepares us to (potentially) skip the OASys integration if it isn't fixed in time for the beta launch by setting an `OASYS_DISABLED` environment variable. This will tell the data service to return some stub data (so we still have the titles), change some titles, tweak the flow slightly and remove some page furniture (mainly displaying risks). It also adds a flag to the API request when creating an application (handled in https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/379).

Hopefully we won't need this, but this puts the groundwork in for us to skip the integration if we need to.

### DIFF
--- a/integration_tests/mockApis/applications.ts
+++ b/integration_tests/mockApis/applications.ts
@@ -21,7 +21,7 @@ export default {
     stubFor({
       request: {
         method: 'POST',
-        url: `/applications`,
+        url: `/applications?createWithRisks=true`,
       },
       response: {
         status: 201,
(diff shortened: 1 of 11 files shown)


pezholio:
[Allow us to skip Oasys integration with a feature flag](https://trello.com/c/khI0VQrF/1057-allow-us-to-skip-oasys-integration-with-a-feature-flag)

richpjames (approved):
LGTM. 
I wonder if it would be worth adding this path to the E2E tests... I can why we wouldn't want to as we're hoping we won't have to use this and I can't think of a way to do it other than duplicating the whole Application journey. But it would be annoying if we had to use it, it was broken and our tests weren't failing. Your call!

pezholio:
I was leaning towards that, but let's wait and see if we actually need it first. Would be annoying to have to put the work in and we never use it!

PR #518, opened 2023-01-31
(1031) Assess - Submit and confirm assessment

richpjames:
# Context
This PR allows reviewers of assessments to submit their review.

[Trello ticket](https://trello.com/c/8Nytt5wz/878-confirmation-screen)

# Changes in this PR
1. There is some general refactoring
2. We start building the functionality from the client upwards through the stack. There are two different endpoints used for submitting assessments dependent on whether the assessment is accepted or rejected so we add those.
3. We add the service, this is where the logic lives to decide which endpoint to call and also to retrieve the human readable `document` that is submitted when the endpoints are called. 
4. We then add some utility functions in order to make handling the assessment decision and for manipulating the copy to show on the confirmation page easier 
5. We add the controller method that handles the submission, this has to perform some validation depending on whether the  checkbox on the Tasklist page is checked
6. We add the view which collects the correct copy to display on the confirmation page
7. We wire up the Tasklist to hit the correct controller when the submit button is clicked

## Screenshots of UI changes

# Release checklist

confirm:

## Pre merge checklist

    advance?

## Post merge checklist

considering it done.

fix it as soon as possible.


server/utils/applicationUtils.ts
```diff
@@ -39,27 +40,36 @@ const createNameAnchorElement = (name: string, applicationId: string) => {
   return htmlValue(`<a href=${paths.applications.show({ id: applicationId })}>${name}</a>`)
 }
 
-const getResponses = (application: Application): ApplicationResponse => {
+export type ApplicationOrAssessmentResponse = Record<string, Array<PageResponse>>
+
+const getResponses = (applicationOrAssessment: Application | Assessment): ApplicationOrAssessmentResponse => {
```
pezholio, line 45: Yeah, I think this is fine for now. I do wonder if we could move some of this stuff to the form / task decorators or something. But we can handle this later 👍

server/utils/assessments/utils.ts
```diff
@@ -284,6 +284,9 @@ const getSectionSuffix = (task: Task) => {
   return `<p><a href="${link}">${copy}</a></p>`
 }
 
+const decisionFromAssessment = (assessment: Assessment) =>
+  assessment?.data?.['make-a-decision']?.['make-a-decision']?.decision || ''
```
pezholio, line 288: Not for now, but I wonder if it might be less brittle if we get the response from the page objects themselves somehow, rather than walking the data in this way (as we're doing it a lot now!)

server/utils/assessments/utils.ts
```diff
@@ -287,6 +287,24 @@ const getSectionSuffix = (task: Task) => {
...
+      return `<p>We've notified the Probation Practitioner that this application has been assessed as suitable.</p>
+      <p>The assessment can now be used to match Robert Brown to a bed in an Approved Premises.</p>`
+      break
+    case 'hold':
+      return `<p>We've notified the Probation Practitioner that this application has been assessed as suitable.</p>
+      <p>This case is now paused until the oral hearing outcome has been provided by the Probation Practitioner and a release date is confirmed.</p>
+      <p>It will be added to the matching queue if the oral hearing is successful.</p>`
+      break
+    default:
+      return `<p>We've sent you a confirmation email.</p>
```
pezholio, line 302: We should probably confirm that sending confirmation emails is in scope for this round of work. I don't think it's a huge job (thanks GOV.UK Notify!), but we need to factor it in

server/controllers/assess/assessmentsController.ts
```diff
@@ -35,10 +35,41 @@ export default class AssessmentsController {
...
         })
       }
     }
   }
+
+  submit(): RequestHandler {
+    return async (req: Request, res: Response) => {
+      const assessment = await this.assessmentService.findAssessment(req.user.token, req.params.id)
+
+      if (req.body?.confirmation !== 'confirmed') {
```
pezholio, line 49: Good catch! I don't think we do this yet in Apply

server/utils/applicationUtils.ts
richpjames: Yeah that would be smart

server/controllers/assess/assessmentsController.ts
richpjames: I'll add it in later :)

PR #524, opened 2023-02-01
(1059) Submitted applications stay in the open application screen

pezholio:
This fixes the application dashboard to group applications by the newly added `status` attribute. I've also updated the tables to be more in line with the prototype.

## Screenshot

image

### DIFF
--- a/server/controllers/apply/applicationsController.test.ts
+++ b/server/controllers/apply/applicationsController.test.ts
@@ -1,7 +1,7 @@
 import type { Request, Response, NextFunction } from 'express'
 import { createMock, DeepMocked } from '@golevelup/ts-jest'
 
-import type { ErrorsAndUserInput } from '@approved-premises/ui'
+import type { ErrorsAndUserInput, GroupedApplications } from '@approved-premises/ui'
 import type { ApprovedPremisesApplication } from '@approved-premises/api'
 import ApplicationsController, { tasklistPageHeading } from './applicationsController'
 import { ApplicationService, PersonService } from '../../services'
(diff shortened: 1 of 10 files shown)


pezholio:
[Submitted applications stay in the open application screen.](https://trello.com/c/mJbwwETZ/1059-submitted-applications-stay-in-the-open-application-screen)

server/controllers/apply/applicationsController.test.ts
```diff
@@ -38,7 +38,8 @@ describe('applicationsController', () => {
 
   describe('index', () => {
     it('renders the index view', async () => {
-      const applications = applicationFactory.buildList(5)
+      const applications = { inProgress: [], requestedFurtherInformation: [], submitted: [] } as GroupedApplications
```
richpjames, line 41: I could be wrong but I don't think we need to cast here, it would be better if we could do `const applications: GroupedApplication = ...`

server/utils/applicationUtils.ts
```diff
@@ -21,13 +21,22 @@ const dashboardTableRows = (applications: Array<Application>): Array<TableRow> =
...
-      textValue(
-        application.submittedAt ? DateFormats.isoDateToUIDate(application.submittedAt, { format: 'short' }) : 'N/A',
-      ),
+      htmlValue(getStatus(application)),
     ]
   })
 }
 
+const getStatus = (application: Application): string => {
+  switch (application.status) {
```
richpjames, line 30: The statuses might not line up nicely but we have a `getStatus` function for assessments that we could potentially add these statuses too as well. Depends how DRY you want to go!

richpjames (approved):
LGTM, left a couple of comments

server/controllers/apply/applicationsController.test.ts
pezholio: Yeah, much better!

server/utils/applicationUtils.ts
pezholio: I think it's better to separate them out, especially as you'd have to do a bunch of typechecking too

PR #525, opened 2023-02-01
Update dependencies in Docker

pezholio:
Updating Shellcheck has introduced a bunch of new dependencies, including `node-gyp`, which needs `python3` and `bzip2` to be installed. This should fix the Docker build and allow the containers to be built cleanly again.


richpjames (approved):
LGTM

PR #536, opened 2023-02-07
(1063) Fix task manager bugs

pezholio:
Before, how we worked out the status of a task was rather "dumb", i.e. we checked if there was any data for the task, and assumed it was complete. This resulted in some unexpected behaviour, and also made it harder to work out if a task was partially complete.

This PR adds a new helper function that walks through the pages of a task and works out if each page in the journey is complete or not, returning a status for that task. This is then used in a new Service class, which memoizes all theses statuses for use in the frontend.

This means that we can much more reliably work out the status of a task, and also makes some of the util functionality much easier to deal with.

At the moment, in the integrations tests, Assess returns "Completed" for all the tasks, as the stubs are not working / updating in the way we expect (so we're stubbing the complete application each time). Given the time constraints we're under, I'm going to leave this for now, but I would like us to work out how our tests for the Assess journey could be more stateful, but that's not for now.

### DIFF
--- a/server/@types/ui/index.d.ts
+++ b/server/@types/ui/index.d.ts
@@ -64,8 +64,11 @@ export type Task = {
 
 export type TaskStatus = 'not_started' | 'in_progress' | 'complete' | 'cannot_start'
 
+export type TaskWithStatus = Task & { status: TaskStatus }
+
 export type FormSection = {
   title: string
+  name: string
   tasks: Array<Task>
 }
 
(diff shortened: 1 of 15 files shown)


pezholio:
[OASys import screen 🐞](https://trello.com/c/hJqPIE1t/1060-oasys-import-screen-%F0%9F%90%9E)

pezholio:
[Task manager 🐞](https://trello.com/c/FyqOusi9/1063-task-manager-%F0%9F%90%9E)

richpjames (approved):
Looks good to me, left a couple comments

PR #541, opened 2023-02-07
Apply - Add ACCT alerts

richpjames:
# Context

When viewing prison case notes and adjudications probation officers also need to see any relevant ACCT alerts there may be. 
Adding these in was missed the first time around doing Apply.

# Changes in this PR

We add in a client and service for the alerts. 
Then we add them to the view for prison information. Since the alerts are read only it's relatively simple

## Screenshots of UI changes

acct-alerts

# Release checklist

As part of our continuous deployment strategy we must ensure that this work is
ready to be released at any point. Before merging to `main` we must first
confirm:

## Pre merge checklist

- [ ] Have all changes to any dependencies been deployed to both `preprod` and
    `production` in a backwards compatible way? (This includes our API,
    infrastructure, third-party integrations and any secrets or environment variables)
- [ ] Has a required data migration been included in this change or been done in
    advance?

## Post merge checklist

Once we've merged we now need to release this through to production before
considering it done.

[Find the build-and-deploy job in CircleCI](https://app.circleci.com/pipelines/github/ministryofjustice/hmpps-temporary-accommodation-ui)
and work through the following steps:

- [ ] Has the product manager asked to provide approval for this feature?
  - [ ] Has the product manager approved?
- [ ] Manually approve release to preprod
- [ ] Verify change released to preprod
- [ ] Manually approve release to prod
- [ ] Verify change released to production

Should a release fail at any step, you as the author should now lead the work to
fix it as soon as possible.

PR #533, opened 2023-02-06
(872) Assess - Review prison information - Case notes

richpjames:
# Context
This PR adds a page for the assessor to review the prison case notes in a similar manner to, and building on top of, #529.

[Trello card](https://trello.com/c/WbPVb0XT/872-review-prison-information-details)

# Changes in this PR
First we do some housekeeping, then add a function to retrieve the case notes from the application followed by the view and UI test.

## Screenshots of UI changes
Assess -- allows me to assess an application (17)

# Release checklist

As part of our continuous deployment strategy we must ensure that this work is
ready to be released at any point. Before merging to `main` we must first
confirm:

## Pre merge checklist

- [ ] Have all changes to any dependencies been deployed to both `preprod` and
    `production` in a backwards compatible way? (This includes our API,
    infrastructure, third-party integrations and any secrets or environment variables)
- [ ] Has a required data migration been included in this change or been done in
    advance?

## Post merge checklist

Once we've merged we now need to release this through to production before
considering it done.

[Find the build-and-deploy job in CircleCI](https://app.circleci.com/pipelines/github/ministryofjustice/hmpps-temporary-accommodation-ui)
and work through the following steps:

- [ ] Has the product manager asked to provide approval for this feature?
  - [ ] Has the product manager approved?
- [ ] Manually approve release to preprod
- [ ] Verify change released to preprod
- [ ] Manually approve release to prod
- [ ] Verify change released to production

Should a release fail at any step, you as the author should now lead the work to
fix it as soon as possible.

PR #529, opened 2023-02-02
(872) Assess - Review prison information details

richpjames:
# Context
Assessors need to be able to review the prison information for the applicant's PiP.
This adds a page to the Assess 'Review Application' section allowing them to see the requisite information

[Trello card](https://trello.com/c/WbPVb0XT/872-review-prison-information-details)

# Changes in this PR

We add a page for viewing the prison information from the application (not from the API). This is very similar to the page where adjudications show in Apply. 
I have opted not to use a Form Page object for this even though, it could be argued, it is part of the form. There is no user input and I think it would complicate things if it used the same page class as proper form pages.
I haven't added the ACCT tab yet, that will come in a later PR.

## Screenshots of UI changes

Assess -- allows me to assess an application

# Release checklist

As part of our continuous deployment strategy we must ensure that this work is
ready to be released at any point. Before merging to `main` we must first
confirm:

## Pre merge checklist

- [ ] Have all changes to any dependencies been deployed to both `preprod` and
    `production` in a backwards compatible way? (This includes our API,
    infrastructure, third-party integrations and any secrets or environment variables)
- [ ] Has a required data migration been included in this change or been done in
    advance?

## Post merge checklist

Once we've merged we now need to release this through to production before
considering it done.

[Find the build-and-deploy job in CircleCI](https://app.circleci.com/pipelines/github/ministryofjustice/hmpps-temporary-accommodation-ui)
and work through the following steps:

- [ ] Has the product manager asked to provide approval for this feature?
  - [ ] Has the product manager approved?
- [ ] Manually approve release to preprod
- [ ] Verify change released to preprod
- [ ] Manually approve release to prod
- [ ] Verify change released to production

Should a release fail at any step, you as the author should now lead the work to
fix it as soon as possible.

PR #543, opened 2023-02-08
(872) Assess - ACCT alerts review tab

richpjames:
# Context

Following on from #529 #533 we need to add a tab to the review information page for ACCT alerts which were added to the Application in #541 

# Changes in this PR
1. We add a util function for retrieving ACCT alerts from an assessment 
2. In the Assessments Controller we pass the ACCT alerts through to the view 
3. Renaming of a function the integration tests
4. We move a function from the caseNotes class to the general Page class so it can be used for both the Apply and Assess tests
5. Add the ACCT alerts to the view

## Screenshots of UI changes

### DIFF
--- a/server/utils/assessments/utils.test.ts
+++ b/server/utils/assessments/utils.test.ts
@@ -518,7 +518,7 @@ describe('utils', () => {
       const assessment = assessmentFactory.build()
       assessment.application.data['prison-information'] = {}
 
-      expect(adjudicationsFromAssessment(assessment)).toEqual('')
+      expect(adjudicationsFromAssessment(assessment)).toEqual([])
     })
   })
 
@@ -535,7 +535,7 @@ describe('utils', () => {
       const assessment = assessmentFactory.build()
       assessment.application.data['prison-information'] = {}
 
-      expect(caseNotesFromAssessment(assessment)).toEqual('')
+      expect(caseNotesFromAssessment(assessment)).toEqual([])
     })
   })
 })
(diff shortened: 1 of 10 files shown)


server/utils/assessments/utils.test.ts
```diff
@@ -538,4 +540,21 @@ describe('utils', () => {
...
+      assessment.application.data['prison-information'] = { 'case-notes': { acctAlerts } }
+
+      expect(acctAlertsFromAssessment(assessment)).toEqual(acctAlerts)
+    })
+
+    it('returns an empty string if the case notes are empty', () => {
+      const assessment = assessmentFactory.build()
+      assessment.application.data['prison-information'] = {}
+
+      expect(acctAlertsFromAssessment(assessment)).toEqual('')
```
pezholio, line 557: Should this be an empty array if they're blank?

server/views/assessments/pages/risk-information/prison-information.njk
```diff
@@ -116,6 +116,31 @@
                 </div>
             </div>
 
+            <div class="govuk-column-full-width" id="acct" role="tabpanel" aria-labelledby="acctTab" hidden>
+                <table class="govuk-table">
```
pezholio, line 120: Does this show the same information as in Apply? If so, might be good to have this in a partial

pezholio (approved):
LGTM - just a couple of minor things

server/views/assessments/pages/risk-information/prison-information.njk
richpjames: I've added a macro for the whole table which takes into account the differences in the 'Prison Case Notes' tab.

PR #546, opened 2023-02-09
Fix/Assessment type guard

richpjames:
The property that we used to assert than an assessment is an assessment (allocatedToStaffMemberId) changed its name and became optional. 
This meant that we were showing the application tasklist for assessments. 
Our types should have alerted us to this error but I *think* that because we were using the `in` operator we weren't. I have replaced the `in` operator with a more typesafe check and now we are testing for the 'allocatedAt' property which always exists for assessments.
Furthermore because this was a simple type guard which only exists in TS land the test coverage check didn't alert us to the lack of tests around the functionality. So I have added some tests around the functionality too. 
Hopefully with these tests and typechecks again we will prevent reoccurence.


pezholio (approved):
LGTM 👍🏻

PR #548, opened 2023-02-09
Fix - Add proper rejectionRationale when rejecting an assessment

richpjames:
Previously we weren't adding the rejection rationale when rejecting an assessment which meant the API was rejecting. 
Here we take the human readable response and add it to the POSTed object

### DIFF
--- a/server/services/assessmentService.ts
+++ b/server/services/assessmentService.ts
@@ -1,4 +1,4 @@
-import type { Request } from 'express'
+import { Request } from 'express'
 import {
   ApprovedPremisesAssessment as Assessment,
   NewClarificationNote,
@@ -11,7 +11,7 @@ import TasklistPage, { TasklistPageInterface } from '../form-pages/tasklistPage'
 import { getBody, updateAssessmentData } from '../form-pages/utils'
 import { ValidationError } from '../utils/errors'
 import { getResponses } from '../utils/applicationUtils'
-import { applicationAccepted } from '../utils/assessments/utils'
+import { applicationAccepted, rejectionRationaleFromAssessmentResponses } from '../utils/assessments/utils'
 
 export default class AssessmentService {
   constructor(private readonly assessmentClientFactory: RestClientBuilder<AssessmentClient>) {}
@@ -73,7 +73,7 @@ export default class AssessmentService {
     const responses = getResponses(assessment)
 
     if (!applicationAccepted(assessment)) {
-      return client.rejection(assessment.id, responses, assessment.rejectionRationale)
+      return client.rejection(assessment.id, responses, rejectionRationaleFromAssessmentResponses(assessment))
     }
 
     return client.acceptance(assessment.id, responses)
--- a/server/utils/applicationUtils.test.ts
+++ b/server/utils/applicationUtils.test.ts
@@ -83,6 +83,27 @@ describe('applicationUtils', () => {
     })
   })
 
+  describe('getResponseForPage', () => {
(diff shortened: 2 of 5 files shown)


server/services/assessmentService.ts
```diff
@@ -85,7 +85,7 @@ export default class AssessmentService {
     const responses = getResponses(assessment)
 
     if (!applicationAccepted(assessment)) {
-      return client.rejection(assessment.id, responses, assessment.rejectionRationale)
+      return client.rejection(assessment.id, responses, (responses?.['make-a-decision']?.[0]?.Decision as string) || '')
```
pezholio, line 88: Could we have a function here called `extractRejectionRationale` or similar? Also, the use of this `0` index makes me a bit nervous.

richpjames: Yeah sure. 
It does feel a bit risky but I'm not too sure what else we can do to achieve the same result - there is optional chaining to ensure we fallback to an empty string but we could wrap it in a try/catch if you'd prefer

pezholio: Yeah, I guess I don't quite understand why it's coming in as an array is all.

pezholio: OIC - Couldn't we use `getResponseForPage` in that case?

server/utils/assessments/utils.test.ts
```diff
@@ -799,4 +800,20 @@ describe('utils', () => {
...
+  describe('rejectionRationaleFromAssessmentResponses', () => {
+    it('returns the rejectionRationale from the assessment when it exists', () => {
+      const assessment = assessmentFactory.build()
+
+      ;(applicationUtils.getResponseForPage as jest.Mock).mockImplementation(() => ({ Decision: 'some rationale' }))
+
+      expect(rejectionRationaleFromAssessmentResponses(assessment)).toEqual('some rationale')
+    })
+
+    it('returns an empty s tring when the rationale doesnt exists', () => {
```
pezholio, line 813: Typo here

server/utils/assessments/utils.ts
```diff
@@ -530,6 +530,14 @@ const caseNotesFromAssessment = (assessment: Assessment) =>
 const acctAlertsFromAssessment = (assessment: Assessment) =>
   assessment.application?.data?.['prison-information']?.['case-notes']?.acctAlerts || []
 
+const rejectionRationaleFromAssessmentResponses = (assessment: Assessment): string => {
+  const response = getResponseForPage(assessment, 'make-a-decision', 'make-a-decision')?.Decision || ''
+
+  if (Array.isArray(response)) return ''
```
pezholio, line 536: Why would this return an array?

server/utils/applicationUtils.test.ts
```diff
@@ -83,6 +83,27 @@ describe('applicationUtils', () => {
     })
   })
 
+  describe('getResponseForPage', () => {
```
pezholio, line 86: 👍

pezholio (approved):

server/utils/assessments/utils.ts
richpjames: For the `make-a-decision` page it will always return a string but TS doesn't know that, `getResponsesForPage` sometimes returns an array.
Rather than cast it as a string which could cause issues down the line I think it's better to guard against it being an array to keep TS happy

pezholio: Ah, nice 👍

PR #549, opened 2023-02-09
Fall back to display document type description when extendedDescription blank

edavey:
Trello:  https://trello.com/c/NM38D3Ms/1224-documenttypedescription-is-needed-if-documentextendeddescription-not-present

The `description` field is taken from the `Document.extendedDescription` which often contains an excellent description such as "Non Statutory Intervention for OPD Community Pathway on 23/08/2016". However it can be blank. In that case it's important to show the applicant more than the document name which in many cases is not very helpful.

A document name might be `e davey.pdf`. If there is no `description` to supplement this meagre information we now supply the `typeDescription`, perhaps "PNC previous convictions" which indicates that this is an important document to include in the application.

# Release checklist

confirm:

## Pre merge checklist

    advance?

## Post merge checklist

considering it done.

fix it as soon as possible.

### DIFF
--- a/server/utils/attachDocumentsUtils.test.ts
+++ b/server/utils/attachDocumentsUtils.test.ts
@@ -112,6 +112,20 @@ describe('attachDocumentsUtils', () => {
       })
     })
 
+    it('uses the typeDescription if the document description is undefined', () => {
+      const document = documentFactory.build({ id: '123', description: null })
+      const selectedDocument = documentFactory.build({
+        id: '123',
+        description: undefined,
+        typeDescription: 'PNC previous convictions',
+      })
+
+      expect(documentWithDescription(document, [selectedDocument, documentFactory.build()])).toEqual({
+        ...document,
+        description: 'PNC previous convictions',
+      })
+    })
+
     it('returns the document if the document with the same ID is not in the selectedDocuments argument', () => {
       const document = documentFactory.build({ id: '123', description: null })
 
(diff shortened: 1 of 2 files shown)


server/utils/attachDocumentsUtils.test.ts
```diff
@@ -112,6 +112,20 @@ describe('attachDocumentsUtils', () => {
       })
     })
 
+    it('uses the typeDescription if the document description is null', () => {
+      const document = documentFactory.build({ id: '123', description: null })
+      const selectedDocument = documentFactory.build({
+        id: '123',
+        description: null,
```
richpjames, line 119: I'm not sure what we will recieve from the API if the description is omitted but I can't imagine it will be a JS `null`. 
I think it's safer here to set it as `undefined`. This will cause `documentWithDescription` to break but it can be resurrected using optional chaining like so: `selectedDocument?.typeDescription`. 
The functionality will work the same if we *do* get a null from the API but we're also covered if the property isn't defined.

richpjames (commented):
Thank you for this. I have made a suggestion to improve robustness

server/utils/attachDocumentsUtils.test.ts
edavey: thanks Rich -- interesting!

I've updated the commit.

richpjames (approved):

PR #556, opened 2023-02-14
Fix/use browserify over esbuild for e2e tests

richpjames:
Esbuild stopped working due to the `watch` command being removed in
v17.0.0. In theory we could roll back to v16.5.0 but this didn’t
work locally. More details here: https://github.com/badeball/cypress-cucumber-preprocessor/issues/929.
Previously we’ve had to add polyfills for certain Node APIs for esbuild too
(see https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/515).
Due to these two issues and the fact we’re not using esbuild anywhere
else in the codebase I think it’s best we use browserify for the time
being instead as it seems to work better ‘out of the box’.


richpjames:
Yes, that's right. If you're building a big app the difference can be huge but as we were just using it for the E2Es it's a few seconds... We can always revert once things are more stable but I think the costs outweight the benefits for now

PR #557, opened 2023-02-14
(1217) - Add optional chaining to tier level

richpjames:
It is possible for someone to not have a tier value. Previously this caused a crash whereas it should direct the user to the exceptional case screen.
Using optional chaining should fix this issue. 
We may will need to change the copy on the exceptional case screen but this can be done once we have the copy ready.


edavey (approved):
This looks correct and clear to me, thks!

PR #558, opened 2023-02-14
(1211) Fix - Add back buttons

richpjames:
Adds back buttons where they were previously missing on: 
- Select offence screen
- Exceptional case screen
- Placement purpose screen
- Rehabilitative interventions screen


server/testutils/factories/application.ts
```diff
@@ -33,6 +33,16 @@ class ApplicationFactory extends Factory<ApprovedPremisesApplication> {
       },
     })
   }
+
+  withPageResponse(section: string, task: string, key: string, value: unknown) {
```
pezholio, line 37: ❤️ Love it!

pezholio: Wonder if named parameters would make this clearer. Something like:

```
withPageResponse(args: {section: string, task: string, key: string, value: unknown})
```

pezholio (approved):
LGTM - just a suggestion for that factory helper

server/testutils/factories/application.ts
richpjames: Yeah good point!

PR #559, opened 2023-02-14
ADR - Data manipulation for views should happen in util files

richpjames:
This decision about our architecture is valuable but undocumented, I thought it was worth documenting.

### DIFF
--- a/doc/architecture/decisions/0008-data-manipulation-for-views-should-happen-in-util-files.md
+++ b/doc/architecture/decisions/0008-data-manipulation-for-views-should-happen-in-util-files.md
@@ -1,0 +1,168 @@
+# 7. Manipulate data for views in util files
+
+Date: 2023-02-14
+
+## Status
+
+Accepted
+
+## Context
+
+- Nunjucks is limited as to how much manipulation it allows and therefore some kinds of manipulation can only happen in Node
+- Further, manipulation in Nunjucks is only testable with integration tests (with our current test setup) which are slow to run
+- There isn't a clear precedent in HMPPS as to where such manipulation should happen
+
+## Decision
+
+We initially did such manipulation in the services but as the app has evolved we now believe that manipulation of data for views that isn't possible in Nunjucks should happen in individual utils which are imported into the Nunjucks templates. These utils should be imported on a file-by-file (as opposed to function-by-function) basis as global into the `nunjucksSetup.ts` file
+
+## Consequences
+
+- Our view code will be cleanly separated from the application code
+- Our nunjucksSetup file will have to import and export a lot of util files
+- The amount of utils files we have will grow at a faster rate
+- Each util function lives separate to the data that it requires, meaning that it is simpler to test at the unit test level
+- As the utils are imported as whole files without specifying individual functions there is little boilerplate involved in using the utils in the views. For example if we were to import functions on a function-by-function basis we would have to a have the following for each function we import:
+
+```typescript
+import { someFunction } from "./someUtilsFile"
+
+...
+
+njkEnv.addGlobal('someFunction', someFunction)
+```
+
+Whereas when functions are import on a file-by-file basis we only need to add the above for each file which could contain many util functions
+
+## Examples
+
+### Previous
+
+```typescript
+Service:
+class MyService {
+...
+  tableRow(token){
+    const client = this.clientFactory(token)
+    const apiData = await client.all()
+
+    return apiData.sort((a, b) => a.name.localeCompare(b.name))
+      .map((i: SomeThing) => {
+        return [
+          this.textValue(i.name),
+          this.textValue(i.code),
+          this.textValue(i.count.toString()),
+          this.htmlValue(
+            `<a href="${paths.show({ id: p.id })}">View<span class="govuk-visually-hidden">about ${
+              i.name
+            }</span></a>`,
+          ),
+        ]
+      })
+  }
+...
+}
+
+Controller:
+class MyController {
+...
+  index(): RequestHandler {
+      return async (req: Request, res: Response) => {
+        const tableRows = await this.myService.tableRows(req.user.token)
+        return res.render('/index', { tableRows })
+      }
+    }
+...
+}
+```
+
+View:
+
+```njk
+{{
+      govukTable({
+        captionClasses: "govuk-table__caption--m",
+        firstCellIsHeader: true,
+        head: [
+          {
+            text: "Name"
+          },
+          {
+            text: "Code"
+          },
+          {
+            text: "Count"
+          },
+          {
+            html: '<span class="govuk-visually-hidden">Actions</span>'
+          }
+        ],
+        rows: tableRows
+      })
+    }}
+
+```
+
+### Proposed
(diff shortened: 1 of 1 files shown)


doc/architecture/decisions/0008-data-manipulation-for-views-should-happen-in-util-files.md
```diff
@@ -0,0 +1,23 @@
...
+
+Date: 2023-02-14
+
+## Status
+
+Proposed
+
+## Context
+
+Nunjucks is limited as to how much manipulation it allows and therefore some kinds of neccessary manipulation can only happen in Node.
```
pezholio, line 11: Should probably talk a bit about testing here too

doc/architecture/decisions/0008-data-manipulation-for-views-should-happen-in-util-files.md
```diff
@@ -0,0 +1,23 @@
...
+There isn't a clear precedent in HMPPS as to where such manipulation should happen.
+We initially did such manipulation in the services but as the app has evolved we now believe that
+
+## Decision
+
+Manipulation of data for views that isn't possible in Nunjucks should happen in individual utils which are imported into the Nunjucks templates.
```
pezholio, line 17: Would be good if we could talk about _how_ we import these into the template - we're starting to lean towards importing whole modules. Also, as a rule, we're importing them as globals, rather than filters

pezholio (commented):
Coulpa comments. Thanks for doing this!

doc/architecture/decisions/0008-data-manipulation-for-views-should-happen-in-util-files.md
```diff
@@ -0,0 +1,23 @@
...
+
+Proposed
+
+## Context
+
+Nunjucks is limited as to how much manipulation it allows and therefore some kinds of neccessary manipulation can only happen in Node.
+There isn't a clear precedent in HMPPS as to where such manipulation should happen.
+We initially did such manipulation in the services but as the app has evolved we now believe that
+
+## Decision
```
edavey, line 15: This is really helpful, thanks.

I think that possibly for people like me who aren't very close to the code it would be important to have an example use-case of the problem? 

Can you include some code snippets e.g. "before" and "after" to illustrate the 2 approaches and make the differences more concrete?

doc/architecture/decisions/0008-data-manipulation-for-views-should-happen-in-util-files.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Status
+
+Proposed
+
+## Context
+
+Nunjucks is limited as to how much manipulation it allows and therefore some kinds of neccessary manipulation can only happen in Node.
+There isn't a clear precedent in HMPPS as to where such manipulation should happen.
+We initially did such manipulation in the services but as the app has evolved we now believe that
```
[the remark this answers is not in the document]
richpjames: Which bit exactly? 'We initially did such manipulation in the services but as the app has evolved we now believe that...'? GH is showing lines 11-13

doc/architecture/decisions/0008-data-manipulation-for-views-should-happen-in-util-files.md
richpjames: Yeah that's a good point

doc/architecture/decisions/0008-data-manipulation-for-views-should-happen-in-util-files.md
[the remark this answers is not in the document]
pezholio: Sorry, line 13 👍

doc/architecture/decisions/0008-data-manipulation-for-views-should-happen-in-util-files.md
richpjames: I'm not clear on why we're importing them as functions over filters - it's not something I've done intentionally. Do you have any reasoning? 
I think importing whole modules makes sense as there is less boilerplate when adding a util function but I've not really thought about filters vs globals

pezholio: Filters can only accept one argument eg:

```
{{ variable | doSomething }}
```

so I've tended to steer clear of them, also, the syntax is weird

doc/architecture/decisions/0008-data-manipulation-for-views-should-happen-in-util-files.md
```diff
@@ -0,0 +1,158 @@
...
+
+We initially did such manipulation in the services but as the app has evolved we now believe that manipulation of data for views that isn't possible in Nunjucks should happen in individual utils which are imported into the Nunjucks templates. These utils should be imported on a file-by-file (as opposed to function-by-function) basis as global into the nunjucksSetup.ts file.
+
+## Consequences
+
+- Our view code will be cleanly seperated from the application code
+- Our nunjucksSetup file will have to import and export a lot of util files
+- The amount of utils files we have will grow at a faster rate
+- Each util function lives seperate to the data that it requires, meaning that it is simpler to test at the unit test level
+- As the utils are imported as whole files without specifying individual functions there is little boilerplate involved in using the utils in the views.
```
edavey, line 25: would it be helpful to illustrate or elaborate a bit on that "boilerplate"? As I don't think I'm seeing it in the example? I'm guessing it's around importing the utils files to the views?

doc/architecture/decisions/0008-data-manipulation-for-views-should-happen-in-util-files.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Examples
+
```
edavey, line 28: this is great, thanks, I really understand the decision and agree way of working now with an example.

doc/architecture/decisions/0008-data-manipulation-for-views-should-happen-in-util-files.md
richpjames: I have given an example now, hopefully it makes things clearer :)

doc/architecture/decisions/0008-data-manipulation-for-views-should-happen-in-util-files.md
```diff
@@ -0,0 +1,158 @@
...
+## Decision
+
+We initially did such manipulation in the services but as the app has evolved we now believe that manipulation of data for views that isn't possible in Nunjucks should happen in individual utils which are imported into the Nunjucks templates. These utils should be imported on a file-by-file (as opposed to function-by-function) basis as global into the nunjucksSetup.ts file.
+
+## Consequences
+
+- Our view code will be cleanly seperated from the application code
+- Our nunjucksSetup file will have to import and export a lot of util files
+- The amount of utils files we have will grow at a faster rate
+- Each util function lives seperate to the data that it requires, meaning that it is simpler to test at the unit test level
```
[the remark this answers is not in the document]
richpjames: TIL

PR #562, opened 2023-02-15
Fix postcode area guidance

pezholio:
SW1A is not a postcode area. It should be SW1. I’ve updated the guidance with this in mind. The copy is a bit wordy now though, so could use a look over @selina-dxw 

## Screenshots

image

image


richpjames (approved):

selina-dxw:
I thought I'd try and be clever and use the Westminster postcode but it's an unusual one because it has a letter...

Can we change the guidance to:

"Please provide the postcode area only. To get the postcode from the full postcode remove the last 3 characters. For example, the postcode area for SW11 4NJ is SW11."

pezholio:
@selina-dxw Ah, of course!

PR #563, opened 2023-02-15
1221 - Allow OASys section responses to be nullable

richpjames:
These inputs should be nullable but if there is a null entry, this causes the schema validator to choke.

Related: #540 #542 
[Ticket](https://trello.com/c/rrtk2YcB/1221-null-should-be-allowed-for-oasys-answers-in-applications-schema)

### DIFF
--- a/server/@types/ui/index.d.ts
+++ b/server/@types/ui/index.d.ts
@@ -7,6 +7,7 @@ import {
   ApprovedPremisesAssessment as Assessment,
   Person,
   OASysSection,
+  OASysQuestion,
   Document,
   ArrayOfOASysOffenceDetailsQuestions,
   ArrayOfOASysRiskOfSeriousHarmSummaryQuestions,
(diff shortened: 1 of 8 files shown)


server/form-pages/apply/risk-and-need-factors/oasys-import/optionalOasysSections.ts
```diff
@@ -7,8 +7,11 @@ import { DataServices } from '../../../../@types/ui'
 import { sentenceCase } from '../../../../utils/utils'
 
 interface Response {
-  needsLinkedToReoffending: Array<string> | string | Array<OASysSection>
-  otherNeeds: Array<string> | string | Array<OASysSection>
+  /** @nullable */
+  needsLinkedToReoffending: Array<string> | string | Array<OASysSection> | null
+  /** @nullable */
+
+  otherNeeds: Array<string> | string | Array<OASysSection> | null
```
pezholio, line 14: That empty line at 13 shouldn't be there.  I think it's stopped the schema generator from making either of those fields nullable

richpjames:
Closed as no longer necessary

PR #565, opened 2023-02-15
Apply - Catch a 403 error from the createApplication method

richpjames:
The API will return a 403 if the CRN is not in the users caseload. 
This PR shows an error when the 403 is thrown and informs the user that the CRN is not in their case load 

[Ticket](https://trello.com/c/q6Barsj1/1231-catch-a-403-if-a-crn-is-not-in-the-users-caseload)

## Screenshot
Apply -- shows an error message if the person is not in the users caseload

### DIFF
--- a/server/controllers/apply/applicationsController.ts
+++ b/server/controllers/apply/applicationsController.ts
@@ -3,11 +3,13 @@ import type { Request, Response, RequestHandler } from 'express'
 import TasklistService from '../../services/tasklistService'
 import ApplicationService from '../../services/applicationService'
 import { PersonService } from '../../services'
-import { fetchErrorsAndUserInput } from '../../utils/validation'
+import { catchAPIErrorOrPropogate, fetchErrorsAndUserInput } from '../../utils/validation'
 import paths from '../../paths/apply'
 import { DateFormats } from '../../utils/dateUtils'
 import Apply from '../../form-pages/apply'
 import { firstPageOfApplicationJourney, getResponses, isUnapplicable } from '../../utils/applications/utils'
+import { ApprovedPremisesApplication as Application } from '../../@types/shared'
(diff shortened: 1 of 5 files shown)


server/controllers/apply/applicationsController.ts
```diff
@@ -85,8 +86,17 @@ export default class ApplicationsController {
...
         const indexOffence = offences.find(o => o.offenceId === offenceId)
-
-        const application = await this.applicationService.createApplication(req.user.token, crn, indexOffence)
+        let application: Application
+
+        try {
+          application = await this.applicationService.createApplication(req.user.token, crn, indexOffence)
+        } catch (error) {
+          if (error.status === 403) {
+            application = error.data
```
pezholio, line 95: Why do we want to continue here? I'd rather we just had some rough copy to start with instead of this potentially getting forgotten about and causing confusion further along in the journey

richpjames: I'm not clear on what should happen: 
- redirecting to CRN entry and showing an error? 
- allowing them to continue and showing some messaging?
- something else?

pezholio: Sorry, that was bad messaging on my part. We should show an error in the same way as if the CRN has not been found. Something like `The CRN ABC123 is not in your team's caseload`

richpjames: No probs - I should have checked earlier. 
I'll have a crack at that and add an integration test

pezholio (approved):
LGTM - I wonder if we should be catching this stuff earlier on in the application (i.e at the point of search, rather than when we try to create an application), but this is Good Enough for now 👍

richpjames:
Yeah I think that would be a better UX but that would require API side changes, right?

PR #567, opened 2023-02-16
(1056) Add emergency question

pezholio:
This adds an question to ask why an applicant is making either a short notice or emergency application. Depending on if the application is emergency (less than 7 days notice) or short notice (less than 4 days notice), the copy is different. I've also added a helper to ensure the date that is provided is in the future, as this seems like it would be important!

## Screenshots

image

image

### DIFF
--- a/server/utils/dateUtils.test.ts
+++ b/server/utils/dateUtils.test.ts
@@ -1,5 +1,14 @@
 import type { ObjectWithDateParts } from '@approved-premises/ui'
-import { DateFormats, InvalidDateStringError, dateAndTimeInputsAreValidDates, dateIsBlank } from './dateUtils'
+import isPast from 'date-fns/isPast'
+import {
+  DateFormats,
+  InvalidDateStringError,
+  dateAndTimeInputsAreValidDates,
+  dateIsBlank,
+  dateIsInThePast,
+} from './dateUtils'
+
+jest.mock('date-fns/isPast')
 
 describe('DateFormats', () => {
   describe('convertIsoToDateObj', () => {
@@ -181,5 +190,19 @@ describe('dateIsBlank', () => {
     }
 
     expect(dateIsBlank(date)).toEqual(true)
+  })
+})
+
+describe('dateIsInThePast', () => {
+  it('returns true if the date is in the past', () => {
+    ;(isPast as jest.Mock).mockReturnValue(true)
+
+    expect(dateIsInThePast('2020-01-01')).toEqual(true)
+  })
+
+  it('returns false if the date is not in the past', () => {
+    ;(isPast as jest.Mock).mockReturnValue(false)
+
+    expect(dateIsInThePast('2020-01-01')).toEqual(false)
   })
 })
--- a/cypress_shared/helpers/index.ts
+++ b/cypress_shared/helpers/index.ts
@@ -1,4 +1,5 @@
 import {
+  AnyValue,
   ApprovedPremisesApplication,
   ArrayOfOASysOffenceDetailsQuestions,
   ArrayOfOASysRiskManagementPlanQuestions,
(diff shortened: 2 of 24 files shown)


pezholio:
[Add emergency questions](https://trello.com/c/yqufdlRh/1056-add-emergency-questions)

server/utils/dateUtils.test.ts
```diff
@@ -183,3 +192,17 @@ describe('dateIsBlank', () => {
     expect(dateIsBlank(date)).toEqual(true)
   })
 })
+
+describe('dateIsInThePast', () => {
+  it('returns true if the date is in the past', () => {
```
richpjames, line 197: Do we *have* to mock this? 
I hope 2020-01-01 consistently stays in the past

pezholio:
There are other screens that need to be presented to users if a case is emergency / short notice, but I've broken this out into a seperate card https://trello.com/c/Dpz6zjlr/1246-contingency-screens

server/utils/applications/getApplicationNoticeType.test.ts
```diff
@@ -0,0 +1,40 @@
...
+import applicationFactory from '../../testutils/factories/application'
+
+jest.mock('./utils')
+
+describe('getApplicationNoticeType', () => {
+  const application = applicationFactory.build({})
+
+  it('returns emergency if the arrival date is less than seven days away', () => {
+    const date = add(new Date(), { days: 4 })
+    const arrivalDate = DateFormats.dateObjToIsoDate(date)
```
richpjames, line 16: Rather than mock `getArrivalDate` could we use `updateApplicationReleaseDate` to add the date?

server/utils/applications/getApplicationNoticeType.ts
```diff
@@ -0,0 +1,19 @@
+import { ApprovedPremisesApplication as Application } from '@approved-premises/api'
+import { differenceInDays, differenceInCalendarMonths } from 'date-fns'
+import { getArrivalDate } from './utils'
+import { DateFormats } from '../dateUtils'
+
+type ApplicationType = 'emergency' | 'short_notice' | 'standard'
+
+export const getApplicationNoticeType = (application: Application): ApplicationType => {
```
richpjames, line 8: What do you think of `noticeTypeFromApplication`? I know it goes against what I've done in the past but I think we should reserve `get` prefixes for a call to the API/session

server/utils/applications/getArrivalDate.ts
```diff
@@ -0,0 +1,48 @@
+import type { ApprovedPremisesApplication as Application } from '@approved-premises/api'
+import { SessionDataError } from '../errors'
+
+export const getArrivalDate = (application: Application, raiseOnMissing = true): string | null => {
```
richpjames, line 4: This could be `arrivalDateFromApplication`? Perhaps a little clearer

richpjames (approved):
LGTM have left some minor comments

server/utils/dateUtils.test.ts
pezholio: Because we're mocking `isPast` for the other test, it makes sense to do it the other way. Also, not mocking would be more work, as we'd have to reimplement the default behaviour.

server/utils/applications/getApplicationNoticeType.test.ts
pezholio: I think I prefer our unit tests to be as [solitary](https://martinfowler.com/bliki/UnitTest.html) as possible, that way if the implementation of `getArrivalDate` changes, we don't have to update code in a ton of other places.

server/utils/applications/getApplicationNoticeType.ts
pezholio: Good call 👍

PR #568, opened 2023-02-16
(1250) Include risk widgets in assessment tasklist

pezholio:
This was in the prototype, but we didn't include it in the actual application

### Before

image

### After

image


pezholio:
[Include Tier widget as per prototype](https://trello.com/c/Q3zaKUPj/1250-include-tier-widget-as-per-prototype)

richpjames (approved):

PR #574, opened 2023-02-22
(1246) Apply - Contingency plan partners screens

richpjames:
# Context
This PR is part of a larger piece of work to add new screens which show dependent on certain answers given earlier in the questionaire. There are 4 screens and the logic to show them is relatively complex so in the interests of a simpler PR review process this PR just delivers the first slice of functionality. 
[Trello card](https://trello.com/c/Dpz6zjlr/1246-contingency-screens) 

We are adding a screen to add the Contingency Plan Partners. This will show for every user initially but later we will add conditions so it only shows for some users. The user can add a number of Contingency Plan Partners and then click 'Save and continue' to progress in the questionaire. 

# Changes in this PR 

At a high level: 

- Add the page class
- Add the view

However the designs necessitate that the user can add a partner agency, press a button and then remain on the same page in order to add further partner agencies. This means we need to do a few things that we don't have to do for other pages in the form: 
-  In order to do it we need to keep track of the `partnerAgencyDetails`and pass this in to the view and have it returned to us via hidden inputs. 
- We also need to be able to see when a partner agency has already been added to the array as when the user submits an agency they are redirected back to the page and that request is handled by the ‘show’ method of the pagesController with the same body.
- We need to use two forms one with a hidden input called 'saveAndContinue' that we use to tell if the user wants to add an agency or progress to the next page 
- When totalling up the number of requests there will be 3 from the 'Contingency plan partners' page as we submit it (2x for each new partner and 1x as a 'Save and continue'). We had previously assumed 1 page = 1 request but this page breaks that rule.

## Screenshots of UI changes

### DIFF
--- a/server/form-pages/apply/risk-and-need-factors/further-considerations/contingencyPlanPartners.test.ts
+++ b/server/form-pages/apply/risk-and-need-factors/further-considerations/contingencyPlanPartners.test.ts
@@ -1,0 +1,69 @@
+import contingencyPlanPartnersFactory from '../../../../testutils/factories/contingencyPlanPartner'
+import { itShouldHaveNextValue, itShouldHavePreviousValue } from '../../../shared-examples'
+import ContingencyPlanPartners, { Body } from './contingencyPlanPartners'
+
+describe('ContingencyPlanPartners', () => {
+  const contigencyPlanPartner1 = contingencyPlanPartnersFactory.build()
+  const contigencyPlanPartner2 = contingencyPlanPartnersFactory.build()
+
+  const body = {
+    ...contigencyPlanPartner1,
+    partnerAgencyDetails: [contigencyPlanPartner2],
+  }
+
+  describe('title', () => {
+    it('should set the title', () => {
+      const page = new ContingencyPlanPartners(body)
+
+      expect(page.title).toEqual('Contingency plans')
+    })
+  })
+
+  describe('body', () => {
+    it('should push the form fields into a new item in the partnerAgencyDetails array', () => {
+      const page = new ContingencyPlanPartners(body)
+
+      expect(page.partnerAgencyDetails).toEqual([contigencyPlanPartner2, contigencyPlanPartner1])
+    })
+  })
+
+  itShouldHaveNextValue(new ContingencyPlanPartners(body), '')
+
+  itShouldHavePreviousValue(new ContingencyPlanPartners(body), 'arson')
+
+  describe('errors', () => {
+    it('should return errors when responses are blank', () => {
+      const page = new ContingencyPlanPartners({} as Body)
+
+      expect(page.errors()).toEqual({
+        namedContact: 'You must specify a named contact',
+        partnerAgencyName: 'You must specify a partner agency name',
+        phoneNumber: 'You must specify a phone number',
+        roleInPlan: 'You must specify a role in plan',
+      })
+    })
+  })
+
+  describe('response', () => {
+    it('should return response', () => {
+      const page = new ContingencyPlanPartners(body)
+
+      expect(page.response()).toEqual({
+        'Contingency plan partners': [
+          {
+            'Named contact': contigencyPlanPartner2.namedContact,
+            'Partner agency name': contigencyPlanPartner2.partnerAgencyName,
+            'Phone number': contigencyPlanPartner2.phoneNumber,
+            'Role in plan': contigencyPlanPartner2.roleInPlan,
+          },
+          {
+            'Named contact': contigencyPlanPartner1.namedContact,
+            'Partner agency name': contigencyPlanPartner1.partnerAgencyName,
+            'Phone number': contigencyPlanPartner1.phoneNumber,
+            'Role in plan': contigencyPlanPartner1.roleInPlan,
+          },
+        ],
+      })
+    })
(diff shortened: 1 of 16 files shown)


server/form-pages/apply/risk-and-need-factors/further-considerations/contingencyPlanPartners.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  isDuplicate(record: Record<string, string>) {
+    return this.partnerAgencyDetails.some(partner =>
+      Object.entries(record).every(([key, value]) => partner[key] === value),
+    )
+  }
+
+  isValid(newPartner: PartnerAgencyDetails) {
+    return this.hasNeccessaryInputs() && !this.isDuplicate(newPartner)
+  }
```
bryangaledxw, line 70: Can these and responseForPartner be private?

server/form-pages/apply/risk-and-need-factors/further-considerations/contingencyPlanPartners.test.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  describe('body', () => {
+    it('should push the form fields into a new item in the partnerAgencyDetails array', () => {
+      const page = new ContingencyPlanPartners(body)
+
+      expect(page.partnerAgencyDetails).toEqual([contigencyPlanPartner2, contigencyPlanPartner1])
+    })
+  })
+
+  itShouldHaveNextValue(new ContingencyPlanPartners(body), '')
```
bryangaledxw: This test won't pass on this commit, right?

server/form-pages/apply/risk-and-need-factors/further-considerations/contingencyPlanPartners.test.ts
```diff
@@ -42,6 +48,23 @@ describe('ContingencyPlanPartners', () => {
...
+
+    it('should return an error when there are no partner agencies added and saveAndContinue is truthy', () => {
+      const page = new ContingencyPlanPartners({ saveAndContinue: '1' } as Body)
+
+      expect(page.errors()).toEqual({
+        partnerAgencyDetails: 'You must add at least one partner agency',
+      })
+    })
+
+    it('shouldnt return an error if partnerAgencyDetails are added saveAndContinue is truthy', () => {
```
bryangaledxw, line 60: "should not return an error if partnerAgencyDetails are added if(?) saveAndContinue is truthy"

server/form-pages/apply/risk-and-need-factors/further-considerations/contingencyPlanPartners.ts
```diff
@@ -0,0 +1,89 @@
...
+import { Page } from '../../../utils/decorators'
+
+import TasklistPage from '../../../tasklistPage'
+import { lowerCase, sentenceCase } from '../../../../utils/utils'
+
+const fields: PartnerAgencyDetails = {
+  partnerAgencyName: 'Name of partner agency',
+  namedContact: 'Named contact',
+  phoneNumber: 'Contact details (phone)',
+  roleInPlan: 'Role in contingency plan',
```
bryangaledxw, line 11: Suggest keep ordering consistent everywhere you're referencing these four fields - different orders are used below

server/utils/formUtils.ts
```diff
@@ -100,6 +100,22 @@ export function convertArrayToRadioItems(array: Array<string>, checkedItem: stri
   })
 }
 
+export function convertKeyValuePairsToSummaryListItems<T>(
+  object: T,
```
bryangaledxw, line 104: Could this have a more descriptive name? 'values'?

server/views/applications/pages/further-considerations/contingency-plan-partners.njk
```diff
@@ -0,0 +1,83 @@
...
+
+{% extends "../../../partials/layout.njk" %}
+{% set hiddenInputs %}
+{% for partnerAgency in page.partnerAgencyDetails%}
+  {% set iteration = loop.index0 %}
+  {% for id, value in partnerAgency %}
+    {% set fieldName = "partnerAgencyDetails[" + iteration + "][" + id + "]"%}
+    <input type="hidden" name="{{fieldName}}" value="{{value}}"/>
+  {%endfor %}
+{%endfor %}
```
bryangaledxw, line 15: Would add a space here for consistent formatting

server/form-pages/apply/schema.json
```diff
@@ -1298,6 +1298,55 @@
...
+                  "type": "string"
+                }
+              }
+            },
+            {
+              "type": "object",
+              "properties": {
+                "saveAndContinue": {
+                  "type": "string"
+                },
```
bryangaledxw, line 1326: This feels like it shouldn't be part of what gets sent to the API. Though I guess no sensible way around?

bryangaledxw (approved):
Looks good, and really helpful explanations in PR and comments. Thanks!

server/form-pages/apply/schema.json
richpjames: Yeah there isn't that I can think of as all of the page's 'bodyProperties' are passed to the server.

PR #580, opened 2023-02-27
(1247) Fetch important data from the application in the frontend

pezholio:
This updates the submit action to send the first class variables to the API, rather than relying on the API to extract the data from the application, which is error prone and complex when you consider how to keep the two applications in line. It makes heavy use of the very useful `retrieveQuestionResponseFromApplication`, and also introduces a `retrieveOptionalQuestionResponseFromApplication` which catches the SessionData error when we want to fetch a question response which we can't guarantee will exist.


pezholio:
[Fetch important data from the application in the frontend](https://trello.com/c/U3x8YUZH/1247-fetch-important-data-from-the-application-in-the-frontend)

cypress_shared/pages/manage/lostBedCreate.ts
```diff
@@ -14,8 +14,6 @@ export default class LostBedCreatePage extends Page {
   }
 
   public completeForm(lostBed: LostBed): void {
-    cy.get('input[name="lostBed[numberOfBeds]"]').type(lostBed.numberOfBeds.toString())
```
richpjames, line 17: Should we not use `ApprovedPremisesLostBed` which does have the number of beds on it? Or are we deprecating that?

server/utils/utils.ts
```diff
@@ -88,6 +88,23 @@ export const retrieveQuestionResponseFromApplication = <T>(
   }
 }
 
+export const retrieveOptionalQuestionResponseFromApplication = <T>(
+  application: ApprovedPremisesApplication,
+  task: string,
+  page: string,
+  question?: string,
```
richpjames, line 95: I solved this problem a different way on my branch. Wasn't sure whether to use the same function with options or a new one but I think it's better this way

server/utils/applications/getApplicationSubmissionData.ts
```diff
@@ -0,0 +1,29 @@
+import type { ApprovedPremisesApplication as Application, SubmitApplication } from '@approved-premises/api'
+import type { ReleaseTypeID } from '../../form-pages/apply/reasons-for-placement/basic-information/releaseType'
+import type { ApTypes } from '../../form-pages/apply/reasons-for-placement/type-of-ap/apType'
+
+import { retrieveOptionalQuestionResponseFromApplication, retrieveQuestionResponseFromApplication } from '../utils'
+
+export const getApplicationSubmissionData = (application: Application): SubmitApplication => {
```
richpjames, line 7: NBD but we could lose the 'get' here?

richpjames (approved):

cypress_shared/pages/manage/lostBedCreate.ts
pezholio: Oops, yes. That was a bit of a slapdash effort on my part. I think it will be deprecated eventually, but not now

server/utils/utils.ts
pezholio: Yeah, I thought about adding options too, but I think it's clearer/simpler this way

server/utils/applications/getApplicationSubmissionData.ts
pezholio: Agreed - not having `get` prefixes seems to be the pattern we're adopting now

PR #584, opened 2023-02-28
(1066) Fix - cancellation reason

richpjames:
# Context
The screenshot on the ticket shows the input missing. Whilst I couldn't reproduce that bug I did notice that the value of the 'reason' input wasn't be retained on refreshes (IE if you fill in a 'reason' but not the date, when you see the error for missing the date the reason will be gone). 

[Trello](https://trello.com/c/JixgUl9x/1066-cancel-placement-screen-is-missing-reasons-for-cancelling-placement-radios)

# Changes in this PR
- Add a stub for the `/profile` endpoint 
- Add a util to resolve the path to a property in an object 
- Use that function in the `convertObjectsToRadioItems` function


pezholio (approved):
Oh, nice catch. Like the solution too!

PR #586, opened 2023-02-28
(1247) Set `in_community` release type

pezholio:
If the release type for a person is either a community order or a bail placement, then the person is currently in the community, so we don’t ask what the release type is. In these circumstances, we return a releaseType of `in_community`


pezholio:
[Fetch important data from the application in the frontend](https://trello.com/c/U3x8YUZH/1247-fetch-important-data-from-the-application-in-the-frontend)

richpjames (approved):

PR #595, opened 2023-03-02
(1276) Check a CRN is in a user's caseload when searching at the start of an application

pezholio:
There was originally a check for this after the user confirmed the person's details, but this was not getting caught as the API was returning a 400, not a 403. After some reflection, I decided that this was too late in the flow, and we should be catching this error when searching. 

We've now added a `checkCaseload` option to the person search API endpoint, which returns a 403 if the person is not in the user's caseload. 

This PR optionally adds this flag to the search, catches the 403 and displays an error. It's worth pointing out that we may get a 403 for limited access offenders too, but we can iterate as we go with this.

### DIFF
--- a/server/data/personClient.ts
+++ b/server/data/personClient.ts
@@ -25,9 +25,17 @@ export default class PersonClient {
     this.restClient = new RestClient('personClient', config.apis.approvedPremises as ApiConfig, token)
   }
 
-  async search(crn: string): Promise<Person> {
+  async search(crn: string, checkCaseload: boolean): Promise<Person> {
+    let query = { crn } as Record<string, string | boolean>
+
+    if (checkCaseload) {
+      query = { ...query, checkCaseload: true }
+    }
+    const queryString: string = qs.stringify(query, { encode: false, indices: false })
+
+    const path = `${paths.people.search({})}?${queryString}`
     const response = await this.restClient.get({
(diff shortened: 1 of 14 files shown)


pezholio:
[API returning 400 rather than 403 when applicant not authorised to create application](https://trello.com/c/gwUTNw2m/1276-api-returning-400-rather-than-403-when-applicant-not-authorised-to-create-application)

server/data/personClient.ts
```diff
@@ -25,9 +25,10 @@ export default class PersonClient {
     this.restClient = new RestClient('personClient', config.apis.approvedPremises as ApiConfig, token)
   }
 
-  async search(crn: string): Promise<Person> {
+  async search(crn: string, checkCaseload: boolean): Promise<Person> {
+    const path = `${paths.people.search({})}?crn=${crn}${checkCaseload ? '&checkCaseload=1' : ''}`
```
richpjames, line 29: Why use `1` over `true` as the value for `checkCaseload`? 
I wonder if we could use `qs` to make this slightly neater?

richpjames (approved):
LGTM

server/data/personClient.ts
pezholio: Actually, scratch that. I'll go with `true` as I've tested it locally and it works fine 👍

PR #596, opened 2023-03-03
Add emergency application flow to integration tests

richpjames:
This isn't exhaustive as I wanted to avoid duplicated what we already have in the long, slow full Apply test but it covers the new screens.


server/testutils/addResponseToApplication.ts
```diff
@@ -2,15 +2,31 @@ import { ApprovedPremisesApplication as Application } from '../@types/shared'
 
 export const addResponseToApplication = (
```
pezholio, line 3: I wonder if this should be broken into two and have an `addResponsesToApplication` that accepts an object, to save the duplication and conditionals.

integration_tests/tests/apply/apply.cy.ts
```diff
@@ -122,7 +121,7 @@ context('Apply', () => {
     isExceptionalCasePage.clickSubmit()
 
     // Then I should be told the application is not eligible
-    Page.verifyOnPage(NotEligiblePage)
+    // Page.verifyOnPage(NotEligiblePage)
```
pezholio, line 124: Why is this commented out?

richpjames: No good reason, fixed now

server/testutils/addResponseToApplication.ts
richpjames: Yeah, seems reasonable

pezholio (approved):
LGTM 👍

PR #601, opened 2023-03-07
(1310) Add a landing page

pezholio:
This adds a dashboard / landing page, which is the first thing the user sees after logging in ([cribbed in part from CAS3](https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/69)). They will see different links depending on what they have access to. We could be smarter by redirecting a user to one service if they can only access that one service, but that can come later. 

It has also raised the issue that at the moment, we don't control access by role. This should be something we address as some point before golive

image

### DIFF
--- a/server/controllers/dashboardController.ts
+++ b/server/controllers/dashboardController.ts
@@ -1,7 +1,7 @@
 import type { Request, RequestHandler, Response } from 'express'
 import paths from '../paths/manage'
 
-export default class ApplicationController {
+export default class DashboardController {
   index(): RequestHandler {
     return (_req: Request, res: Response) => {
       res.redirect(paths.premises.index({}))
--- a/server/controllers/dashboardController.ts
+++ b/server/controllers/dashboardController.ts
@@ -1,10 +1,9 @@
 import type { Request, RequestHandler, Response } from 'express'
-import paths from '../paths/manage'
 
 export default class DashboardController {
   index(): RequestHandler {
     return (_req: Request, res: Response) => {
-      res.redirect(paths.premises.index({}))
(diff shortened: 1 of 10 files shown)


pezholio:
[Add a homepage with services that can be accessed by the user](https://trello.com/c/T4rQ3p4H/1310-add-a-homepage-with-services-that-can-be-accessed-by-the-user)

server/controllers/dashboardController.ts
```diff
@@ -1,7 +1,7 @@
 import type { Request, RequestHandler, Response } from 'express'
 import paths from '../paths/manage'
 
-export default class ApplicationController {
+export default class DashboardController {
```
richpjames, line 3: This is going to make finding the Applications controller so much easier!

server/views/dashboard/index.njk
```diff
@@ -0,0 +1,50 @@
...
+
+{% set pageTitle = applicationName %}
+{% set mainClasses = "app-container govuk-body" %}
+{% set bodyClasses = "dashboard--index" %}
+
+{% block content %}
+
+  {% include "../_messages.njk" %}
+
+  <h1>Approved Premises</h1>
```
richpjames, line 11: I think it's better to pass the title in from the controller

richpjames (approved):

server/views/dashboard/index.njk
pezholio: Good call 👍

server/controllers/dashboardController.ts
pezholio: I considered keeping the name the same, but I'm glad I didn't now! 😆

PR #606, opened 2023-03-08
Fix E2E tests for Apply and Manage

richpjames:
This tidies up some of the testing for the Partner Agency page and makes sure we submit the number of lost beds in the correct format. 
 
There are also some other miscelaneous fixes/improvements in the PR


pezholio (approved):
Nice work - some lovely quality of life fixes in the mix too 👍

PR #607, opened 2023-03-08
Update type generation action to use an app

pezholio:
This updates the Type Generation action to use a Github App, as outlined here:

https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#authenticating-with-github-app-generated-tokens

This uses the app to generate a one-time usage token, which is more broad than the included `GITHUB_TOKEN`, so when the PR opens, the CI gets run.

### DIFF
--- a/.github/workflows/generate-types.yml
+++ b/.github/workflows/generate-types.yml
@@ -12,6 +12,12 @@ jobs:
   generate-types:
     runs-on: ubuntu-latest
     steps:
+      - name: Generate Token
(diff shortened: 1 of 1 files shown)


bryangaledxw (approved):

yndajas:
Is [this doc about updating the token](https://github.com/ministryofjustice/hmpps-approved-premises-ui/blob/main/doc/how-to/replace_type_generation_pr_token.md) redundant after this change?

PR #609, opened 2023-03-08
1249 - Add a 'Rejected'/'Suitable' tag if an Assessment has been completed

richpjames:
# Context
PPs would like to be able to see if an assessment has been accepted or rejected rather than simply 'completed'

# Changes in this PR
If an assessment has been completed we check if it has a status of 'accepted' or 'rejected' and return the most suitable tag.

### Before
image

### After
Assess -- shows a list of assessments
ts of UI changes

### DIFF
--- a/server/utils/assessments/utils.ts
+++ b/server/utils/assessments/utils.ts
@@ -334,7 +334,8 @@ const daysUntilDue = (assessment: Assessment): number => {
 
 const getStatus = (assessment: Assessment): string => {
   if (assessment.status === 'completed') {
-    return `<strong class="govuk-tag govuk-tag">Completed</strong>`
+    if (assessment.decision === 'accepted') return `<strong class="govuk-tag govuk-tag--green">Suitable</strong>`
+    if (assessment.decision === 'rejected') return `<strong class="govuk-tag govuk-tag--red">Rejected</strong>`
   }
 
   if (assessment.data) {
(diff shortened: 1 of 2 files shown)


server/utils/assessments/utils.ts
```diff
@@ -334,7 +334,8 @@ const daysUntilDue = (assessment: Assessment): number => {
 
 const getStatus = (assessment: Assessment): string => {
   if (assessment.status === 'completed') {
-    return `<strong class="govuk-tag govuk-tag">Completed</strong>`
+    if (assessment.decision === 'accepted') return `<strong class="govuk-tag govuk-tag--green">Suitable</strong>`
```
pezholio, line 337: Is "Suitable" definitely the wording we want here?

pezholio (approved):
LGTM - I'm assuming you've been your usual thorough self and checked that wording - I probably wouldn't have done it! 😆

server/utils/assessments/utils.ts
richpjames: Beth says she hasn't been able to go through this with Selina yet so the colours and copy are provisional for now.

richpjames: Suitable was suggested in Slack: https://mojdt.slack.com/archives/C02SHKETZ7F/p1676460188359239 
But I think Accepted might make more sense

pezholio: No drama - let's go with as is 👍

PR #620, opened 2023-03-14
(1316) List tasks to a workflow manager

richpjames:
# Context
Workflow managers need to be able to see tasks that are allocated as well as tasks that are unallocated.
This is similar to and will in future replace the assessment lists but I think it's worth keeping the assessment lists around for now until tasks are more feature-complete.

[Trello](https://trello.com/c/UlWNtoEL/1316-list-tasks-to-a-workflow-manager)

# Changes in this PR
1. We begin with some general housekeeping
2. We add the task client
3. We add the task service
4. We add a simple controller for the view
5. We add some util functions for generating the tables in the views and handling the neccessary dates
6. We add the view

## Screenshots of UI changes
Allocated tasks
Unallocated tasks

### DIFF
--- a/cypress_shared/pages/tasks/listPage.ts
+++ b/cypress_shared/pages/tasks/listPage.ts
@@ -1,0 +1,29 @@
+import Page from '../page'
+import paths from '../../../server/paths/tasks'
+
+import { allocatedTableRows, unallocatedTableRows } from '../../../server/utils/tasks/table'
+
+import { Task } from '../../../server/@types/shared'
+import { shouldShowTableRows } from '../../helpers'
+
+export default class ListPage extends Page {
+  constructor(private readonly allocatedTasks: Array<Task>, private readonly unallocatedTasks: Array<Task>) {
+    super('Tasks')
+    this.allocatedTasks = allocatedTasks
+    this.unallocatedTasks = unallocatedTasks
+  }
+
+  static visit(allocatedTasks: Array<Task>, unallocatedTasks: Array<Task>): ListPage {
+    cy.visit(paths.index({}))
+    return new ListPage(allocatedTasks, unallocatedTasks)
+  }
+
+  shouldShowAllocatedTasks(): void {
+    shouldShowTableRows(this.allocatedTasks, allocatedTableRows)
+  }
+
+  shouldShowUnallocatedTasks(): void {
+    cy.get('a').contains('Unallocated').click()
+    shouldShowTableRows(this.unallocatedTasks, unallocatedTableRows)
+  }
+}
--- a/integration_tests/tests/tasks/list.cy.ts
+++ b/integration_tests/tests/tasks/list.cy.ts
@@ -1,0 +1,31 @@
+import ListPage from '../../../cypress_shared/pages/tasks/listPage'
+
+import taskFactory from '../../../server/testutils/factories/task'
+
+context('Tasks', () => {
+  beforeEach(() => {
+    cy.task('reset')
+    cy.task('stubSignIn')
+  })
+
+  it('shows a list of tasks', () => {
+    cy.task('stubAuthUser')
+
+    // Given I am logged in
+    cy.signIn()
+
+    const allocatedTasks = taskFactory.buildList(5)
+    const unallocatedTasks = taskFactory.buildList(5, { allocatedToStaffMember: undefined })
+
+    cy.task('stubTasks', [...allocatedTasks, ...unallocatedTasks])
+
+    // When I visit the tasks dashboard
+    const listPage = ListPage.visit(allocatedTasks, unallocatedTasks)
+
+    // Then I should see the tasks that are allocated
+    listPage.shouldShowAllocatedTasks()
+
+    // And the tasks that are unallocated
+    listPage.shouldShowUnallocatedTasks()
+  })
+})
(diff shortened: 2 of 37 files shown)


server/utils/taskUtils.ts
```diff
@@ -1,9 +1,84 @@
 import { Task } from '../@types/shared'
+import { TableRow } from '../@types/ui'
+import { DateFormats } from './dateUtils'
+import { sentenceCase } from './utils'
 
-type GroupedTask = { allocated: Array<Task>; unallocated: Array<Task> }
+const DUE_DATE_APPROACHING_DAYS_WINDOW = 3
+
+const allocatedTableRows = (tasks: Array<Task>): Array<TableRow> => {
```
pezholio, line 8: I think it's fine to be repetitive here, rather than reusing the result of `unallocatedTableRows` as that `splice` there seems like a magic number and makes the code difficult to parse. I did wonder about doing the reverse, but removing items by an index seems equally magical 🪄 

Maybe we could have small functions for each row item, i.e:

```
const personRow = (task: Task) => {
  return {
    text: task?.person?.name || '',
  }
}

etc, etc
```

And then call them like so:

```
const allocatedTableRows = (tasks: Array<Task>): Array<TableRow> => {
    return tasks.map(t => [
      personRow(t),
     ....
    ])
}
```

server/utils/dateUtils.ts
```diff
@@ -102,6 +103,14 @@ export class DateFormats {
       [`${key}`]: DateFormats.dateObjToIsoDate(date),
     } as ObjectWithDateParts<K>
   }
+
+  /**
+   * @param dateString an ISO date string.
+   * @returns the date as a timestamp, useful when sorting.
+   */
+  static isoToTimestamp(dateString: string) {
```
pezholio, line 108: 👍

cypress_shared/pages/tasks/listPage.ts
```diff
@@ -0,0 +1,34 @@
+import Page from '../page'
+import paths from '../../../server/paths/tasks'
+// import { shouldShowTableRows } from '../../helpers'
+// import {
+//   awaitingAssessmentTableRows,
+//   completedTableRows,
+//   requestedFurtherInformationTableRows,
+// } from '../../../server/utils/assessments/utils'
```
pezholio, line 8: Does this need to be commented out now?

integration_tests/tests/tasks/list.cy.ts
```diff
@@ -0,0 +1,32 @@
+import ListPage from '../../../cypress_shared/pages/tasks/listPage'
+// import Page from '../../../cypress_shared/pages/page'
```
pezholio, line 2: Same here

integration_tests/tests/tasks/list.cy.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    const allocatedTasks = taskFactory.buildList(5)
+    const unallocatedTasks = taskFactory.buildList(5, { allocatedToStaffMember: undefined })
+
+    cy.task('stubTasks', [...allocatedTasks, ...unallocatedTasks])
+
+    // When I visit the tasks dashboard
+    const listPage = ListPage.visit(allocatedTasks, unallocatedTasks)
+
+    // // Then I should see the tasks that are allocated
```
pezholio, line 26: Does this need to be double commented?

pezholio (commented):
Just a couple of things, mainly tidying up stuff

server/utils/taskUtils.ts
richpjames: Yeah I was on the fence about whether to couple the two. I wanted to show that the `allocatedTableRows` is just the `unallocatedTableRows` with an person that it's allocated to and avoid having to duplicate all the logic for handling if an optional property isn't defined. 
I quite like your suggestion, will give it a go

server/utils/taskUtils.ts
```diff
@@ -1,9 +1,90 @@
...
+  const rows: Array<TableRow> = []
+
+  tasks.forEach(task => {
+    rows.push([nameCell(task), daysUntilDueCell(task), allocationCell(task), statusCell(task), taskTypeCell(task)])
```
pezholio, line 12: 👌

pezholio: This is much clearer 👍

server/utils/taskUtils.ts
```diff
@@ -1,9 +1,90 @@
 import { Task } from '../@types/shared'
+import { TableCell, TableRow } from '../@types/ui'
+import { DateFormats } from './dateUtils'
+import { sentenceCase } from './utils'
 
-type GroupedTask = { allocated: Array<Task>; unallocated: Array<Task> }
+const DUE_DATE_APPROACHING_DAYS_WINDOW = 3
+
+const allocatedTableRows = (tasks: Array<Task>): Array<TableRow> => {
```
pezholio, line 8: I wonder if this new code should be in a `tasks` subfolder, then we can have all the table-related code in a file of it's own (something like `table.ts`), then we can import it in an `index.ts`, so we can use the entire module all in one. Just keeps things a bit tidier is all.

pezholio (approved):
LGTM - just one tiny thing about the utils, but not a biggie by any stretch

server/utils/taskUtils.ts
richpjames: Yeah much better isn't it

richpjames:
[List Tasks to a workflow manager](https://trello.com/c/UlWNtoEL/1316-list-tasks-to-a-workflow-manager)

richpjames:

PR #621, opened 2023-03-15
Refactor question retrieval

pezholio:
This refactors the `retrieveQuestionResponseFromApplication` and `retrieveOptionalQuestionResponseFromApplication` methods to support Assessments too, as well as using the actual Page classes to use the decorator metadata to determine the name of the task and the page, rather than hardcoding. 

There's a lot of commits here, but they're mainly just replacing the calls to the new helpers. Rather than building the application data, we're mocking, which we have to do, as the tests don't have access to all the metadata at run time. This is alos better practice as we're making our tests less sociable.

### DIFF
--- a/server/utils/retrieveQuestionResponseFromApplicationOrAssessment.ts
+++ b/server/utils/retrieveQuestionResponseFromApplicationOrAssessment.ts
@@ -1,0 +1,60 @@
+import {
+  ApprovedPremisesApplication as Application,
+  ApprovedPremisesAssessment as Assessment,
+} from '@approved-premises/api'
+import { TasklistPageInterface } from '../form-pages/tasklistPage'
+import { getPageName, pageDataFromApplicationOrAssessment } from '../form-pages/utils'
+import { SessionDataError } from './errors'
+import { camelCase } from './utils'
+
+/**
+ * Retrieves response for a given question from the application object or throws an error if it does not exist.
+ * @param applicationOrAssessment the application or assessment to fetch the response from.
+ * @param Page the page to retrieve the response from.
+ * @param {string} question the question that we need the response for. Defaults to the camel-cased name of the `Page`.
+ * @returns the response for the given Page/question.
+ */
+export const retrieveQuestionResponseFromApplicationOrAssessment = (
+  applicationOrAssessment: Application | Assessment,
+  Page: unknown,
+  question?: string,
+) => {
+  const pageData = pageDataFromApplicationOrAssessment(Page as TasklistPageInterface, applicationOrAssessment)
+  const pageName = getPageName(Page)
+  const q = question || camelCase(pageName)
+
+  if (!pageData) {
+    throw new SessionDataError(`Question ${q} was not found in the session`)
+  }
+
+  const response = pageData[q]
+
+  if (!response) {
+    throw new SessionDataError(`Question ${q} was not found in the session`)
+  }
+
+  return response
+}
+
+/**
+ * Retrieves response for a given question from the application object or returns undefined if it does not exist.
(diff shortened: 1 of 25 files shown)


server/utils/retrieveQuestionResponseFromApplicationOrAssessment.ts
```diff
@@ -0,0 +1,62 @@
...
+  ApprovedPremisesAssessment as Assessment,
+} from '@approved-premises/api'
+import { TasklistPageInterface } from '../form-pages/tasklistPage'
+import { getPageDataFromApplicationOrAssessment, getPageName } from '../form-pages/utils'
+import { SessionDataError } from './errors'
+import { camelCase } from './utils'
+
+/**
+ * Retrieves response for a given question from the application object or throws an error if it does not exist.
+ * @param application the application to fetch the response from.
```
richpjames, line 12: These params are out of date now

server/utils/retrieveQuestionResponseFromApplicationOrAssessment.ts
```diff
@@ -0,0 +1,62 @@
...
+  if (!response) {
+    throw new SessionDataError(`Question ${q} was not found in the session`)
+  }
+
+  return response as T
+}
+
+/**
+ * Retrieves response for a given question from the application object or returns undefined if it does not exist.
+ * @param application the application to fetch the response from.
```
richpjames, line 42: These are also outdated

richpjames: I wonder if there is something we can do to ensure the jsdocs are correct at a type level for the function they document. 
Maybe [eslint-plugin-jsdoc](https://www.npmjs.com/package/eslint-plugin-jsdoc) can help?

server/utils/retrieveQuestionResponseFromApplicationOrAssessment.ts
```diff
@@ -0,0 +1,62 @@
...
+    throw new SessionDataError(`Question ${q} was not found in the session`)
+  }
+
+  const response = pageData[q]
+
+  if (!response) {
+    throw new SessionDataError(`Question ${q} was not found in the session`)
+  }
+
+  return response as T
```
richpjames, line 37: I know this was already what we were doing in the previous function so happy for it to go in as is but not sure it's wise to cast this as `T`. 
We might use it to trick ourselves into believing a variable is a type that it's not

richpjames (commented):
Looks good generally! Have left some comments around the new function(s)

server/utils/retrieveQuestionResponseFromApplicationOrAssessment.ts
```diff
@@ -0,0 +1,59 @@
...
+ * @param page the page that we need the response for in camelCase.
+ * @param {string} question [question=page] the page that we need the response for. Defaults to the value of `page`.
+ * @returns the response for the given task/page/question.
+ */
+export const retrieveQuestionResponseFromApplicationOrAssessment = <T>(
+  applicationOrAssessment: ApprovedPremisesApplication,
+  Page: unknown,
+  question?: string,
+) => {
+  const pageData = getPageDataFromApplicationOrAssessment(Page as TasklistPageInterface, applicationOrAssessment)
```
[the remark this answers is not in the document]
pezholio: Yeah, some of the pages don't match the `TasklistPageInterface` for some reason. I thought it'd be better to just cast rather than getting stuck in the weeds.

server/utils/retrieveQuestionResponseFromApplicationOrAssessment.ts
pezholio: Yeah, good call

PR #623, opened 2023-03-16
(1362) Don't ask for matching info when rejecting applications

pezholio:
In the last PR #622, I noticed that we were still asking for matching info when rejecting applications. This didn't seem like something we'd want to do, so I've updated the `getSections` method to filter this section out when an application has been rejected. There was also a weird bug that was causing the tasklist to render incorrectly when the user does not receive information back from a PP in a timely manner, so I've fixed that too.

This builds on #622, so is in draft until that is merged.

### DIFF
--- a/cypress_shared/helpers/assess.ts
+++ b/cypress_shared/helpers/assess.ts
@@ -223,12 +223,12 @@ export default class AseessHelper {
     tasklistPage.shouldShowTaskStatus('required-actions', 'Completed')
   }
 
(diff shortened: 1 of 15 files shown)


pezholio:
[Don't ask for matching info when rejecting applications](https://trello.com/c/WGWvLv45/1362-dont-ask-for-matching-info-when-rejecting-applications)

richpjames (approved):

PR #625, opened 2023-03-16
Run tests in Parallel

pezholio:
This breaks all the parts of our test suite into constituent parts, so we can run them in parallel. It doesn’t save a ton of time (maybe 2 minutes at the most), but ensures we get quick feedback from any failures, and keeps things just that little bit tidier.

We also share a bunch of the setup in a new shared action, which installs the dependencies, generates the types, and runs typecheck.

### DIFF
--- a/.github/shared/setup/action.yml
+++ b/.github/shared/setup/action.yml
@@ -1,0 +1,25 @@
+name: Build project
+description: Specific steps to setup the repo
+runs:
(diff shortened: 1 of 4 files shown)


richpjames (commented):
I like where this is going!

richpjames (approved):
LGTM 
🏎️

PR #627, opened 2023-03-17
(1317) Allow a workflow manager to allocate a task

richpjames:
# Context

Now that we have a task view (#620) workflow managers need to be able to allocate a task from the list to someone.
To do this we add a link to the list of tasks that links to a screen where they can see some information about the application related to the task and assign the task to a list of users.

This is very similar to the allocations work done in #554 but we can't show as much details as those screens showed as we only have the application ID and not the assessment ID.

[Trello ticket](https://trello.com/c/3GNg5hD7/1317-allow-a-workflow-manager-to-allocate-a-task)

# Changes in this PR
1. The first 6 commits are housekeeping and tidying from #620
2. Following this we add the 'show' method to the tasks controller which shows the form where the allocation happens
3. We add the 'create' method to the tasks controller to create the allocation 
4. We add the link to the task dashboard created in #620 that enabled to 'show' method from step 2 to be used
5. We add a util to show a summary of the application to the top of the 'show' view
6. We add the view for 'show'
7. We remove the functionality that we have now duplicated from Assess

## Screenshots of UI changes

Tasks -- allows me to allocate a task (1)
Tasks -- allows me to allocate a task


server/utils/tasks/table.ts
```diff
@@ -19,7 +27,13 @@ const unallocatedTableRows = (tasks: Array<Task>): Array<TableRow> => {
   const rows = [] as Array<TableRow>
 
   tasks.forEach(task => {
-    rows.push([nameCell(task), daysUntilDueCell(task), statusCell(task), taskTypeCell(task)])
+    rows.push([
+      nameCell(task),
+      daysUntilDueCell(task),
+      statusCell(task),
+      taskTypeCell(task),
+      allocationLinkCell(task, 'Allocate'),
```
pezholio, line 35: Love this rows.push approach, makes things so much clearer!

server/utils/tasks/index.ts
```diff
@@ -21,4 +25,36 @@ const groupByAllocation = (tasks: Array<Task>) => {
 
   return result
 }
-export { groupByAllocation, allocatedTableRows, unallocatedTableRows }
+
+const applicationSummary = (application: Application): Array<SummaryListItem> => {
```
pezholio, line 29: Might be useful to get who is currently allocated, as we do have that information (via `allocatedToStaffMember`)

pezholio (approved):
LGTM - just a tiny thing. Might be barking up the wrong tree though!

server/utils/tasks/table.ts
richpjames: Agreed! 
Something to take forward to any other tables we do.

server/utils/tasks/index.ts
pezholio: Oh, course, we're fetching the application at the point aren't we? D'oh!

richpjames: Yeah, it's a good idea though.  
I feel like it's something the user would want/need to see at this point

richpjames:

PR #632, opened 2023-03-20
Fix - submitting an allocation

richpjames:
Previously we were omitting a `taskType` in the call to the `/applications/{applicationId}/allocations` endpoint and it was required. 
We do not know what type an application could be in `taskController.show` method. 
We could pass the type via the URL or through a flash but we may also want to implement an API endpoint that we can call to get the next neccessary taskType for the application in question. Whilst we decide the best way forward we can hardcode the `taskType` to be 'Assessment'


richpjames:

richpjames:

pezholio (approved):

PR #634, opened 2023-03-21
Remove non existent endpoints

pezholio:
In manage, we were calling endpoints which had not been implemented after cancelling a booking / marking a booking as departed. Rather than implementing the endpoints, or changing the implementation of existing endpoints, I thought it'd be easier to show a flash message for now, as we'll be revisiting Manage soon anyway.


richpjames (approved):
Seems reasonable!

PR #652, opened 2023-03-24
(1369) Support mutliple assessment types for allocations

pezholio:
This updates the allocations endpoints to support multiple types of task. I've had a bit of a change around with some of the controllers to be a bit more RESTful, and there's some tweaks I'd like to make at the API end, but this gets the functionality working and the tests passing for now.

### DIFF
--- a/server/@types/shared/models/Task.ts
+++ b/server/@types/shared/models/Task.ts
@@ -8,10 +8,10 @@ import type { TaskType } from './TaskType';
 
 export type Task = {
     applicationId: string;
-    person?: Person;
-    dueDate?: string;
-    allocatedToStaffMember?: ApprovedPremisesUser;
-    status?: 'not_started' | 'in_progress' | 'complete';
-    taskType?: TaskType;
+    person: Person;
+    dueDate: string;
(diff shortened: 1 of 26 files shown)


server/@types/shared/models/Task.ts
```diff
@@ -8,10 +8,10 @@ import type { TaskType } from './TaskType';
...
 export type Task = {
     applicationId: string;
-    person?: Person;
-    dueDate?: string;
-    allocatedToStaffMember?: ApprovedPremisesUser;
-    status?: 'not_started' | 'in_progress' | 'complete';
-    taskType?: TaskType;
+    person: Person;
+    dueDate: string;
+    allocatedToStaffMember: ApprovedPremisesUser;
```
richpjames, line 13: This makes things a lot easier :D

server/utils/tasks/table.ts
richpjames, line 1: Now you've removed all these null checks you may be able to remove the `/* istanbul ignore file */` in `server/utils/tasks/index.ts` 🤞

richpjames (approved):
Looks good!

PR #653, opened 2023-03-24
More temporary Pact files to project root

pezholio:
This is a more logical place for them to live. Also I've added a setup task to delete the `tmp` file at the start of the test run, to prevent stale files from giving us false positives/negatives.


richpjames (approved):
LGTM 
Helm lint seems to be stuck for some reason

PR #658, opened 2023-03-27
(1327) Search for a premises

richpjames:
# Context
Matchers will need to be able to search for a bed from a placement request.
To begin with this will be done by taking the placement requirements as filled in by an assessor and prepopulating a search to the [bed search](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/main/src/main/resources/static/api.yml#L1692) API endpoint. At a later point the matcher will be able to adjust the parameters that are used to search for premises

# Changes in this PR
1. Various small fixes/refactorings
2. We add the client to call the search endpoint
3. We add utils for converting the search params that the UI recieves to ones that API can use
4. We add the bed service
5. We add the bed controller
6. We add a util for mapping the Premises characterisitics into a more human readable format 
8. We add the view
9. We add a link from the placement request index page to the new search screen 

## Screenshots of UI changes


server/utils/matchUtils.ts
```diff
@@ -24,3 +28,60 @@ export const mapApiCharacteristicForUi = (characteristic: CharacteristicPair) =>
 export const mapApiCharacteristicsForUi = (characteristics: Array<CharacteristicPair>) => {
   return `<ul class="govuk-list">${characteristics.map(mapApiCharacteristicForUi).join('')}</ul>`
 }
+
+export const summaryCardRows = (
+  bedSearchResult: BedSearchResult,
+): Array<{ key: { text: string }; value: { html: string } | { text: string } }> => {
+  return [
+    townRow(bedSearchResult),
```
pezholio, line 36: Love this approach 👍

server/utils/matchUtils.test.ts
```diff
@@ -37,4 +47,17 @@ describe('matchUtils', () => {
       expect(mapApiCharacteristicForUi(pipe)).toBe('<li>PIPE</li>')
     })
   })
+
+  describe('summaryCardsRow', () => {
+    it('calls the correct row functions', () => {
+      const bedSearchResult = bedSearchResultFactory.build()
+      expect(summaryCardRows(bedSearchResult.results[0])).toEqual([
+        townRow(bedSearchResult.results[0]),
```
pezholio, line 55: Should we test the individual row methods here?

cypress_shared/helpers/index.ts
```diff
@@ -47,7 +47,16 @@ const riskToSelfSummariesFromApplication = (
 }
 
 const tableRowsToArrays = (tableRows: Array<TableRow>): Array<Array<string>> => {
-  return tableRows.map(row => Object.keys(row).map(i => (row[i].html ? Cypress.$(row[i].html).text() : row[i].text)))
+  return tableRows.map(row => Object.keys(row).map(i => uiObjectValue(row[i])))
+}
+
+const uiObjectValue = (tableCell: TableCell) => {
```
pezholio, line 53: Oh, this is _very_ useful! I think I did a similar thing in my PR, but didn't extract it to a helper

cypress_shared/helpers/index.ts
```diff
@@ -47,7 +47,16 @@ const riskToSelfSummariesFromApplication = (
...
+  const tableCellHtml = tableCell as HtmlItem
+  if (typeof tableCellHtml.html === 'string') {
```
pezholio, line 55: Would `if ('html' in item) {` be better here than casting?

pezholio (approved):
This is awesome stuff - thanks! Just a couple of minor things I think

server/services/bookingService.test.ts
```diff
@@ -56,10 +57,13 @@ describe('BookingService', () => {
   })
 
   describe('listOfBookingsForPremisesId', () => {
+    let bookings: Array<Booking>
```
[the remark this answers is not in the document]
richpjames: I'm not 100% sure it will work but lets see what happens

cypress_shared/helpers/index.ts
richpjames: Yes

server/utils/matchUtils.test.ts
richpjames: Our test coverage checker doesn't count them as functions however I think it would be worthwhile to test that the complete row is as we expect - will add it in

server/controllers/dashboardController.test.ts
```diff
@@ -1,5 +1,8 @@
 import type { NextFunction, Request, Response } from 'express'
 import { DeepMocked, createMock } from '@golevelup/ts-jest'
+import applyPaths from '../paths/apply'
+import assessPaths from '../paths/tasks'
```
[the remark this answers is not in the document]
richpjames: Ah ok, that makes sense. 
The link was broken (most likely by me) but I can fix it

server/services/bookingService.test.ts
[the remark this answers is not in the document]
richpjames: Ok, didn't work in CI :( 
My new approach is to create the bookings async and await them, doesn't add much complexity as the test is already async

PR #660, opened 2023-03-28
(1372) Show read-only view of an Application

pezholio:
This shows a read-only view of an Application once it has been completed. We should probably also only show Applications to anyone who isn't the owner of an Application once they have been completed, but this is probably the responsibility of the API.

I've also updated the summaryLists to use the new `box` attribute included in GOV.UK frontend. It was a bit more complex than I first envisaged, so if anything's not clear, or you want me to put this stuff in another PR - let me know!

## Screenshots

image
(Read-only view)

image
(Using new `box` attribute on Assess)


pezholio:
[Show read-only view of Application](https://trello.com/c/6xGOPgul/1372-show-read-only-view-of-application)

server/utils/reviewUtils.test.ts
```diff
@@ -109,6 +109,27 @@ describe('reviewSections', () => {
         pages: { 'basic-information': {}, 'type-of-ap': {} },
       },
       application,
+      true,
+    )
+  })
+
+  it('calls the rowFunction for each task with the task, application and showActions if specified', () => {
+    const application = applicationFactory.build()
+    const spy = jest.fn()
```
richpjames, line 141: naming it `rowFunctionSpy` would make this a bit clearer

server/utils/assessments/utils.test.ts
```diff
@@ -62,6 +63,7 @@ jest.mock('../reviewUtils')
 jest.mock('./documentUtils')
```
richpjames, line 63: I don't think we need to mock `summaryListUtils` any more?

richpjames (approved):
Looks good, is there any scss we can remove too? Or is it still all needed?

PR #663, opened 2023-03-29
ADR - using beds instead of spaces

adamhughes86:
Creating a new ADR for clarity on Beds and Spaces


doc/architecture/decisions/0010-use-beds-instead-of-spaces-for-matching.md
```diff
@@ -0,0 +1,23 @@
...
+Date: 2023-03-29
+
+## Status
+
+Accepted
+
+## Context
+
+There are different approaches that can be made to match people into AP's and the approach chosen will
+the UI, API, and business processes. Generally it comes down to matching people into beds or spaces within an AP.
```
richpjames, line 12: I think this first sentence is missing a word '...the approach chosen will effect the UI...' maybe?

richpjames (commented):
very useful to have this clarified and documented, thanks

doc/architecture/decisions/0010-use-beds-instead-of-spaces-for-matching.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Decision
+
+We are going to match to beds. The API should serve up beds when matching and people should be booked directly into
+a bed. This is to make sure people are booked into the correct place and reduce risk of overbooking. Beds will be booked directly by the matcher which should speed up and optimise the booking process. This should improve occupancy rates. AP Managers will be able to appeal a booking if there are concerns, and they will have the ability to move people around their AP.
+
+## Consequences
+
+Some AP Manager's have raised concerns during research about loosing control of their AP's. They will intentionally be loosing control of the booking process but will retain the ability to move people around.
```
pezholio, line 21:
```suggestion
Some AP Managers have raised concerns during research about losing control of their AP's. They will intentionally be losing control of the booking process but will retain the ability to move people around.
```

doc/architecture/decisions/0010-use-beds-instead-of-spaces-for-matching.md
```diff
@@ -0,0 +1,23 @@
...
+
+There were also concerns around additonal appeals being made and therefore more work. The appeals process on bookings may increase but this is expected. It should be mitigated by the streamlined approach to bookings. Appeals should primarily be made if there is an issue with the people within an AP.
```
pezholio, line 23: Maybe add an example as to why there might be an issue (e.g. risk of conflict between residents and/or staff)

doc/architecture/decisions/0010-use-beds-instead-of-spaces-for-matching.md
adamhughes86: Good catch

doc/architecture/decisions/0010-use-beds-instead-of-spaces-for-matching.md
adamhughes86: Good idea

PR #668, opened 2023-03-30
(1369) Use new task endpoint

pezholio:
The `/tasks/{id}` endpoint now returns allocatable users, negating the need for us to do this stuff on the frontend. Frustratingly, the contract tests didn't pick this up, so we may have to go back to the drawing board on these 😢


pezholio:
[Update assessment allocations](https://trello.com/c/pRd8wzWQ/1369-update-assessment-allocations)

PR #673, opened 2023-04-03
(1388) Search screen iteration

richpjames:
# Context
On #654 we added functionality for a user to be able to search for a bed from the details of a placement request.
This PR builds on those changes and adds information about the person the placement request is for and the ability to change the parameters of the search.
Following this we will add the ability to filter by essential/desired characteristics/AP type

# Changes in this PR
1. We rename the bedController to bedSearchController - this more accurately reflects what the controller is doing and aligns better with Temporary Accommodation.
2. Add the read-only 'identity' section to the search results page
3. Add a util to retrieve the placement's start date
4. We add the search inputs to the UI

## Screenshots of UI changes

### DIFF
--- a/cypress_shared/pages/page.ts
+++ b/cypress_shared/pages/page.ts
@@ -212,4 +212,14 @@ export default abstract class Page {
         })
     })
   }
+
+  clearDateInputs(prefix: string): void {
+    cy.get(`#${prefix}-day`).clear()
+    cy.get(`#${prefix}-month`).clear()
+    cy.get(`#${prefix}-year`).clear()
+  }
+
+  getTextInputByIdAndClear(id: string): void {
+    cy.get(`#${id}`).clear()
+  }
 }
--- a/integration_tests/tests/match/match.cy.ts
+++ b/integration_tests/tests/match/match.cy.ts
@@ -1,7 +1,12 @@
 import ListPage from '../../../cypress_shared/pages/match/listPlacementRequestsPage'
 import SearchPage from '../../../cypress_shared/pages/match/searchPage'
 
-import { bedSearchResultFactory, personFactory, placementRequestFactory } from '../../../server/testutils/factories'
+import {
+  bedSearchParametersFactory,
+  bedSearchResultFactory,
+  personFactory,
+  placementRequestFactory,
+} from '../../../server/testutils/factories'
 import Page from '../../../cypress_shared/pages/page'
 
 context('Placement Requests', () => {
@@ -33,9 +38,39 @@ context('Placement Requests', () => {
     listPage.clickFindBed(placementRequests[0])
 
     // Then I should be taken to the search page
-    const searchPage = Page.verifyOnPage(SearchPage)
+    const searchPage = Page.verifyOnPage(SearchPage, person.name)
 
     // And I should see the search results
+    let numberOfSearches = 0
     searchPage.shouldDisplaySearchResults(bedSearchResults)
+    numberOfSearches += 1
+
+    const newSearchParameters = bedSearchParametersFactory.build()
+
+    // When I enter details on the search page
+    searchPage.changeSearchParameters(newSearchParameters)
+    searchPage.clickSubmit()
+    numberOfSearches += 1
+
+    cy.task('verifySearchSubmit').then(requests => {
+      expect(requests).to.have.length(numberOfSearches)
+
+      const initialSearchRequestBody = JSON.parse(requests[0].body)
+      const secondSearchRequestBody = JSON.parse(requests[1].body)
+
+      expect(initialSearchRequestBody).to.contain({
+        durationDays: placementRequests[0].duration,
+        startDate: placementRequests[0].expectedArrival,
+        postcodeDistrict: placementRequests[0].location,
+        maxDistanceMiles: placementRequests[0].radius,
+      })
+
+      expect(secondSearchRequestBody).to.contain({
+        durationDays: newSearchParameters.durationDays,
+        startDate: newSearchParameters.startDate,
+        postcodeDistrict: newSearchParameters.postcodeDistrict,
+        maxDistanceMiles: newSearchParameters.maxDistanceMiles,
+      })
+    })
   })
 })
--- a/server/controllers/match/bedSearchController.ts
+++ b/server/controllers/match/bedSearchController.ts
@@ -1,7 +1,13 @@
 import type { Request, RequestHandler, Response } from 'express'
+
 import { BedSearchParametersUi } from '../../@types/ui'
+import matchPaths from '../../paths/match'
+import assessPaths from '../../paths/assess'
+import applyPaths from '../../paths/apply'
 import { PersonService } from '../../services'
 import BedService from '../../services/bedService'
+
+import { startDateFromParams } from '../../utils/matchUtils'
 
 export default class BedSearchController {
   constructor(private readonly bedService: BedService, private readonly personService: PersonService) {}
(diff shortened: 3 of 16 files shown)


server/controllers/match/bedSearchController.ts
```diff
@@ -1,41 +1,38 @@
...
+        }
 
-        const person = await this.personService.findByCrn(req.user.token, crn as string)
+        params.startDate = startDateFromParams(params)
 
         res.render('match/search', {
           pageHeading: 'Find a bed',
-          bedSearchResults,
-          person,
+          bedSearchResults: await this.bedService.search(req.user.token, {
```
pezholio, line 28: I think I'd prefer this to be declared as a variable outside the call the `render`

server/controllers/match/bedSearchController.ts
```diff
@@ -1,41 +1,38 @@
...
+            ...params,
+          } as BedSearchParametersUi),
+          person: await this.personService.findByCrn(req.user.token, params.crn as string),
```
pezholio, line 31: Same here ☝️

server/controllers/match/bedSearchController.ts
```diff
@@ -1,41 +1,38 @@
...
+          formPath: paths.beds.search({}),
+          ...params,
+          expectedArrivalUi: params.expectedArrival && DateFormats.isoDateToUIDate(params.expectedArrival),
```
pezholio, line 34: As this is UI-specific, could we call it in the view itself? Would save the slightly awkward variable naming too.

pezholio: It also doesn't appear to be called in the template?

richpjames: [It is called](https://github.com/ministryofjustice/hmpps-approved-premises-ui/blob/eb0bc711566f5b82ed4e9b4f1e79da6d3b5981d1/server/views/match/search.njk#L40) but agree that it would be better to do the manipulation in the view

pezholio:
Also https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/675 will introduce the assessment / applications IDs, so we can use the real thing 👍

server/views/match/search.njk
```diff
@@ -5,12 +5,60 @@
...
+			</div>
+		</div>
+		<div class="search-identity-table__column search-identity-table__table-cell__last-col">
+			<div class="search-identity-table__table-cell">
+				<dt>
+					<strong>Assessment</strong>
 
+				</dt>
+				<dd>
+					<a href="/">View</a>
```
pezholio, line 50: We can use the `applicationId` and `assessmentId` from the placement request here now

pezholio: Also, for accessibility's sake, we should have some hidden text, ie:

```
<a href="">
  View <span class="govuk-visually-hidden">Assessment</span>
</a>
```

server/views/match/search.njk
```diff
@@ -5,12 +5,60 @@
...
+				</dd>
+			</div>
+			<div class="search-identity-table__table-cell">
+				<dt>
+					<strong>Application</strong>
+				</dt>
+				<dd>
+					<a href="/">View</a>
```
pezholio, line 58: Same here re: the hidden text

integration_tests/tests/match/match.cy.ts
```diff
@@ -33,9 +38,31 @@ context('Placement Requests', () => {
...
+
+    // When I enter details on the search page
+    searchPage.changeSearchParameters(newSearchParameters)
+    searchPage.clickSubmit()
+    numberOfSearches += 1
+
+    cy.task('verifySearchSubmit').then(requests => {
+      expect(requests).to.have.length(numberOfSearches)
+
+      const body = JSON.parse(requests[1].body)
```
pezholio, line 58: Should we verify the payload of both requests? And should `body` have a more descriptive name, i.e. `newSearchRequestBody` or something?

server/controllers/match/bedSearchController.ts
```diff
@@ -1,41 +1,37 @@
...
   search(): RequestHandler {
     return async (req: Request, res: Response) => {
       const query = Object.keys(req.query).length ? req.query : undefined
+      const body = Object.keys(req.body).length ? req.body : undefined
+
+      if (query || body) {
+        const params = {
+          ...query,
+          ...body,
+        }
```
pezholio, line 22: Could we put this call above the conditional and then have `if (params) {` in the conditional? There might be some weird TS/JS reason why not, so it's no biggie if you'va already tried it / it doesn't work.

server/controllers/match/bedSearchController.ts
```diff
@@ -1,41 +1,37 @@
...
-
-        const person = await this.personService.findByCrn(req.user.token, crn as string)
+        const bedSearchResults = await this.bedService.search(req.user.token, params as BedSearchParametersUi)
+        const person = await this.personService.findByCrn(req.user.token, params.crn as string)
 
         res.render('match/search', {
           pageHeading: 'Find a bed',
           bedSearchResults,
           person,
+          formPath: paths.beds.search({}),
```
pezholio, line 33: Is there a reason for passing this to the template instead of using it directly?

server/views/match/search.njk
```diff
@@ -58,7 +62,74 @@
...
+
+			{{ govukInput({
+				id: "postcodeDistrict",
+				name: "postcodeDistrict",
+				value: postcodeDistrict,
+				label: {
+				text: "Preferred location"
+				},
+				classes: "govuk-input--width-3"
+     		}) }}
```
pezholio, line 109: The alignment is a bit off here - is this the autoformatter just doing something janky, or is there an unclosed HTML tag somewhere?

pezholio (commented):
Couple of wee things, but looking good so far 👍

server/controllers/match/bedSearchController.ts
pezholio: Fair enough 👍

server/views/match/search.njk
richpjames: I think it's just the diff on Github, it looks ok locally for me

server/views/match/search.njk
```diff
@@ -1,15 +1,137 @@
...
+				<dd>
+					<a href="{{assessmentPath}}">View <span class="govuk-visually-hidden">assessment</span></a>
+				</dd>
+			</div>
+			<div class="search-identity-table__table-cell">
+				<dt>
+					<strong>Application</strong>
+				</dt>
+				<dd>
+					<a href="{{applicationPath}}">View <span class="govuk-visually-hidden">Application</span></a>
```
pezholio, line 58: Probably best to have lower case `Application` here, as it's lowercase for `assessment`

server/controllers/match/bedSearchController.ts
pezholio: 🙌

cypress_shared/pages/page.ts
```diff
@@ -212,4 +212,14 @@ export default abstract class Page {
         })
     })
   }
+
+  clearDateInputs(prefix: string): void {
```
pezholio, line 216: Noice!

pezholio (approved):
LGTM 🚢

cypress_shared/pages/page.ts
richpjames: copied from TA. 
Didn't realise this was neccessary until @patrickjfl made me aware Cypress doesn't do it automatically

PR #687, opened 2023-04-06
(1058) Update ESAP screens with new questions and flow

pezholio:
This adds another couple of screens to the ESAP flow. 

When someone selects an ESAP AP, they are first asked if the person is managed by the National Security Division. If the answer is no, then they are asked if they have agreed the case with the Community Head of Public Protection. If the answer is yes, they are prompted for details (in a similar way to if the Tier is not valid) and they can continue with the flow. If the answer is no, they are told they cannot proceed with an ESAP application. 

They can then choose to continue with the application or exit to process. If they continue, they're taken back to choose an AP type.

## Screenshots

Apply -- allows completion of the ESAP flow

Apply -- allows completion of the ESAP flow (1)

Apply -- allows completion of the ESAP flow (2)

Apply -- allows completion of the ESAP flow (3)

### DIFF
--- a/cypress_shared/helpers/apply.ts
+++ b/cypress_shared/helpers/apply.ts
@@ -454,6 +454,22 @@ export default class ApplyHelper {
     tasklistPage.shouldShowTaskStatus('oasys-import', 'Not started')
   }
 
+  completeEsapFlow() {
+    // And I should be able to start the next task
+    cy.get('[data-cy-task-name="type-of-ap"]').click()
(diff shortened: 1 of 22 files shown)


pezholio:
[Update ESAP screens with new questions and flow](https://trello.com/c/IJFzqOAF/1058-update-esap-screens-with-new-questions-and-flow)

richpjames (approved):
LGTM!

PR #688, opened 2023-04-06
Fix/Search bugs

richpjames:
A couple of bug fixes I spotted whilst doing the demo: 
1. The incorrect start date was being used causing a crash
3. Ensure the start date input is populated correctly

### DIFF
--- a/server/controllers/match/bedSearchController.ts
+++ b/server/controllers/match/bedSearchController.ts
@@ -8,6 +8,7 @@ import { PersonService } from '../../services'
 import BedService from '../../services/bedService'
 
 import { startDateFromParams } from '../../utils/matchUtils'
+import { objectIfNotEmpty } from '../../utils/utils'
 
 export const placementCriteria = [
   'isIAP',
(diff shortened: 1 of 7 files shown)


server/controllers/match/bedSearchController.ts
```diff
@@ -7,7 +7,7 @@ import applyPaths from '../../paths/apply'
 import { PersonService } from '../../services'
 import BedService from '../../services/bedService'
 
-import { startDateFromParams } from '../../utils/matchUtils'
```
pezholio, line 10: This doesn't seem relevant to this commit message - bad rebase?

richpjames: Yeah that commit was no longer needed

server/controllers/match/bedSearchController.ts
```diff
@@ -37,8 +38,8 @@ export default class BedSearchController {
 
   search(): RequestHandler {
     return async (req: Request, res: Response) => {
-      const query = (Object.keys(req.query).length ? req.query : undefined) as unknown as BedSearchParametersUi
-      const body = (Object.keys(req.body).length ? req.body : undefined) as unknown as BedSearchParametersUi
+      const query = objectIfNotEmpty(req.query) as unknown as BedSearchParametersUi
```
pezholio, line 41: 👌 - nice!

pezholio: Could we add a generic to do the casting too maybe? So:

```typescript
const query = objectIfNotEmpty<BedSearchParametersUi>(req.query)
```

pezholio (approved):
LGTM - just one tiny suggestion, but might not actually be possible!

PR #693, opened 2023-04-11
Fix - flaky Match integration test

richpjames:
if there is only one required characteristic then selectedRequiredCharacteristics & requiredCharacteristics will be a string and not an array, this change ensures they will always be an array


pezholio (approved):
LGTM 👍

PR #706, opened 2023-04-13
(1069) Apply - transgender and complex case flow

richpjames:
# Context

We need to ask some questions at the start of the journey if the PiP is transgender or if they have a transgender history.

[Trello card](https://trello.com/b/ps9bUN4V/approved-premises-team-board)

# Changes in this PR
There's a number of new screens we've added including a couple that are 'dead ends' in that the application can no longer continue. We should think of a way to delete the application from the DB if neccessary too otherwise it could hang about indefinitely. 

## Screenshots of UI changes
boardTakenPlacePage
complexCaseBoardPage
maleApPage
transgenderPage
boardTakenPlacePage (4)](https://user-images.githubusercontent.com/44123869/231818281-fe84f633-60a8-408b-84af-99fa71e38611.png)
350.png)

# Release checklist

confirm:

## Pre merge checklist

    advance?

## Post merge checklist

considering it done.

fix it as soon as possible.


integration_tests/pages/apply/isPersonTransgender.ts
```diff
@@ -0,0 +1,17 @@
+import { ApprovedPremisesApplication } from '../../../server/@types/shared'
+import ApplyPage from './applyPage'
+
+export default class IsPersonTrangender extends ApplyPage {
+  constructor(application: ApprovedPremisesApplication) {
+    super(
+      `Is ${application.person.name} transgender or do they have a transgender history?`,
```
pezholio, line 7: Just wondering if we can automatically glean this from the `genderIdentity` field? For example if their `sex` is Female, but their `genderIdentity` is Male (or vice versa - not even sure how we'd deal with non-binary folks), then we know we need to ask the question. It might even the worth going to the dropout part from the off if we know the person is a cisgender woman, as we don't support female APs at the mo.

I think this is probably OK to ship as it is though, but I'll check in with the team about what the idea flow would be.

server/views/applications/pages/basic-information/pause-application.njk
```diff
@@ -0,0 +1,31 @@
...
+            {{ govukBackLink({
+		        text: "Back",
+		        href:  paths.applications.pages.show({
+                    id: applicationId,
+                    task: 'basic-information',
+                    page: 'board-taken-place'
+                })
+	        }) }}
+            <h1 class="govuk-heading-l">{{page.title}}</h1>
+            <p class="govuk-body">From the information you’ve provided the person requires a complex case board needs to take place before you can apply for an Approved Premises (AP) placement.</p>
```
pezholio, line 22: Might be good for us to somehow add a `paused` status to an application for things like this. Fine to ship as-is, but something for us to ponder.

pezholio (approved):
This looks really good. Thanks for getting all the ducks in a row! Nothing blocking this from shipping as-is, but some things for us to consider.

integration_tests/pages/apply/isPersonTransgender.ts
richpjames: Yeah this is a valid point. I would be wary of automating it though, I think we need to allow for some human input on it. It doesn't seem like much hassle to answer a one yes/no question. 
I think it is worth flagging early if the applicant if a cis woman to say we don't support female APs otherwise people could waste a fair amount of time.

server/views/applications/pages/basic-information/pause-application.njk
richpjames: yeah, good point

selina-dxw:
Looks good to me! 

One small thing, Dan's asked us not to use "AP's" so can we change that one instance in the 'Refer to ndelius' screen to "AP" please?

PR #710, opened 2023-04-17
(952) Apply - Move on date

richpjames:
# Context
We have to make some changes to the move on date to discourage users from choosing their own preferred duration for placement and to encourage them to use the standard durations for a given AP.

# Changes in this PR
1. Add a departure date property to the placementDuration page so we can show it in the view
2. Add the new inputs that are needed to explain the reason for deviating from the standard duration and to add the new duration.
3. Update the view

### Before
before

### After

after

### DIFF
--- a/server/form-pages/apply/schema.json
+++ b/server/form-pages/apply/schema.json
@@ -1474,10 +1474,17 @@
         "placement-duration": {
           "type": "object",
           "properties": {
+            "differentDuration": {
+              "enum": [
+                "no",
+                "yes"
+              ],
+              "type": "string"
+            },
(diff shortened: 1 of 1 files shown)


selina-dxw:
Few small changes:
Add ‘an’ and ‘(AP)’ to ’Expected date of stay in an Approved Premises (AP)’

Add a question above the radio buttons: ‘Does this person require a different placement duration?’

Change question in ‘yes’ revealed options from ‘How many weeks will the person stay at the AP?’ to ‘How many weeks placement does this person require?’

Thanks

PR #712, opened 2023-04-18
1176 - Add 'summary' text area to Rehabilitative Interventions page

richpjames:
# Context

We need to add a text area for users to explain how the rehabilitative interventions will assist in rehabilitation. 
We also need to add the option 'Sex Worker' to the list of possible interventions.

[Trello ticket](https://trello.com/c/U9GFzgdS/1176-add-textarea-to-the-rehabilitative-activities-screen-and-add-checklist-option)

### Before
Apply -- allows completion of application emergency flow (1)

### After

### DIFF
--- a/integration_tests/fixtures/applicationData.json
+++ b/integration_tests/fixtures/applicationData.json
@@ -206,6 +206,7 @@
         "financeBenefitsAndDebt",
         "attitudesAndBehaviour",
         "abuse",
+        "sexWorking",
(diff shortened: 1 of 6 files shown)


pezholio:
I think this is fine, but not sure what "Sex worker" means in this context. Have asked in Slack.

pezholio (approved):
LGTM 👍

PR #719, opened 2023-04-19
Set default placement duration if not set

pezholio:
Since #710 the placement duration is no longer a required response, as we assume it's the default. This meant that when we submit an assessment and try to send the duration, we get an error.

This adds a new `getDefaultPlacementDurationInWeeks` helper which we use as the source of truth for the default placement lengths. If the duration is not set, we send the default length to the API.

### DIFF
--- a/server/testutils/mockQuestionResponse.ts
+++ b/server/testutils/mockQuestionResponse.ts
@@ -1,7 +1,10 @@
 import { ApprovedPremisesApplication as Application } from '@approved-premises/api'
-import { retrieveQuestionResponseFromApplicationOrAssessment } from '../utils/retrieveQuestionResponseFromApplicationOrAssessment'
+import {
+  retrieveOptionalQuestionResponseFromApplicationOrAssessment,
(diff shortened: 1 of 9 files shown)


server/testutils/mockQuestionResponse.ts
```diff
@@ -46,4 +49,29 @@ const mockQuestionResponse = ({
...
+  releaseType = 'other',
+  alternativeRadius,
+  duration,
+}: {
+  releaseType?: string
+  duration?: string
+  alternativeRadius?: string
+}) => {
+  ;(retrieveOptionalQuestionResponseFromApplicationOrAssessment as jest.Mock).mockImplementation(
+    // eslint-disable-next-line consistent-return
```
richpjames, line 62: What do you think about turning `consistent return` off globally?

richpjames (approved):
LGTM

server/testutils/mockQuestionResponse.ts
pezholio: I think it _can_ be useful. If it becomes an annoyance we can think again.

PR #722, opened 2023-04-20
Change search params to weeks

pezholio:
Before we were asking user to enter weeks in their search, but we were actually sending days to the API. This fixes this by accepting weeks on the frontend, then converting this to days before sending to the backend. I did consider changing the API to accept weeks, but I have a sneaking suspicion that things may change in the future, so have made the change on the frontend for now.


server/controllers/match/search/bedSearchController.ts
```diff
@@ -43,7 +43,7 @@ export default class BedSearchController {
       const placementRequest = await this.placementRequestService.getPlacementRequest(req.user.token, req.params.id)
       const searchParams = mapPlacementRequestToBedSearchParams(placementRequest)
 
-      const query = objectIfNotEmpty<BedSearchParametersUi>(searchParams)
+      const query = objectIfNotEmpty<BedSearchParametersUi>(searchParams as unknown as Record<string, unknown>)
```
richpjames, line 46: I think we could improve things here by changing the param type of `object` that `objectIfNotEmpty` accepts to be `object: Record<string, unknown> | T | undefined`. 
This way we can remove the `[index: string]: unknown` from `BedSearchParametersUi` *and* avoid casting here. My bad I should've realised when I wrote the function originally.

richpjames (approved):
Looks good, thanks for doing this. 
Just made a note about a change to the types but other than that it's all good

server/controllers/match/search/bedSearchController.ts
pezholio: Nice - good catch 👍

PR #724, opened 2023-04-20
(1330) Create a booking

pezholio:
This adds the final part of the booking journey to match - actually creating a booking. We add the client and service methods, as well as the controller and views to wire up the whole end to end journey

## Screenshots

Placement Requests -- shows a list of placementRequests (1)

### DIFF
--- a/integration_tests/mockApis/placementRequests.ts
+++ b/integration_tests/mockApis/placementRequests.ts
@@ -1,8 +1,9 @@
 import { SuperAgentRequest } from 'superagent'
 
 import type { PlacementRequest } from '@approved-premises/api'
-import { stubFor } from '../../wiremock'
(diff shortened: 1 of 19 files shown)


server/testutils/factories/newPlacementRequestBookingConfirmation.ts
```diff
@@ -0,0 +1,17 @@
+import { Factory } from 'fishery'
+import { faker } from '@faker-js/faker/locale/en_GB'
+
+import type { NewPlacementRequestBookingConfirmation } from '@approved-premises/api'
+
+import { DateFormats } from '../../utils/dateUtils'
+
```
richpjames, line 7: would it be worth having this in the same file as newPlacementRequestBookingFactory?

richpjames (approved):
LGTM

PR #730, opened 2023-04-24
(1464) Send first class fields every time an application is updated

pezholio:
Now https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/583 is in, we need to send first class fields to the API every time an application is updated. This also allows an `ApplicationSummary` to be returned when showing a list of applications, saving response times and memory.

The only downside to this is we now can't filter out inapplicable applications - we should arguably be doing this on the backend though, so I'll make a card for us to investigate how to do this.


pezholio:
[Send first-class fields every time an application is updated](https://trello.com/c/8Sc7QCGk/1464-send-first-class-fields-every-time-an-application-is-updated)

richpjames (approved):
Looks good. 
I'm not clear what an unapplicable application is though?

pezholio:
@richpjames At the moment, this is when `is-exceptional-case` is set to `no` (See https://github.com/ministryofjustice/hmpps-approved-premises-ui/blob/a6ea4430b27040e6d344dcfa30349950aaec58e8/server/utils/applications/utils.ts#L102), but we can potentially use this for a bunch of questions (i.e the Transgender flow)

PR #732, opened 2023-04-24
(1472) Set an application as inapplicable

pezholio:
Further to #730 and the commenting out of the filtering out of inapplicable applications, we've now changed the API to filter out inapplicable applications. With this in mind, we need to now set an `isInapplicable` field when updating an application. We can also update our `isInapplicable` helper to set any female AP applications (as part of the transgender flow) to set those applications to inapplicable too.


pezholio:
[Set an Application as inapplicable](https://trello.com/c/lXRTu9Dr/1472-set-an-application-as-inapplicable)

richpjames (approved):

PR #738, opened 2023-04-27
(1477) Improve contingency plan partners page

pezholio:
This improves the contingency plan partners page to use frontend JS, rather than having to make a call for each partner. This was causing some issues when we're starting to remove the session functionality from Apply, and also making for a more confusing journey.

For accessibility purposes, we show an initial list of partner fields, then hide the ones we don't need. This diverges from the approach in the MOJ pattern (https://design-patterns.service.justice.gov.uk/components/add-another/), which clones the elements.

### DIFF
--- a/integration_tests/fixtures/applicationData.json
+++ b/integration_tests/fixtures/applicationData.json
@@ -373,8 +373,7 @@
           "phoneNumber": "01234567892",
           "roleInPlan": "Test role 2"
         }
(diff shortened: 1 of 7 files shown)


pezholio:
[Improve contingency plan partners page](https://trello.com/c/X3W0H6Wi/1477-improve-contingency-plan-partners-page)

PR #741, opened 2023-04-27
(1473) Don't save in progress Applications in the session

pezholio:
To make the code simpler, as well as cut down on the amount of session storage in Redis, we're now going to call the API
each time we fetch an application page. This is largely straightforward, apart from the tests. To do this, we're leaning on [Wiremock scenarios](https://wiremock.org/docs/stateful-behaviour/) and the application’s data to cycle through each task and page, and build up a scenario and scenario states to return different data and set new states when each scenario gets called.

The code is a bit complex, but I've tried to break it down to make it as easy to digest as possible. Given that it's data-driven, I'm hoping we won't need to touch it too much 🤞


server/form-pages/apply/reasons-for-placement/basic-information/placementDate.ts
```diff
@@ -91,4 +91,16 @@ export default class PlacementDate implements TasklistPage {
```
richpjames, line 91: I think this commit message needs attention: 'As `noticeTypeFromApplication`
relies on data from the current page AND data from the current page, '

richpjames (approved):

server/form-pages/apply/reasons-for-placement/basic-information/placementDate.ts
pezholio: Whoops!

PR #742, opened 2023-04-27
1396 - Automated accessibility testing and fixes

richpjames:
This PR adds automated accessibility testing to the integration tests. 
We will use the [Cypress Axe package](https://www.npmjs.com/package/cypress-axe). 
We will add the test to the page object so each page visited by the integration tests is tested for common accessibility issues. 
This means that some pages which will not be shown to the user are tested (and will have to be compliant) but I think this is a small price to pay to abstracting away the accessibility checks. 
There are lots of issues that were highlighted by cypress axe that are fixed in this PR. The majority of them were minor but repeated mistakes we had made throughout the application but some needed more serious reworking of components.


server/testutils/jest.setup.ts
```diff
@@ -20,8 +22,8 @@ expect.extend({
     return {
       pass,
       message: pass
-        ? () => `expected ${received} not to match ${expected}`
-        : () => `expected ${received} to match ${expected}`,
+        ? () => `expected received not to match expected ${diffStringsUnified(expected, received)}`
+        : () => `expected received to match expected ${diffStringsUnified(expected, received)}`,
```
pezholio, line 26: This is :chefskiss: - nice work!

integration_tests/pages/page.ts
```diff
@@ -13,6 +13,8 @@ export default abstract class Page {
 
   constructor(private readonly title: string) {
     this.checkOnPage()
+    cy.injectAxe()
+    cy.checkA11y()
```
pezholio: This is an ace way to make sure we don't forget to check a page!

integration_tests/pages/page.ts
```diff
@@ -14,6 +14,7 @@ export default abstract class Page {
   constructor(private readonly title: string) {
     this.checkOnPage()
     cy.injectAxe()
+    cy.configureAxe({ rules: [{ id: 'aria-allowed-attr', reviewOnFail: true }] })
```
pezholio, line 17: I wonder if we can keep these rules in a JSON file in the root of the project, so they're easier to refer to?

pezholio (approved):
Amazing work! Thanks for taking the time to do this 👍

integration_tests/pages/page.ts
richpjames: I'm hoping this rule will be temporary and I feel like adding a seperate config file for the rules would encourage turning off rules in future rather than fixing the issues they highlight.

richpjames: Yeah that makes sense

PR #743, opened 2023-04-27
Handle non-array requiredCharacteristics

pezholio:
When one requiredCharacteristic is selected, Express “helpfully” flattens this down to a single string, rather than an array. This is not in any way helpful, and has been the cause of many a flickering integration test. To get round this, we need to cast the characteristic into an array and flatten it before sending the characteristics on their merry way to the API.


richpjames (approved):
Yeah this makes sense for now and is how we've handled it elsewhere. 
It's a potential rabbithole but I'm thinking we may be able to globally disable this qs behaviour in a similar vein to this https://github.com/expressjs/express/issues/3039

It would allow us to delete a lot of code! But probably not worth spending time on at this point

PR #747, opened 2023-05-02
(1484) Provide details for exemption application - Clicking no should kick you out of the application

pezholio:
This changes eligibility rules for exceptionDetails page, if the application is not in a valid tier and the PP has not agreed the case with a manager, then the application should be treated as ineligible.


pezholio:
[Provide details for exemption application - Clicking no should kick you out of the application](https://trello.com/c/vLMbn8Lg/1484-provide-details-for-exemption-application-clicking-no-should-kick-you-out-of-the-application)

richpjames (approved):

PR #749, opened 2023-05-02
Update create application script

pezholio:
This script has succumed to a bit of bit-rot, and no longer works. Now the API clears the database on each restart locally, it’s useful to have a script to generate an in-progress application. This uses the application data in the fixtures, updates the release date to be a date in the future and adds some questions about the tier to make a valid application.

The script used to create an assessment too, but I feel that generating an in-progress application is enough for us to do local testing.


pezholio:
Yeah, good plan. I'll do that 👍

PR #756, opened 2023-05-05
Fix save and come back later buttons

pezholio:
These now link back to the dashboards. There’s no sense saving the application as the state will have already been saved.


pezholio:
[Save and come back later buttons](https://trello.com/c/v1pUY9dW/1490-save-and-come-back-later-buttons)

richpjames (approved):

PR #759, opened 2023-05-05
Sort Oasys questions by number and letter

pezholio:
In reality, the Oasys questions are prefixed with letters, so sorting didn’t work as expected We can use `localeCompare` to sort alphabetically and numerically.

### Before

image

### After

image


pezholio:
[OASYS - Risk Management plan questions are in reverse order](https://trello.com/c/ZeI75ouv/1486-oasys-risk-management-plan-questions-are-in-reverse-order)

PR #778, opened 2023-05-15
(1517) Update placement requirements to use characteristics

pezholio:
When asking assessors for placement requirements, we used to have a bit of a hodge-podge of room / premises requirements, which didn't necessarily map 1:1 to the room and premises characteristics in the API. We've now rationalised these characteristics, and updated the requirements in the API to map 1:1 to characteristics. 

This PR now uses that generated `PlacementCriteria` type to generate a list of criteria that we can use in both the Assessment and the Matching stages, together with translations. This is all strongly-typed, so if/when the criteria in the API changes, then we'll get a TypeError in the API and know where to make the changes.

### DIFF
--- a/server/utils/assessments/placementRequestData.ts
+++ b/server/utils/assessments/placementRequestData.ts
@@ -1,4 +1,9 @@
-import { ApprovedPremisesAssessment as Assessment, PlacementCriteria, PlacementRequest } from '@approved-premises/api'
+import {
+  ApType,
+  ApprovedPremisesAssessment as Assessment,
+  PlacementCriteria,
+  PlacementRequirements,
+} from '@approved-premises/api'
 import { pageDataFromApplicationOrAssessment } from '../../form-pages/utils'
 import {
   retrieveOptionalQuestionResponseFromApplicationOrAssessment,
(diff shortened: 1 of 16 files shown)


pezholio:
[Investigate requirements and characteristics for matching](https://trello.com/c/F9sfJL4A/1517-investigate-requirements-and-characteristics-for-matching)

server/utils/assessments/placementRequestData.ts
```diff
@@ -61,55 +81,33 @@ export const criteriaFromMatchingInformation = (
   const essentialCriteria = [] as Array<PlacementCriteria>
   const desirableCriteria = [] as Array<PlacementCriteria>
 
-  placementRequirements.forEach(requirement => {
+  if (matchingInformation.apType !== 'normal') {
+    essentialCriteria.push(matchingInformation.apType)
+  }
+
+  if (matchingInformation.mentalHealthSupport.length) {
```
bryangaledxw, line 88: Any reason for this to not just be `if (matchingInformation.mentalHealthSupport.length) {`?

bryangaledxw (approved):
Admittedly I'm struggling jumping into this cold, but looks good

server/utils/assessments/placementRequestData.ts
pezholio: If it's an empty string, it will still return as a truthy value, so we have to use `length`

bryangaledxw: Empty string is falsey in JS, right?

bryangaledxw: I feel like I forget this every 6 months, then I'm like "really???" 😛

PR #780, opened 2023-05-16
(1526) Return error when a booking or lost bed overlaps

pezholio:
This borrows heavily from the Temporary Accommodation work to handle conflict errors. I've also had to add a view / endpoint to view a lost bed. This is a simple stub at the moment, just to allow users to follow the link to see the conflicting lost bed.

## Screenshots

Booking -- should show errors when there is a booking conflict

LostBed -- should show an error when there are booking conflicts

### DIFF
--- a/integration_tests/mockApis/lostBed.ts
+++ b/integration_tests/mockApis/lostBed.ts
@@ -3,8 +3,9 @@ import { Response, SuperAgentRequest } from 'superagent'
 import type { LostBed } from '@approved-premises/api'
 
 import { getMatchingRequests, stubFor } from '../../wiremock'
-import { errorStub } from '../../wiremock/utils'
+import { bedspaceConflictResponseBody, errorStub } from '../../wiremock/utils'
 import { lostBedReasons } from '../../wiremock/referenceDataStubs'
+import paths from '../../server/paths/api'
 
 export default {
   stubLostBedCreate: (args: { premisesId: string; lostBed: LostBed }): SuperAgentRequest =>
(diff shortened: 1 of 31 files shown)


pezholio:
[Return error when a booking overlaps](https://trello.com/c/2kobylGn/1526-return-error-when-a-booking-overlaps)

server/controllers/manage/lostBedsController.ts
```diff
@@ -36,6 +36,7 @@ export default class LostBedsController {
 
       const lostBed: NewLostBed = {
         ...req.body.lostBed,
+        bedId,
```
bryangaledxw, line 44: Is this intentionally its own commit?

integration_tests/mockApis/lostBed.ts
```diff
@@ -20,6 +20,19 @@ export default {
       },
     }),
 
+  stubLostBed: (args: { premisesId: string; lostBed: LostBed }): SuperAgentRequest =>
+    stubFor({
+      request: {
+        method: 'GET',
+        url: `/premises/${args.premisesId}/lost-beds/${args.lostBed.id}`,
```
bryangaledxw, line 27: I'm not doing it as "change-all-at-once" things, but in TA, I'm starting to use static-paths here, rather than plain strings

server/utils/validation.ts
```diff
@@ -1,22 +1,41 @@
...
 import errorLookup from '../i18n/en/errors.json'
 import { TasklistAPIError, ValidationError } from './errors'
+import { generateConflictBespokeError } from './bookingUtils'
 
 interface InvalidParams {
   propertyName: string
   errorType: string
 }
 
+export const generateConflictErrorAndRedirect = (
```
bryangaledxw, line 15: Yep, like this approach!

bryangaledxw: I guess the downside is validation.ts knowing about this specific error type, but 🤷‍♀️

integration_tests/components/component.ts
```diff
@@ -0,0 +1,18 @@
+export type PageElement = Cypress.Chainable<JQuery>
+
+export default abstract class Component {
+  shouldShowKeyAndValue(key: string, value: string): void {
+    cy.get('.govuk-summary-list__key').contains(key).siblings('.govuk-summary-list__value').should('contain', value)
+  }
+
+  shouldShowKeyAndValues(key: string, values: Array<string>): void {
```
bryangaledxw, line 8: Did you do something similar to these as `assertDefinition`? I had `Page` inherit `Component` to avoid duplication when pages want to use these

bryangaledxw (approved):

server/controllers/manage/lostBedsController.ts
pezholio: Yeah, this was a bug introduced by a previous feature

integration_tests/mockApis/lostBed.ts
pezholio: It's a good plan. I was just lazy and copied it from elsewhere 😁

integration_tests/components/component.ts
pezholio: Yeah, that was more blind copy/pasting. We don't need this at the moment, so I've got rid

PR #783, opened 2023-05-17
(1515) Update the assessments index page to show summaries

pezholio:
Following on from the change to the API in https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/629 the call to `/assessments` now returns a subset of assessment data using a `AssessmentSummary` object. This PR updates all the relevant parts that are touched by the Assessments index to support an AssessmentSummary, rather than a full-blown assessment.


Harriethw (approved):
LGTM!

PR #784, opened 2023-05-18
(1528) Update matching information

pezholio:
This adds some new requirement fields to the matching information, as well as some additional fields to provide an alternative placement length, and provide some information to the CRU manager (matcher). The latter field is not yet shown to the matcher in the frontend, but some API work is required to put this in place, and the PR was already big enough!

## Screenshot

image

PR #787, opened 2023-05-19
(1529) Send and display notes in placement requests

pezholio:
In #784 we included the concept of notes to sent to the matcher. The API has now been updated to accept these notes when an assessment is approved, so we can now send this data along with the rest of the placement request data, as well as display it to the matcher before they proceed to the search screen.

### Before

image

### After

image


bryangaledxw (approved):
LGTM

PR #788, opened 2023-05-19
(1529) Update the match screen filters

pezholio:
This updates the match screen to support both required and optional characteristics. If a characteristic is required, these are added in a hidden field and displayed in a Filter component similar to the MoJ component here - https://design-patterns.service.justice.gov.uk/components/filter/. I've had to tweak it a bit as our requirements are slightly different.

The characteristics are also broken up into categories to match the latest design changes.

### Before

### After

Placement Requests -- allows me to make a booking


richpjames (approved):
LGTM.

PR #795, opened 2023-05-24
(137) Update bed listing to match designs

pezholio:
This updates Manage to use a bed listing, rather than room listing to manage beds to match the latest designs in the prototype. 

## Screenshots

image

image


pezholio:
[Update bed listing to match designs](https://trello.com/c/yvB9J6KQ/137-update-bed-listing-to-match-designs)

server/views/premises/show.njk
```diff
@@ -28,9 +28,9 @@
       },
       menus: [{
         items: [ {
-          text: "Manage rooms",
+          text: "Manage beds",
```
richpjames, line 31: The commit message is a little confusing for this change. 
Should it be 'We’re now acting directly on beds, not rooms'?

richpjames (approved):

PR #797, opened 2023-05-24
Layout fixes

pezholio:
There's some pages that are missing wrapper row divs, so this PR adds them in. Also the GOV.UK design system says the Back Buttons should appear before the `main` element (https://design-system.service.gov.uk/components/back-link/#how-it-works), so I've made some changes there, as well as adding some missing back buttons.

This advice conflicts with an aceesibility rule, so I've updated the Axe rules, given that it's only minor.

### DIFF
--- a/integration_tests/pages/page.ts
+++ b/integration_tests/pages/page.ts
@@ -23,8 +23,14 @@ export default abstract class Page {
   constructor(private readonly title: string) {
     this.checkOnPage()
     cy.injectAxe()
-    // Temporary rule whilst this issue is resolved https://github.com/w3c/aria/issues/1404
-    cy.configureAxe({ rules: [{ id: 'aria-allowed-attr', reviewOnFail: true }] })
+    cy.configureAxe({
+      rules: [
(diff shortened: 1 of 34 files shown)


integration_tests/pages/page.ts
```diff
@@ -23,8 +23,14 @@ export default abstract class Page {
   constructor(private readonly title: string) {
     this.checkOnPage()
     cy.injectAxe()
-    // Temporary rule whilst this issue is resolved https://github.com/w3c/aria/issues/1404
-    cy.configureAxe({ rules: [{ id: 'aria-allowed-attr', reviewOnFail: true }] })
+    cy.configureAxe({
+      rules: [
+        // Temporary rule whilst this issue is resolved https://github.com/w3c/aria/issues/1404
+        { id: 'aria-allowed-attr', reviewOnFail: true },
+        // Ignore the "All page content should be contained by landmarks", which conflicts with GOV.UK guidance
```
richpjames, line 30: would be good to link to the GDS guidance IMO

richpjames (approved):
LGTM

PR #807, opened 2023-05-31
(176) Add submit placement request flow

richpjames:
# Context
[Trello card](https://trello.com/c/ncPgE6LL/176-add-a-flow-to-create-and-submit-a-placement-request)

Users need to be able to create Placement Requests from completed Applications.
This be in the form of a form completed via the read-only view of applications.
This PR is the first step towards having that form and does a lot of the behind-the-scenes heavy lifting to enable it.

# Changes in this PR
Following the approach used for assessments in #469 we add new client and service methods for creating and saving Placement Requests. 
We then add a 'pages controller' which allows the user to navigate through the form. 
Following this we add the first page of the form in order to prove it works through E2E tests. 
We then add the button to start this form journey from the read-only application view.

## Screenshots of UI changes
placementReasonPage

Placement Requests -- allows me to complete form

### DIFF
--- a/server/paths/match.ts
+++ b/server/paths/match.ts
@@ -3,11 +3,13 @@ import { path } from 'static-path'
 const placementRequestsPath = path('/placement-requests')
 const placementRequestPath = placementRequestsPath.path(':id')
 const placementRequestBookingsPath = placementRequestPath.path('bookings')
+const newPlacementRequestPath = placementRequestsPath.path('new')
 
 export default {
   placementRequests: {
     index: placementRequestsPath,
     show: placementRequestPath,
+    create: newPlacementRequestPath,
     bookings: {
       confirm: placementRequestBookingsPath.path('confirm'),
       create: placementRequestBookingsPath,
--- a/server/testutils/factories/placementApplication.ts
+++ b/server/testutils/factories/placementApplication.ts
@@ -1,0 +1,16 @@
+import { Factory } from 'fishery'
+import { faker } from '@faker-js/faker/locale/en_GB'
+import { PlacementApplication } from '../../@types/shared'
+import { DateFormats } from '../../utils/dateUtils'
+
+export default Factory.define<PlacementApplication>(() => ({
+  id: faker.string.uuid(),
+  applicationId: faker.string.uuid(),
+  createdByUserId: faker.string.uuid(),
+  schemaVersion: faker.string.uuid(),
+  outdatedSchema: false,
+  createdAt: DateFormats.dateObjToIsoDateTime(faker.date.recent()),
+  submittedAt: DateFormats.dateObjToIsoDateTime(faker.date.recent()),
+  data: {},
+  document: {},
+}))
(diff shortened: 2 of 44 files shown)


pezholio:
As mentioned, the PlacementRequest object will now be `PlacementApplication`. You can probably generate the types ahead of time by editing the `script/generate_types` script and pointing to the api.yml in this branch https://github.com/ministryofjustice/hmpps-approved-premises-api/tree/172-add-crud-endpoints-to-placement-requests

Hopefully, we can get https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/678 merged in soon 👍

richpjames:
@pezholio good idea, I was going to suggest waiting for the API branch to be merged but would be good to get ahead

server/testutils/factories/placementApplication.ts
```diff
@@ -0,0 +1,16 @@
+import { Factory } from 'fishery'
+import { faker } from '@faker-js/faker/locale/en_GB'
+import { PlacementApplication } from '../../@types/shared'
+import { DateFormats } from '../../utils/dateUtils'
+
+export default Factory.define<PlacementApplication>(() => ({
+  id: faker.string.uuid(),
+  applicationId: faker.string.uuid(),
+  createdByUserId: faker.string.uuid(),
+  schemaVersion: '1',
```
pezholio, line 10: This will be a UUID so best to have it as that

server/utils/journeyTypeFromArtifact.ts
```diff
@@ -0,0 +1,9 @@
+import { FormArtifact, JourneyType } from '../@types/ui'
+import isAssessment from './assessments/isAssessment'
+import { isPlacementApplication } from './placementRequests/isPlacementApplication'
+
+export const journeyTypeFromArtifact = (artifact: FormArtifact): JourneyType => {
+  if (isAssessment(artifact)) return 'assessments'
+  if (isPlacementApplication(artifact)) return 'placement-applications'
```
pezholio, line 7: Wonder if it might be worth documenting how this works, it could be a bit confusing as Assessments and PlacementApplications both has applicationIDs

server/paths/match.ts
```diff
@@ -16,4 +20,12 @@ export default {
       search: placementRequestPath.path('beds/search'),
     },
   },
+  placementApplications: {
```
pezholio, line 25: I wonder whether this belongs in its own placementApplications file (and the same for the routes). We'll be able to review and accept placement applictions too, and I don't think match is the best place for it.

server/controllers/match/placementRequests/pagesController.ts
```diff
@@ -0,0 +1,64 @@
+import type { NextFunction, Request, RequestHandler, Response } from 'express'
```
pezholio, line 1: Again, not sure `match` is the right folder here, as it kinda sits seperate from the match journey

server/paths/match.ts
richpjames: yeah agreed

server/utils/journeyTypeFromArtifact.ts
richpjames: I believe that assessments have the actual application attached as opposed to the ID so the order doesn't matter but that could change at some point and the order would matter

pezholio (approved):
LGTM 👍

PR #808, opened 2023-05-31
Correct next page from oral hearing page

richpjames:
Previously we weren't showing the placement purpose page if the person selected that they don't know when the oral hearing date is.
This PR corrects this and ensures that the 'previous' function returns correctly from the placement purpose page.


pezholio (approved):
LGTM 👍

PR #816, opened 2023-06-02
Remove Standard AP option from Match

pezholio:
Checking the "Standard AP" option on Match throws a 400 error, as `standard` is not a recognised PlacementCriteria. We should remove this from the options, as all APs are "standard", unless they have a specialism.


richpjames (approved):
Might be worth adding the PR message to the commit message for context later down the line

PR #818, opened 2023-06-02
Run tests in parallel

pezholio:
This follows on from the changes we made in the API Repo to run tests in parallel in CircleCI (https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/675). It uses [parallelism](https://circleci.com/docs/parallelism-faster-jobs/) to split the Unit and Integration tests into four equal parts and run them in parallel in different containers. For the unit tests, we have to initially run without coverage, then join up the coverage reports in a seperate step and check the combined reports. 

The time saving isn't as drastic as we've seen in the API, but still seems to shave about 5 minutes off the total run time! 🙌


.circleci/config.yml
```diff
@@ -20,7 +20,135 @@ parameters:
...
+          name: Run wiremock
+          command: java -jar wiremock.jar --port 9999
+          background: true
+      - run:
+          name: Run the node app.
+          command: npm run compile-sass && npm run start-feature
+          background: true
+      - run:
+          name: Wait for node app to start
+          command: sleep 5
```
richpjames, line 139: As a future optimisation could it would be good to know more precisely when the node app has started

.circleci/config.yml
```diff
@@ -67,6 +195,19 @@ workflows:
   version: 2
   build-test-and-deploy:
     jobs:
+      - build:
+          filters:
+            tags:
+              ignore: /.*/
```
richpjames, line 201: What is this doing?

.circleci/config.yml
```diff
@@ -20,7 +20,135 @@ parameters:
...
+      name: hmpps/node
+      tag: << pipeline.parameters.node-version >>
+    steps:
+      - checkout
+      - attach_workspace:
+          at: ~/app
+      - run:
+          name: Check Coverage
+          command: |
+            npx nyc merge ./coverage/ ./coverage/.nyc_output
```
richpjames, line 109: I'm a bit unclear on what coverage checking jest is doing internally and what we're using istanbul for

package.json
```diff
@@ -22,7 +22,7 @@
     "lint:fix": "eslint . --cache --max-warnings 0 --fix && node_modules/.bin/shellcheck ./script/**/*[^utils] ./script/utils/**",
     "typecheck": "tsc && tsc -p integration_tests",
     "test": "jest --collectCoverage",
-    "test:ci": "jest --collectCoverage --maxWorkers=50%",
+    "test:ci": "jest --collectCoverage --runInBand --coverageThreshold=\"{}\"",
```
richpjames, line 25: What does this coverage threshold mean?

richpjames (commented):
Looks good! I have some questions

.circleci/config.yml
pezholio: Yeah, I'm sure we did this on another project. It should just be a case of polling a port

.circleci/config.yml
pezholio: I nabbed this from the template repo, it basically skips these checks when we push a git tag to the repo. As we don't have / use tags, this is probably unecessary noise

.circleci/config.yml
richpjames: Yeah we did/do something similar in ap-tools

.circleci/config.yml
pezholio: This merges all the coverage reports from the unit test run into one file called `.nyc_output` (which is what `nyc` looks for)

package.json
pezholio: This allows us to create the coverage report, but not check the coverage, as we're only running a small subset of tests in each container. They then get persisted to the workspace as `./coverage/coverage_${CIRCLE_NODE_INDEX}.json` and joined together with `nyc merge`

.circleci/config.yml
pezholio: Oh good call - I'll nick that

pezholio: That works!

package.json
richpjames: makes sense

.circleci/config.yml
richpjames: thanks

richpjames (approved):

PR #822, opened 2023-06-05
Fix invisible checkboxes on date of offence page

richpjames:
By setting 'govuk-visually-hidden' class on the label we were hiding the checkboxes completely. Now we only set it on the text itself via a span


pezholio (approved):
LGTM 👍

PR #825, opened 2023-06-07
Call the reallocatable tasks endpoint for workflow

pezholio:
For workflow managers, we now need to call the `tasks/reallocatable` endpoint, which returns all tasks that can be reallocated.

### DIFF
--- a/server/services/taskService.ts
+++ b/server/services/taskService.ts
@@ -5,10 +5,10 @@ import TaskClient from '../data/taskClient'
 export default class TaskService {
   constructor(private readonly taskClientFactory: RestClientBuilder<TaskClient>) {}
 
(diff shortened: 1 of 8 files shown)


server/services/taskService.ts
```diff
@@ -5,10 +5,10 @@ import TaskClient from '../data/taskClient'
 export default class TaskService {
   constructor(private readonly taskClientFactory: RestClientBuilder<TaskClient>) {}
 
-  async getAll(token: string): Promise<Array<Task>> {
+  async getAllReallocatable(token: string): Promise<Array<Task>> {
```
richpjames, line 8: I think the tests will need updating for this?

richpjames (approved):

PR #824, opened 2023-06-07
(176) Placement request form

richpjames:
# Context
Building on #807 this adds more pages to the Request a Placement form. 
We also add the submission endpoint and confirmation page. 

[Trello card](https://trello.com/c/ncPgE6LL/176-add-a-flow-to-create-and-submit-a-placement-request)

# Changes in this PR
- Initially some tidying up and corrections from #807. 
- We then add the majority of the form pages. 
- Before adding the 'Check Your Answers' page we have to adapt some of the CYA utils used for Applications/Assessments to be able to accept Placement Applications. We need to add some unique utils to build the Placement Application as the journey only has one one section and task and we want the individual pages to be shown on individual cards on the CYA screen. This is due to the PA flow not having a tasklist so we need to keep all the pages in one task. 
- We then add the Confirmation page 
- We then retroactively add the 'submission' functionality to submit the PA in the API when the user has passed the CYA screen 

## Screenshots of UI changes
check-your-answers
dates-of-placement
previous-rotl
reasons
same-ap
updates-to-application

PR #828, opened 2023-06-07
(176) Request a placement - Existing application flow

richpjames:
# Context
During the 'request a placement' journey if a user responds to the 'reason for placement' question that the reason is 'An additional placement on an existing application' then we ask about arrival dates, duration and reason and then skip a bunch of the questions from the ROTL flow added in #824 and direct them straight to the 'Update to application' page.

[Trello card](https://trello.com/c/ncPgE6LL/176-add-a-flow-to-create-and-submit-a-placement-request)

# Changes in this PR

## Screenshots of UI changes
Placement Applications -- allows me to complete the form for a ROTL request


server/form-pages/placement-application/request-a-placement/additionalPlacementDetails.ts
```diff
@@ -0,0 +1,76 @@
...
+} & ObjectWithDateParts<'arrivalDate'>
+
+@Page({
+  name: 'additional-placement-details',
+  bodyProperties: ['arrivalDate', 'arrivalDate-day', 'arrivalDate-month', 'arrivalDate-year', 'duration', 'reason'],
+})
+export default class AdditionalPlacementDetails implements TasklistPage {
+  title = 'Placement details'
+
+  question = 'Why are you requesting a placement?'
```
pezholio, line 19: Does this need to be here?

server/utils/placementRequests/checkYourAnswersUtils.ts
```diff
@@ -19,9 +18,10 @@ export const getPageTitle = (placementApplication: PlacementApplication, pageNam
   return new Page(placementApplication.data?.['request-a-placement'][pageName], placementApplication).title
 }
 
-export const pagesForReview = () => {
-  const task = PlacementRequest.sections.map(section => section.tasks).flat()[0]
-  return Object.keys(task.pages).filter(page => page !== 'check-your-answers')
+export const pagesForReview = (placementApplication: PlacementApplication) => {
```
pezholio, line 21: Bryan made a really good point the other day - if someone goes in and fills out a question, then edits their response to something that alters the branching, the previous response will still show, so we really should be walking the tree to make sure we only return relevant responses, and filter the other ones out. Not for now though, just to be aware.

pezholio (approved):
LGTM - Good plan on having integration tests for all the journeys. They're small enough for there to be not that much of a hit on build time, but important enough for us to have to have confidence!

server/form-pages/placement-application/request-a-placement/additionalPlacementDetails.ts
richpjames: Nope

server/utils/placementRequests/checkYourAnswersUtils.ts
richpjames: Yeah I saw that, great perception.  
Hopefully it won't bite us in the short term

PR #832, opened 2023-06-08
Fix Application statuses

pezholio:
At the moment, any application that is not either `inProgress` or `submitted` shows as "In Progress" and in the "In Progress" tab on the applications view. Also, only `submitted` applications show the read only view, meaning that if an application has been approved, we see the tasklist again, which is not what we want. 

This PR tightens this up, meaning only in-progress applications see the tasklist, and all possible statuses are handled on the dashboard.


richpjames (approved):
Nice!

PR #834, opened 2023-06-08
(176) Request a placement - Parole flow

richpjames:
# Context
This adds the third and final flow to the 'Request a placement' form. 
We add 2 screens, one asking about the decision to release and another requesting the user add additional relevant documents.
[Trello card](https:/trello.com/c/ncPgE6LL/176-add-a-flow-to-create-and-submit-a-placement-request)

updates-to-applicatiosn
updates-to-application

### DIFF
--- a/server/form-pages/placement-application/request-a-placement/additionalDocuments.ts
+++ b/server/form-pages/placement-application/request-a-placement/additionalDocuments.ts
@@ -1,0 +1,87 @@
+import type { ApprovedPremisesApplication, Document, PlacementApplication } from '@approved-premises/api'
+import { DataServices, TaskListErrors } from '@approved-premises/ui'
+import { Page } from '../../utils/decorators'
+
+import TasklistPage from '../../tasklistPage'
+
+type AdditionalDocumentsBody = {
+  selectedDocuments?: Array<Document>
+}
+
+type AdditionalDocumentsResponse = {
+  selectedDocuments?: Array<Document>
+  documentIds?: Array<string> | string
+  documentDescriptions?: Record<string, string>
+}
+
+@Page({ name: 'additional-documents', bodyProperties: ['selectedDocuments'] })
+export default class AdditionalDocuments implements TasklistPage {
+  title = 'Select any additional documents that are required to support your application'
+
+  documents: Array<Document> | undefined
+
+  application: ApprovedPremisesApplication
+
+  constructor(public body: AdditionalDocumentsBody, public readonly placementApplication: PlacementApplication) {}
+
+  static async initialize(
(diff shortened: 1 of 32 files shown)


server/views/placement-applications/pages/request-a-placement/decision-to-release.njk
```diff
@@ -0,0 +1,48 @@
...
+          text: page.questions.informationFromDirectionToRelease,
+          classes: "govuk-label--m"
+        },
+        hint: {
+            html: 
+            'This should include:
+            <ul class="govuk-list govuk-list--bullet govuk-hint">
+                <li>the type of AP placement</li>
+                <li>the placement location</li>
+                <li>any factors that mav affect the release date</li>
```
pezholio, line 39:
```suggestion
                <li>any factors that may affect the release date</li>
```

server/form-pages/placement-application/request-a-placement/additionalDocuments.ts
```diff
@@ -0,0 +1,87 @@
+import type { ApprovedPremisesApplication, Document, PlacementApplication } from '@approved-premises/api'
```
pezholio, line 1: I think this belongs in another commit?

server/utils/placementRequests/checkYourAnswersUtils.test.ts
```diff
@@ -20,47 +21,78 @@ describe('checkYourAnswersUtils', () => {
...
+            ],
           },
-        ])
-      })
+        },
+      ])
     })
   })
+
+  it('if the page name is "additional-documents"', () => {
```
pezholio, line 68: I feel like this stuff should be done after the additional documents page is added, as it doesn't make sense linerarly otherwise

server/form-pages/placement-application/request-a-placement/reasonForPlacement.ts
```diff
@@ -25,7 +25,17 @@ export default class ReasonForPlacement implements TasklistPage {
...
+      return 'previous-rotl-placement'
+    }
+    if (this.body.reason === 'additional_placement') {
+      return 'additional-placement-details'
+    }
+    if (this.body.reason === 'release_following_decision') {
+      return 'decision-to-release'
+    }
+
+    return ''
```
pezholio, line 38: Wonder if we should have a case statement and throw an error as the default here? The reason should be one of those three reasons, otherwise we have a problem

richpjames: Yeah good point

pezholio (approved):
LGTM 👍

PR #835, opened 2023-06-12
(202) Content changes

richpjames:
We have had some guidance on some content changes that need to be made. 
They are contained in this document: https://docs.google.com/document/d/19ZqaHCrv6xah3OycbnFe_1_Wadnz-7Js4QD8v5-Gg64/ 
This PR completes all of the suggestions apart from: 
- the 'Application timeliness' points. This page doesn't exist in Apply currently. It only exists in the prototype
- The error messages for the matching information will be hard to implement. We can work on these post go-live

[Trello card](https://trello.com/c/lJNPX09s/202-content-changes-for-apply-and-assess)


patrickjfl (approved):
LGTM 👍

PR #838, opened 2023-06-14
(228) Update week inputs to weeks and days

pezholio:
This updates all the places where we ask for and about placement length to ask for weeks and days. On all the form pages, rather than storing just the weeks and days, we gather them both up and add them together into a duration field, which we then send to the API as the "duration". The search already accepts duration in days, so we don't have to do anything API-side to make all this work correctly.

## Screenshots

Placement Applications -- allows me to complete form if the reason for placement is an additional placement on an existing application

Placement Applications -- allows me to complete form if the reason for placement is ROTL

Placement Requests -- allows me to search for available rooms (1)

### DIFF
--- a/server/form-pages/apply/move-on/placementDuration.ts
+++ b/server/form-pages/apply/move-on/placementDuration.ts
@@ -1,4 +1,4 @@
-import { addDays } from 'date-fns'
+import { addDays, formatDuration, weeksToDays } from 'date-fns'
 import type { TaskListErrors, YesOrNo } from '@approved-premises/ui'
 import { ApprovedPremisesApplication } from '@approved-premises/api'
 import { SessionDataError } from '../../../utils/errors'
@@ -11,11 +11,16 @@ import { getDefaultPlacementDurationInWeeks } from '../../../utils/applications/
 
 type PlacementDurationBody = {
   differentDuration: YesOrNo
+  durationDays?: string
+  durationWeeks?: string
   duration?: string
   reason?: string
 }
 
-@Page({ name: 'placement-duration', bodyProperties: ['differentDuration', 'duration', 'reason'] })
+@Page({
+  name: 'placement-duration',
+  bodyProperties: ['differentDuration', 'duration', 'durationDays', 'durationWeeks', 'reason'],
+})
 export default class PlacementDuration implements TasklistPage {
   title = 'Placement duration and move on'
 
@@ -29,7 +34,9 @@ export default class PlacementDuration implements TasklistPage {
     reason: 'Why does this person require a different placement duration?',
   }
 
-  constructor(public body: Partial<PlacementDurationBody>, private readonly application: ApprovedPremisesApplication) {}
+  constructor(public body: Partial<PlacementDurationBody>, private readonly application: ApprovedPremisesApplication) {
+    this.body.duration = String(this.lengthInDays())
+  }
 
   previous() {
     return 'dashboard'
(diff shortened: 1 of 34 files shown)


pezholio:
[Update week inputs to weeks and days](https://trello.com/c/CJs6Dw02/228-update-week-inputs-to-weeks-and-days)

server/form-pages/apply/move-on/placementDuration.ts
```diff
@@ -71,6 +88,18 @@ export default class PlacementDuration implements TasklistPage {
     return errors
   }
 
+  private lengthInDays(): number | null {
+    if (this.body.differentDuration === 'yes') {
+      if (this.body.durationDays && this.body.durationWeeks) {
```
richpjames, line 93: So if someone only entered weeks OR days would this not convert them? but that's not possible because of the validation

server/form-pages/apply/schema.json
```diff
@@ -1666,9 +1672,7 @@
               "properties": {
                 "arePlansInPlace": {
                   "type": "string",
-                  "enum": [
-                    "yes"
-                  ]
+                  "const": "yes"
```
richpjames, line 1675: I'm not sure that these schema changes belong in this commit ?

server/views/components/formFields/form-page-day-weeks-input/macro.njk
```diff
@@ -0,0 +1,66 @@
+{% from "../form-page-input/macro.njk" import formPageInput %}
+{% from "govuk/components/fieldset/macro.njk" import govukFieldset %}
+
+{% macro formPageDaysAndWeeksInput(fieldName, legendClasses, context) %}
```
richpjames, line 4: 👏👏👏

server/utils/assessments/dateUtils.ts
```diff
@@ -38,6 +38,17 @@ const daysUntilDue = (assessment: AssessmentSummary): number => {
   return differenceInDays(dueDate, new Date())
 }
 
+const daysToWeeksAndDays = (days: string | number): { days: number; weeks: number } => {
+  const daysAsNumber = Number(days)
+  // eslint-disable-next-line no-bitwise
+  const durationWeeks = (daysAsNumber / 7) | 0
```
richpjames, line 44: this should be `(daysAsNumber / 7) || 0`. We're using the bitwise operator here which I think is unintentional so we can remove line 43 too

server/form-pages/assess/matchingInformation/matchingInformationTask/matchingInformation.ts
```diff
@@ -165,7 +167,10 @@ export default class MatchingInformation implements TasklistPage {
   }
 
   get suggestedLengthOfStay() {
-    return placementDurationFromApplication(this.assessment.application)
+    return formatDuration(daysToWeeksAndDays(placementDurationFromApplication(this.assessment.application)), {
```
richpjames, line 192: 🤯 what a handy function. 
I think it would be good to have this as a static method on DateFormats so we don't have to duplicate the options each time

richpjames (approved):
No showstoppers  but I've left a few comments

server/form-pages/apply/move-on/placementDuration.ts
pezholio: Yeah, they should put a zero if it's an exact number of weeks or days

server/utils/assessments/dateUtils.ts
pezholio: We do need the bitwise operator for this, but I'm not sure _why_ it works! I've gone for `const durationWeeks = Math.floor(daysAsNumber / 7)` as it's more readable

server/form-pages/assess/matchingInformation/matchingInformationTask/matchingInformation.ts
pezholio: Good call. I've added a refactor in the last commit

PR #839, opened 2023-06-14
Fix coverage

pezholio:
The Coverage step in CircleCI was using the default coverage levels for `nyc`, which is 80%. This ensures we're using the same coverage values in CI and local and adds the missing test that was causing coverage to break locally.


server/form-pages/placement-application/request-a-placement/checkYourAnswers.test.ts
```diff
@@ -9,6 +11,26 @@ describe('CheckYourAnswers', () => {
   }
   const placementApplication = placementApplicationFactory.build()
 
+  describe('initialize', () => {
```
richpjames, line 14: ooops, thanks Stu!

richpjames (approved):

server/form-pages/placement-application/request-a-placement/checkYourAnswers.test.ts
pezholio: Just as much my fault! 😆

PR #842, opened 2023-06-14
(237) Add guidance when the user selects an appeal

pezholio:
This adds a big obvious warning with guidance when a placement gets cancelled through the appeals process. It’s a bit of a hack, but it is enough to see us through until we have a proper online process

# Screenshot

image


pezholio:
[Add guidance for appealing a placement](https://trello.com/c/Mkfvl2fP/237-add-guidance-for-appealing-a-placement)

richpjames (approved):

PR #843, opened 2023-06-14
Fix confirmationPageMessage

pezholio:
Since the decision was changed to a single `Accept` option, the confirmation page message was showing as though it had been rejected.

We may need to update the copy if there is no release date, but this ensures the copy is correct for now


richpjames (approved):
ahh thought I'd caught all these places the decision was used but obviously not

PR #852, opened 2023-06-15
Send correct placement duration

pezholio:
If the placement durations were not set, we were sending a literal value of `'null'`, a string, which was then getting cast as a zero, causing the placement requests to have a length of zero days, meaning the API would error. 

Javascript is fun sometimes isn't it? 😆


richpjames (commented):
Nice work tracking this down, it can't have been much fun. 
I have left a comment which I think would be worth addressing before merge but I don't think it will take long

PR #854, opened 2023-06-16
Update e2e test command

pezholio:
This was running all the e2e tests (local and dev), so was breaking.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -193,7 +193,7 @@ jobs:
       - run:
           name: E2E Check
           command: |
-            npx playwright test
+            npm run test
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -193,7 +193,7 @@ jobs:
       - run:
           name: E2E Check
           command: |
-            npx playwright test
+            npx run test
```
richpjames, line 196: I think this should be `npm run test`

richpjames (approved):

.circleci/config.yml
pezholio: Ooops!

PR #856, opened 2023-06-16
Move placement request button to page actions

pezholio:
This moves the Placement Request button to the top of the Application's read only view, as well as only showing if if the user owns that application. As it stands, we can't create a placement request for a non-approved application, but there's no easy way to check this as it stands, so this will do for now.

## Screenshot

image


integration_tests/tests/match/placementApplications.cy.ts
```diff
@@ -218,4 +229,19 @@ context('Placement Applications', () => {
       checkYourAnswersPage.clickSubmit()
     })
   })
+
+  it('does not allow me to create an application if I did not create the application', () => {
```
richpjames, line 240: suggest 'does not allow me to create a placement application'

richpjames (approved):

pezholio:
[Remove 'Create placement request' button from readonly application view under certain circumstances](https://trello.com/c/6qutZsta/238-remove-create-placement-request-button-from-readonly-application-view-under-certain-circumstances)

PR #863, opened 2023-06-19
(12) Add occupancy calendar

richpjames:
This adds an occupancy calendar to the Manage view, allowing users to see bookings at a glance. If any beds are overbooking, we show and "Overbooked" cell for the period and link through to the bed. In future, we'll show the booking details on this page, together with details of the overbooking.

## Screenshot

image

### DIFF
--- a/server/utils/addOverbookingsToSchedule.ts
+++ b/server/utils/addOverbookingsToSchedule.ts
@@ -1,0 +1,41 @@
+import { BedOccupancyEntryUi } from '@approved-premises/ui'
+import { addDays, differenceInDays, isBefore, subDays } from 'date-fns'
+
+export const addOverbookingsToSchedule = (schedule: Array<BedOccupancyEntryUi>): Array<BedOccupancyEntryUi> => {
+  const overbookings: Array<BedOccupancyEntryUi> = []
+
+  // Ensure the schedule is ordered by date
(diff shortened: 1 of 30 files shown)


server/utils/addOverbookingsToSchedule.ts
```diff
@@ -0,0 +1,41 @@
...
+  // Ensure the schedule is ordered by date
+  const orderedSchedule = schedule.sort((a, b) => a.startDate.getTime() - b.startDate.getTime())
+
+  // Loop through each schedule item and check for overbookings
+  const newSchedule = orderedSchedule.map((item, index) => {
+    const nextItem = orderedSchedule[index + 1]
+    // Is the next item a lost bed or booking?
+    if (nextItem && ['booking', 'lost_bed'].includes(nextItem.type)) {
+      // If it is, is the start date before our end date?
+      if (nextItem.startDate < item.endDate) {
```
richpjames, line 16: Is there a reason we couldn't use `isBefore` here?

richpjames (commented):
Looks great Stu, thanks!

server/utils/addOverbookingsToSchedule.ts
pezholio: Absolutely no reason! Only reason was because the code was making my eyes bleed! 😆

pezholio (approved):

pezholio (approved):

PR #870, opened 2023-06-20
Fix Ensure desirable search filters keep their state between searches

richpjames:
Previously if a filter was selected then when the form was submitted it would be unselected when the page was reloaded... despite showing the results filtered as intended.
We now feed the selected characteristics (params.requiredCharacteristics) back into the filters so they remain selected.
I am not entirely sure how the integration tests were passing before but I have corrected them now so they take both required characteristics and optional characteristics into account.§


pezholio (approved):
LGTM - I think we could probably clean this behaviour up a bit with an external function or even a class, but if this works, let's get it in for now 👍

PR #880, opened 2023-06-23
Add Tooltips to calendar

pezholio:
As we’re adding `overflow: ellipsis` to all cells, it doesn’t make sense to reduce the content of the cells. We’re also adding some tooltips so users can see the full cell content when they mouseover.

Xs8bsbTa4o


server/utils/calendarUtils.test.ts
```diff
@@ -170,32 +169,53 @@ describe('calendarUtils', () => {
     })
 
     it('if the bedOccupancyEntry.type is lost it returns the markup for a lost bed cell', () => {
-      const lostBedOccupancyEntry = bedOccupancyEntryUiFactory.build({ type: 'lost_bed' })
+      const lostBedOccupancyEntry = bedOccupancyEntryUiFactory.build({
```
richpjames, line 172: It would be safe to use the `bedOccupancyEntryLostBedUiFactory` here

server/utils/calendarUtils.test.ts
```diff
@@ -170,32 +169,53 @@ describe('calendarUtils', () => {
...
+      expect(labelForScheduleItem(lostBedOccupancyEntry, premisesId, bedId)).toMatchStringIgnoringWhitespace(
+        `<span title="${expectedText}" class="tooltip">
+          <a href="/premises/${premisesId}/beds/${bedId}/lost-beds/${lostBedOccupancyEntry.lostBedId}" class="govuk-link govuk-link--lost_bed">${expectedText}</a>
+        </span>`,
+      )
     })
 
     it('if the bedOccupancyEntry.type is booking it returns the markup for a booking cell', () => {
-      const bedOccupancyEntry = bedOccupancyEntryUiFactory.build({ type: 'booking', bookingId: '123' })
+      const bedOccupancyEntry = bedOccupancyEntryUiFactory.build({
```
richpjames, line 189: same here but with `bedOccupancyEntryBookingUiFactory`

richpjames (approved):
LGTM

server/utils/calendarUtils.test.ts
pezholio: Good call!

PR #883, opened 2023-06-23
Fix branching for tasks and orphaned questions

richpjames:
Previously if a user went through a task in one of our forms but then changed their minds and chose a different route through the form we could show answers for questions they had answered but were no longer relevant. 
Thanks to our friends in CAS3 we have a new approach where we walk the three calling `.next` on each form page to ensure we have the correct answers. 
Unfortunately compared to CAS3 there are some complications that mean we can't use [their implementation](https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/416) wholesale: 
1. We start the form with an optional question which means it won't always have a response. In order to avoid this we have to check that there is a 'body' for each form page (the question has been answered). We then have to calculate the next page by moving to the next page name in our array of pages names but only if there is another page name.
2. For Placement Applications as we only have one task we can't snip off the 'Check your answers' section like we do for Apply/Assess so we have to manually filter it out. Concievably there could be a use for `forPagesInTask` where we want the CYA page to be included so we would have to counter this some how.


server/utils/applications/getResponses.ts
```diff
@@ -0,0 +1,21 @@
...
+export const getResponses = (formArtifact: FormArtifact): ApplicationOrAssessmentResponse => {
+  const responses = {}
+
+  const formSections = getSections(formArtifact)
+
+  formSections.forEach(section => {
+    section.tasks.forEach(task => {
+      const responsesForTask: Array<PageResponse> = []
+
+      forPagesInTask(formArtifact, task, page => responsesForTask.push(page.response()))
```
pezholio, line 14: This method gets introduced in a later commit. If it's not too much of a faff rebase wise, it'd be good to get the commits in the right order 🥺

server/utils/placementRequests/mapTaskForSummaryList.ts
```diff
@@ -0,0 +1,20 @@
+import { ApprovedPremisesApplication as Application, PlacementApplication } from '../../@types/shared'
```
pezholio, line 1: This is a new function, so the commit message probably doesn't reflect what's going on here

server/utils/placementRequests/mapTaskForSummaryList.ts
```diff
@@ -0,0 +1,20 @@
...
+import { forPagesInTask } from '../applications/forPagesInTask'
+import { mapPageForSummaryList } from './checkYourAnswersUtils'
+
+export const mapTaskForSummaryList = (
+  placementApplication: PlacementApplication,
+  application: Application,
+): Array<SummaryListWithCard> => {
+  const pageSummaries: Array<SummaryListWithCard> = []
+  const placementApplicationTask = PlacementApplicationForm.sections[0].tasks[0]
+  delete placementApplicationTask.pages['check-your-answers']
```
pezholio, line 13: Do we need to delete `check-your-answers` here, as `forPagesInTask` excludes it?

pezholio (commented):
Just a couple of comments, mainly on commit history/messages.

server/utils/placementRequests/mapTaskForSummaryList.ts
richpjames: Yeah you're right, this was introduced before I realised it's better to exclude it at `forPagesInTask`

server/utils/applications/getResponses.ts
richpjames: Was having such a mare with the history yesterday hence why it's a bit messy, I can sort it

pezholio (approved):
LGTM 👍

PR #889, opened 2023-06-26
(259) Fix date issues

pezholio:
A couple of issues were noticed that were a hangover from only having weeks as placement lengths. This PR fixes them all.


richpjames (approved):
LGTM

PR #890, opened 2023-06-26
Add tab role to Prison Info tabs

brownbearsoftware:
Event listeners not being attached by `tabPanelTableScript` because selector is based on role attribute which is missing.


pezholio (commented):
I think this is causing the accessibility tests to fail, because of the incorrect use of `role`. Probably a better fix would be to update `tabPanelTableScript` to use an ID for a selector.

pezholio (approved):

PR #892, opened 2023-06-27
Setup prod / preprod for auditing

pezholio:
This adds the relevant secrets to our prod/preprod environments. The secrets already exist in the our shared namespace and are in use by CAS3, so we can use the same ones.


richpjames (approved):
I guess these values get overridden somewhere? 
Where do the real values come from... just for my knowledge

pezholio:
@richpjames They're stored in the Cloud Platform as Kubenetes secrets, which get added manually by the team, in the same way as HMPPS Auth secrets etc

PR #897, opened 2023-06-27
(268) Fix placement application bug

pezholio:
This updates the frontend to take into account the fixes we have done in https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/757 to identify tasks by their own ID, rather than the application ID


richpjames (approved):
LGTM

PR #899, opened 2023-06-27
(271) Add arrival time

richpjames:
We need to log an arrival time when people arrive into an AP

Changes will need to be made to the API before CI will pass on this PR

[Trello card](https://trello.com/c/K2lxJqid/271-record-accurate-time-of-arrival)

### Before
Arrivals -- creates an arrival (2)

### After
Arrivals -- creates an arrival (3)


pezholio (approved):
LGTM 👍

PR #901, opened 2023-06-27
Update getSections

pezholio:
The PlacementRequest form object only has one section, so .slice(0, -1) removes the only section.


richpjames (approved):
Oooops :(

PR #902, opened 2023-06-27
Allow CRN search to handle 500 errors for missing OASys and NOMS number

richpjames:
If we recieve a 500 error from the service in the controller we now check to see if it has a 'data.detail' property with the correct content and if so we add the error to the flash. We also send 'checkOasys' and 'checkNoms' from the view as this method is also called from a different view when booking someone into an AP where they might not be concerned if there is an OASys or NOMS present and they may not have time to contact support.

[Trello card](https://trello.com/c/w2aKx7CW/253-return-user-friendly-error-for-an-application-when-a-user-has-no-noms-number-or-oasys-entry)

## Screenshots of UI changes

no-noms-record
no-oasys-record

### DIFF
--- a/server/controllers/peopleController.ts
+++ b/server/controllers/peopleController.ts
@@ -2,13 +2,14 @@ import type { Request, RequestHandler, Response } from 'express'
 
 import PersonService from '../services/personService'
 import { addErrorMessageToFlash } from '../utils/validation'
+import { supportEmail } from '../utils/phaseBannerUtils'
 
 export default class PeopleController {
   constructor(private readonly personService: PersonService) {}
 
   find(): RequestHandler {
     return async (req: Request, res: Response) => {
-      const { crn, checkCaseload } = req.body
+      const { crn, checkCaseload, checkOasys, checkNoms } = req.body
 
       if (crn) {
         try {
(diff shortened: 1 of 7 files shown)


server/utils/phaseBannerUtils.ts
```diff
@@ -3,6 +3,8 @@ import applyPaths from '../paths/apply'
 import managePaths from '../paths/manage'
 import matchPaths from '../paths/match'
 
+export const supportEmail = 'APServiceSupport@digital.justice.gov.uk'
```
pezholio, line 6: Could we use this var in `feedbackLink` below?

server/controllers/peopleController.ts
```diff
@@ -19,6 +20,18 @@ export default class PeopleController {
             addErrorMessageToFlash(req, `No person with an CRN of '${crn}' was found`, 'crn')
           } else if (checkCaseload && err.status === 403) {
             addErrorMessageToFlash(req, `The CRN '${crn}' is not in your caseload`, 'crn')
+          } else if (err.status === 500 && err.data.detail === 'No OASys present for CRN' && checkOasys) {
+            addErrorMessageToFlash(
+              req,
+              `The CRN '${crn}' does not have an OASys record. Email AP Service Support (${supportEmail}) with the person's name and CRN for help starting an AP application.`,
```
pezholio, line 26: Would be good if we could make this HTML so the error message is a bit more readable, but probs not worth doing now

server/controllers/peopleController.ts
```diff
@@ -19,6 +20,18 @@ export default class PeopleController {
             addErrorMessageToFlash(req, `No person with an CRN of '${crn}' was found`, 'crn')
           } else if (checkCaseload && err.status === 403) {
             addErrorMessageToFlash(req, `The CRN '${crn}' is not in your caseload`, 'crn')
+          } else if (err.status === 500 && err.data.detail === 'No OASys present for CRN' && checkOasys) {
```
pezholio, line 23: Could we have a nested conditional here, so something like:

```
          } else if (err.status === 500) {
              if (err.data.detail === 'No OASys present for CRN' && checkOasys) {
                 ... Do Oasys stuff here
              } else if (err.data.detail === 'No nomsNumber present for CRN' && checkOasys) {
                 ... Do nomsNumber stuff here
              } else {
                 throw err
              }
          }
```

pezholio (approved):
LGTM, just a couple of non-blocking things

server/controllers/peopleController.ts
richpjames: Yeah I think this would be good for the long term but decided not to do it in the interests of time

PR #908, opened 2023-06-28
Add Cancel Lost Bed button

richpjames:
# Context

Users need to be able to cancel a lost bed. 
This PR adds a 'Cancel lost bed' button to the Lost Bed show template

# Changes in this PR
First we add the client and service methods. 
We then add a controller method.
This is followed by the view changes

### Before
befooree

### After

LostBed -- managing lost beds -- should allow me to cancel a lost bed

### DIFF
--- a/integration_tests/mockApis/placementApplication.ts
+++ b/integration_tests/mockApis/placementApplication.ts
@@ -82,4 +82,12 @@ export default {
         url: paths.placementApplications.submit({ id: applicationId }),
       })
     ).body.requests,
+
+  verifyPlacementApplicationReviewSubmit: async (applicationId: string) =>
+    (
+      await getMatchingRequests({
+        method: 'POST',
+        url: paths.placementApplications.submitDecision({ id: applicationId }),
+      })
+    ).body.requests,
 }
--- a/integration_tests/tests/match/placementApplications.cy.ts
+++ b/integration_tests/tests/match/placementApplications.cy.ts
@@ -282,6 +282,14 @@ context('Placement Applications', () => {
     decisionPage.clickSubmit()
 
     // Then I should be taken to the confirm submission page
+    cy.task('verifyPlacementApplicationReviewSubmit', placementApplication.id).then(requests => {
+      expect(requests).to.have.length(1)
+      expect(requests[0].url).to.equal(paths.placementApplications.submitDecision({ id: placementApplication.id }))
+
+      const body = JSON.parse(requests[0].body)
+
+      expect(body).to.contain.keys('decision', 'decisionSummary', 'summaryOfChanges')
+    })
 
     Page.verifyOnPage(ReviewApplicationConfirmPage)
   })
--- a/server/controllers/manage/lostBedsController.ts
+++ b/server/controllers/manage/lostBedsController.ts
@@ -115,6 +115,14 @@ export default class LostBedsController {
       req.body.endDate = endDate
 
       try {
+        if (req.body.cancel === '1') {
+          await this.lostBedService.cancelLostBed(req.user.token, id, premisesId, { notes: req.body.notes })
+
+          req.flash('success', 'Bed cancelled')
+
+          return res.redirect(paths.lostBeds.index({ premisesId }))
+        }
+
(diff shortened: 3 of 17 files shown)


server/data/lostBedClient.test.ts
```diff
@@ -134,4 +135,36 @@ describeClient('LostBedClient', provider => {
...
+          method: 'POST',
+          path: paths.premises.lostBeds.cancel({ premisesId, id: lostBed.id }),
+          body: lostBedCancellationData,
+          headers: {
+            authorization: `Bearer ${token}`,
+          },
+        },
+        willRespondWith: {
+          status: 200,
+          body: lostBed,
```
patrickjfl, line 161: should it respond with a LostBed, or a LostBedCancellation?

server/routes/manage.ts
```diff
@@ -130,6 +130,7 @@ export default function routes(controllers: Controllers, router: Router, service
       },
     ],
   })
+  get(paths.lostBeds.cancel.pattern, lostBedsController.cancel(), { auditEvent: 'CANCEL_LOST_BED' })
```
pezholio, line 133: This gives me the ick a little bit, but I can see why you've done it 😃

pezholio: An alternative might be to check for the presence of the cancel button in the update action and then call the cancellation endpoint instead. Would mean we only need one controller endpoint then too!

integration_tests/mockApis/placementApplication.ts
```diff
@@ -82,4 +82,12 @@ export default {
         url: paths.placementApplications.submit({ id: applicationId }),
       })
     ).body.requests,
+
+  verifyPlacementApplicationReviewSubmit: async (applicationId: string) =>
```
pezholio, line 86: 👍

server/controllers/manage/lostBedsController.ts
```diff
@@ -127,4 +127,21 @@ export default class LostBedsController {
...
+      const { notes = '' } = req.body
+
+      try {
+        await this.lostBedService.cancelLostBed(req.user.token, id, premisesId, { notes })
+
+        req.flash('success', 'Bed cancelled')
+
+        return res.redirect(paths.lostBeds.index({ premisesId }))
+      } catch (err) {
+        return catchValidationErrorOrPropogate(req, res, err, paths.lostBeds.show({ premisesId, bedId, id }))
```
patrickjfl, line 143: do we want to redirect to the show lost bed path, or stay on the edit screen?

pezholio (approved):
A tentative LGTM - would prefer it if the cancellation button wasn't a GET - could you maybe try using it in the update method as I suggested? If it's a pain, then don't worry about it. We may want to have a whole other cancellation journey anyway.

server/routes/manage.ts
richpjames: Yeah I did think about this it was 50/50 but then I thought it might make auditing a bit harder, but could we use the bodyParams to send the value of the cancel button?

integration_tests/tests/match/placementApplications.cy.ts
```diff
@@ -282,6 +282,14 @@ context('Placement Applications', () => {
     decisionPage.clickSubmit()
 
     // Then I should be taken to the confirm submission page
+    cy.task('verifyPlacementApplicationReviewSubmit', placementApplication.id).then(requests => {
```
patrickjfl, line 285: ah, thank you :)

patrickjfl (commented):
Looks good, just a few minor bits I wasn't sure about that I've added comments for.

Also there's a typo in the final commit message.

server/routes/manage.ts
pezholio: Ah good point - yeah using the bodyParams should work 👍

server/controllers/manage/lostBedsController.ts
richpjames: Yeah good shout, now I've made the change Stu suggested they will be directed to the index

server/data/lostBedClient.test.ts
richpjames: good spot this

richpjames: or bad error on my part 🤷

PR #909, opened 2023-06-28
Fix case note dates undefined

richpjames:
We have been sent some undefined properties where there should be dates. 
Whilst our API devs are busy (and for extra safety) I have introduced a function which we can pass an object that we expect to contain a date, a string we can use to find that date and the function we want to use to format that date if it exists. 
We can use this in a lot of places to avoid writing`object?.maybeDate ? formatDate(object?.maybeDate) : 'No date'` over and over. 

I have used it on all the dates on the case notes pages as I'm not sure we can trust this data


pezholio (approved):
:chefskiss: Nice one!

PR #911, opened 2023-06-29
Disable caseload check in dev and preprod

pezholio:
This adds a `DISABLE_CASELOAD_CHECK` env var. If it is set, then we pass `checkCaseload=false` to the person search. This allows us to bypass the caseload check, making verification of data issues much easier


richpjames (approved):

pezholio:
I think we're going to go with another approach here, so closing for now

PR #912, opened 2023-06-29
Add isEmergencyApplication and isEsapApplication first class citizens

richpjames:
We need to pass the api isEmergencyApplication and isEsapApplication properties when submitting the application

### DIFF
--- a/integration_tests/tests/apply/apply.cy.ts
+++ b/integration_tests/tests/apply/apply.cy.ts
@@ -366,6 +366,8 @@ context('Apply', () => {
         'releaseType',
         'type',
         'isInapplicable',
+        'isEsapApplication',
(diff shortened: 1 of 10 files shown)


integration_tests/tests/apply/apply.cy.ts
```diff
@@ -343,7 +343,7 @@ context('Apply', () => {
     // Then I should be told I am ineligible for an ESAP placement
   })
 
-  it('allows completion of the form', function test() {
+  it.only('allows completion of the form', function test() {
```
pezholio, line 346: Think this snuck in by mistake

pezholio (approved):
LGTM, apart from the `it.only`

PR #916, opened 2023-06-29
Allow an undefined releaseDate

richpjames:
Fixes: https://ministryofjustice.sentry.io/issues/4283102953/?query=is%3Aunresolved&referrer=issue-stream&stream_index=11


server/form-pages/apply/move-on/placementDuration.ts
```diff
@@ -106,7 +106,7 @@ export default class PlacementDuration implements TasklistPage {
       if (placementDate && placementDate.startDateSameAsReleaseDate === 'yes') {
         const releaseDate = basicInformation['release-date']
 
-        if (!releaseDate) return undefined
+        if (!releaseDate || !releaseDate?.releaseDate) return undefined
```
pezholio, line 109: Do we want to add a test for this?

richpjames: Have added one now, I think this happening is very unlikely due to the validation but just being extra careful

pezholio (approved):

PR #933, opened 2023-07-05
(264) Apply - update contingency plans questions logic

richpjames:
# Context

There was a bug in the Contingency Plan Apply questions logic which meant that Emergency cases weren't being shown the Contingency Plan screens.
After speaking to Dan I have discovered that we should be showing the Contingency Plan questions to all cases that will arrive within 28 days of the application being made.
Now if the case is emergency from the 'Additional circumstances' screen the user is taken to the 'Contingency Plan Questions' screen and then to the Trigger Plan questions

[Trello card](https://trello.com/c/hqOCJnRr/264-check-contingency-plan-questions)

### DIFF
--- a/server/utils/applications/shouldShowContingencyPlanPages.test.ts
+++ b/server/utils/applications/shouldShowContingencyPlanPages.test.ts
@@ -1,12 +1,24 @@
+import { addDays } from 'date-fns'
 import { applicationFactory } from '../../testutils/factories'
-import { shouldShowContingencyPlanPartnersPages } from './shouldShowContingencyPlanPages'
+import {
+  shouldShowContingencyPlanPartnersPages,
+  shouldShowContingencyPlanQuestionsPage,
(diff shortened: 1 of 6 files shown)


server/utils/applications/shouldShowContingencyPlanPages.test.ts
```diff
@@ -34,7 +43,7 @@ describe('shouldShowContingencyPlanPartnersPage', () => {
   it('returns true if the application has an end date for "Post Sentence Supervision (PSS)"', () => {
     mockOptionalQuestionResponse({ pssDate: '20/02/2023' })
 
-    expect(shouldShowContingencyPlanPages(application)).toEqual(true)
+    expect(shouldShowContingencyPlanPartnersPages(application)).toEqual(true)
```
pezholio, line 46: I think this should be in the previous commit

pezholio (approved):
LGTM - just one minor commit history thing

PR #935, opened 2023-07-05
(244) Add a screen for overbooked beds

pezholio:
This adds a new screen allowing users to see the detail of an overbooked bed. The original plan was to show this on the bed view page, but this would require duplicating work to calculate overbookings on the backend. Rather than doing this, we Base64 encode the overbooking entry as JSON with the original bookings included and pass this in as a URL item. We can then grab the data from the URL and display it on a seperate page.

## Screenshot

image


server/utils/bedUtils.ts
```diff
@@ -70,3 +72,25 @@ const bedLink = (bed: BedSummary, premisesId: string): string =>
...
+export const encodeOverbooking = (overbooking: BedOccupancyOverbookingEntryUi): string => {
+  const json = JSON.stringify(overbooking)
+
+  return Buffer.from(json).toString('base64')
+}
+
+export const decodeOverbooking = (string: string): BedOccupancyOverbookingEntryUi => {
+  const json = Buffer.from(string, 'base64').toString('utf-8')
+  const obj = JSON.parse(json, (name, value) => {
+    if (['startDate', 'endDate'].includes(name)) {
```
richpjames, line 93: Maybe I'm being overcautious but would it be worth putting all of this in a `try/catch` in case any of it blows up?

richpjames (approved):
Looks good, nice work Stu 🪄

server/utils/bedUtils.ts
pezholio: I think we'd want it to blow up with an invalid date exception if there was something wrong with the input.

PR #946, opened 2023-07-11
Handle generic api errors

pezholio:
At the moment, when we catch a generic error from the API, the frontend doesn't know what to do with it and we get a 500. This fixes the problem by updating the `catchValidationErrorOrPropogate` method - If there are no `invalid-params` in the problem response from the API, but there is a `detail` response, then we return the error detail as a string, and send it to the `errorSummary` without a href attribute (as it doesn’t refer to a field).

### Screenshot

image

### DIFF
--- a/server/utils/validation.ts
+++ b/server/utils/validation.ts
@@ -44,11 +44,16 @@ export const catchValidationErrorOrPropogate = (
 ): void => {
   const errors = extractValidationErrors(error)
 
-  const errorMessages = generateErrorMessages(errors)
-  const errorSummary = generateErrorSummary(errors)
+  if (typeof errors === 'string') {
+    request.flash('errorSummary', { text: errors })
+  } else {
+    const errorMessages = generateErrorMessages(errors)
+    const errorSummary = generateErrorSummary(errors)
 
-  request.flash('errors', errorMessages)
-  request.flash('errorSummary', errorSummary)
+    request.flash('errors', errorMessages)
+    request.flash('errorSummary', errorSummary)
+  }
+
   request.flash('userInput', request.body)
 
   response.redirect(redirectPath)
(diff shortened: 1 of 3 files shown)


server/utils/validation.ts
```diff
@@ -92,11 +97,14 @@ export const errorMessage = (field: string, text: string): ErrorMessage => {
   }
 }
 
-const extractValidationErrors = (error: SanitisedError | Error) => {
+const extractValidationErrors = (error: SanitisedError | Error): Record<string, string> | string => {
   if ('data' in error) {
-    if (error.data['invalid-params']) {
+    if (error.data['invalid-params']?.length) {
```
richpjames, line 102: If we're checking the length here in order to see if it is an array I think it would be clearer to use `Array.isArray` which would also ensure it's not a string.

server/utils/validation.ts
```diff
@@ -92,11 +97,14 @@ export const errorMessage = (field: string, text: string): ErrorMessage => {
...
       return generateErrors(error.data['invalid-params'])
     }
+    if (typeof error.data === 'object' && error.data !== null && 'detail' in error.data) {
```
richpjames, line 105: I *think* just the `'detail' in error.data` would be sufficient here but I'd rather be safe than sorry

richpjames (approved):
I've left a few comments. 
I think it's unrelated to this PR but I noticed in the screenshot the copy on the error isn't in red as it normally is in the Error summary component: https://design-system.service.gov.uk/components/error-summary/ (I assume because it's not an `<a>`). 
Also I wonder if we could do some future work to improve the error copy itself but this PR is a big improvment to what currently exists

server/utils/validation.ts
pezholio: I was getting type errors there, so had to add all this extra gumph

PR #949, opened 2023-07-13
Tighten up application journey

pezholio:
This copies a bunch of work from CAS3 here https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/452 to: 

- invalidate the Check Your Answers screen
- fix the missing tasklist when a user does not check the confimation box
- hide the submit button until the form is complete

I've adapted some of it to work with both Apply and Assess, which hopefully makes it more portable too!

### DIFF
--- a/integration_tests/pages/assess/taskListPage.ts
+++ b/integration_tests/pages/assess/taskListPage.ts
@@ -1,7 +1,7 @@
 import { ApprovedPremisesAssessment } from '../../../server/@types/shared/models/ApprovedPremisesAssessment'
 import TaskList from '../taskListPage'
 
-export default class TaskListPage extends Page {
+export default class TaskListPage extends TaskList {
   constructor() {
     super('Assess an Approved Premises (AP) application')
   }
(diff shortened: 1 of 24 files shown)


integration_tests/pages/assess/taskListPage.ts
```diff
@@ -18,4 +18,15 @@ export default class TaskListPage extends Page {
   shouldNotShowSection = (section: string): void => {
     cy.get('.app-task-list').should('not.contain', section)
   }
+
```
richpjames, line 21: Would it be worth combining these two pages taskListPage.ts into one? they are basically the same other than 'Apply/Assess' in the constructor and the `visit` method

integration_tests/tests/assess/assess.cy.ts
```diff
@@ -50,7 +50,7 @@ context('Assess', () => {
       })
     })
 
-    it('allows me to assess an application', function test() {
+    it.only('allows me to assess an application', function test() {
```
richpjames, line 53: Oops

server/controllers/apply/applicationsController.ts
```diff
@@ -106,23 +106,22 @@ export default class ApplicationsController {
...
-        const errorMessage = 'You must confirm the information provided is complete, accurate and up to date.'
-        const errorObject = { text: errorMessage }
-
-        return res.render('applications/tasklist', {
-          application,
-          errorSummary: [
-            {
-              text: errorMessage,
-              href: '#confirmation',
+        req.flash('errorSummary', {
```
richpjames, line 109: could we use `addErrorMessageToFlash` here instead of calling flash directly?

richpjames (commented):
Have left a couple comments about opportunities to DRY things up

richpjames (approved):

PR #952, opened 2023-07-13
255 - Only allow withdrawing an application if the it has been submitted

richpjames:
Previously we showed the 'Withdraw' button for all applications but we would like to narrow this down to only applications that haven't been submitted. If the application has been submitted we will show a message stating 'Cannot withdraw' in the 'Withdraw' column

### DIFF
--- a/server/utils/applications/utils.ts
+++ b/server/utils/applications/utils.ts
@@ -39,7 +39,7 @@ const dashboardTableRows = (applications: Array<ApplicationSummary>): Array<Tabl
         application.arrivalDate ? DateFormats.isoDateToUIDate(application.arrivalDate, { format: 'short' }) : 'N/A',
       ),
       htmlValue(getStatus(application)),
-      createWithdrawAnchorElement(application.id),
+      createWithdrawElement(application.id, application),
     ]
   })
 }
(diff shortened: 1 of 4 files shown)


pezholio (approved):

server/utils/applications/utils.ts
```diff
@@ -74,8 +74,11 @@ const createNameAnchorElement = (name: string, applicationId: string) => {
   )
 }
 
-export const createWithdrawAnchorElement = (applicationId: string) => {
-  return htmlValue(`<a href=${paths.applications.withdraw.confirm({ id: applicationId })}>Withdraw</a>`)
+export const createWithdrawElement = (applicationId: string, application: ApplicationSummary) => {
+  if (application?.submittedAt)
+    return htmlValue(`<a href=${paths.applications.withdraw.confirm({ id: applicationId })}>Withdraw</a>`)
+
+  return textValue('Cannot withdraw')
```
pezholio, line 81: I feel like returning nothing if we can't withdraw, with the table heading as "Actions" might be better here

richpjames: Makes sense to me

PR #956, opened 2023-07-14
Emergency Applications should be 28 days or less

pezholio:
We've since discovered, that, instead of 7 days, an application should be treated as emergency if it is 28 days or less away. This has resulted in confusion with people getting assigned applications they assume to be emergency applications, even though they are not supposed to be getting sent emergency applications.  This is a small change that changes the `noticeTypeFromApplication` rule to 28 days.


richpjames (commented):
We can update `shouldShowContingencyPlanQuestionsPage` function to use the `noticeTypeFromApplication` as emergency now

PR #958, opened 2023-07-17
(317) Apply -  Select preferred APs

richpjames:
# Context
PPs need to be able to select a list of their top 5 APs. 
This screen allows for that. Initially they can use 5 select boxes but in future this could be augmented with autocompletes and only adding further inputs if the user requests them via javascript. Consider it MVP.
[Trello card](https://trello.com/c/PkFQW3p2/317-select-preferred-ap-screen)

# Changes in this PR
- There is some arranging and refactoring before we add the main changes:
- Add the Page Class for the Preferred APs screen. This is slightly complicated by the need to lookup APs from the AP
- Add the view for the Preferred APs screen.
## Screenshots of UI changes
Apply -- allows completion of application emergency flow (1)
Apply -- allows completion of application emergency flow

### DIFF
--- a/server/form-pages/apply/risk-and-need-factors/location-factors/preferredAps.ts
+++ b/server/form-pages/apply/risk-and-need-factors/location-factors/preferredAps.ts
@@ -1,0 +1,96 @@
+import type { DataServices, SelectOption, TaskListErrors } from '@approved-premises/ui'
+import { ApprovedPremisesApplication as Application, ApprovedPremises } from '../../../../@types/shared'
+
+import { Page } from '../../../utils/decorators'
+
+import TasklistPage from '../../../tasklistPage'
+import { numberToOrdinal } from '../../../../utils/utils'
+
+export type PreferredApsBody = {
+  preferredAp1: string
+  preferredAp2: string
+  preferredAp3: string
+  preferredAp4: string
+  preferredAp5: string
+  selectedAps: Array<SelectOption>
+}
+
+const preferredAps = new Array(5).fill('').map((_, i) => `preferredAp${i + 1}`)
+
+@Page({
+  name: 'preferred-aps',
+  bodyProperties: [...preferredAps, 'selectedAps'],
+})
+export default class PreferredAps implements TasklistPage {
+  title = 'Select a preferred AP'
+
+  allPremises: Array<ApprovedPremises | { name: string; id: string }> = []
+
+  preferredApOptions = preferredAps
+
+  preferredApLabels = this.preferredApOptions.map((_, i) => `${numberToOrdinal(i)} choice AP`)
+
+  selectedAps: Array<SelectOption>
+
+  body: PreferredApsBody
+
+  constructor(private _: Partial<PreferredApsBody>) {}
+
+  static async initialize(
+    body: Record<string, unknown>,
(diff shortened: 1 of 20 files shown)


server/form-pages/apply/risk-and-need-factors/location-factors/preferredAps.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  next() {
+    return ''
+  }
+
+  errors() {
+    const errors: TaskListErrors<this> = {}
+
+    if (!this.body.preferredAp1 || this.body.preferredAp1 === 'no-preference') {
+      errors.preferredAp1 = 'You must select one preferred Approved Premises'
```
pezholio, line 82: Do we need to select at least one? PPs may not have a preference at all

server/form-pages/apply/risk-and-need-factors/location-factors/preferredAps.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  static async initialize(
+    body: Record<string, unknown>,
+    _: Application,
+    token: string,
+    { premisesService }: DataServices,
+  ) {
+    const premisesSelectList = [
+      { text: 'No preference', value: 'no-preference' },
+      ...(await premisesService.getPremisesSelectList(token)),
```
pezholio, line 47: I find using `getPremisesSelectList` a bit confusing here, as we're mixing presentation logic in with other logic. I'd feel happier if we just added a `getAll` method to the PremisesService and then used that. Then we could use `convertObjectsToSelectOptions` on the list of premises in the view layer.

pezholio (commented):
Just a coupla things. I think there's a tension between our "old way" of doing things (with view logic in the services) and the current way of doing things.

server/form-pages/apply/risk-and-need-factors/location-factors/preferredAps.ts
richpjames: yeah fair enough ill make that change

server/form-pages/apply/risk-and-need-factors/location-factors/preferredAps.ts
```diff
@@ -0,0 +1,97 @@
...
+  constructor(private _: Partial<PreferredApsBody>) {}
+
+  static async initialize(
+    body: Record<string, unknown>,
+    _: Application,
+    token: string,
+    { premisesService }: DataServices,
+  ) {
+    const allPremises = await premisesService.getAll(token)
+    allPremises.unshift({ name: 'No preference', id: 'no-preference' })
```
pezholio, line 46: Do we need this line now?

server/form-pages/apply/risk-and-need-factors/location-factors/preferredAps.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    const selectedAps: Array<SelectOption> = []
+
+    preferredAps.forEach(id => {
+      const premisesSelectOption = allPremises.find(premises => {
```
pezholio, line 51: We can probs rename this to `selectedAp` or something

pezholio (approved):

PR #960, opened 2023-07-18
Fix: Date of Offence checkboxes

richpjames:
A misordering of the `<input>`/`<label>` elements were causing the checks not to show.

### DIFF
--- a/server/form-pages/apply/risk-and-need-factors/risk-management-features/dateOfOffence.ts
+++ b/server/form-pages/apply/risk-and-need-factors/risk-management-features/dateOfOffence.ts
@@ -82,8 +82,8 @@ export default class DateOfOffence implements TasklistPage {
   renderTableRow(offence: string) {
     return [
       this.textValue(sentenceCase(offence)),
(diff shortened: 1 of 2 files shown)


server/form-pages/apply/risk-and-need-factors/risk-management-features/dateOfOffence.ts
```diff
@@ -82,37 +82,40 @@ export default class DateOfOffence implements TasklistPage {
...
+      DateOfOffence.checkbox(offence, 'current', this.isSelected(offence, 'current')),
+      DateOfOffence.checkbox(offence, 'previous', this.isSelected(offence, 'previous')),
     ]
   }
 
   renderTableBody() {
     return Object.keys(offences).map(offence => this.renderTableRow(offence))
   }
 
+  static checkbox(offence: string, date: 'current' | 'previous', checked: boolean) {
```
pezholio, line 94: Would be good if we had a small test for this as we're losing coverage a bit now

pezholio (approved):
LGTM apart from a missing test (which I guess you meant to put in anyway 😉 )

PR #964, opened 2023-07-18
Fix/Correct Dates of Placement duration

richpjames:
The weeks and days were the wrong way around.
The reason we missed this was because in the unit test the integer for durationDays and durationWeeks were both '1' so it looks correct for both values.


pezholio:
Ooof! Well spotted!

pezholio (approved):

PR #966, opened 2023-07-19
(325) Show a placement request in the CRU view

pezholio:
This adds an endpoint to show the details of a Placement Request for CRU managers. If a booking already exists, they can also extend or cancel it. They can't currently create a booking, but this is next in the pipeline.

## Screenshot

image

image


pezholio:
[Show a Placement Request](https://trello.com/c/Acx8MQGC/325-show-a-placement-request)

server/utils/allAPTypes.ts
```diff
@@ -0,0 +1,8 @@
+import { ApType } from '../@types/shared'
+
+export const allApTypes = {
```
richpjames, line 3: best to type rather than cast: `export const allApTypes: Record<ApType, string> = `

server/utils/placementRequests/adminSummary.ts
```diff
@@ -0,0 +1,78 @@
...
+        value: {
+          text: placementRequest.person.crn,
+        },
+      },
+      {
+        key: {
+          text: 'Tier',
+        },
+        value: {
+          text: placementRequest.risks.tier.value.level,
```
richpjames, line 25: We should handle the cases where `tier.value` is undefined (strict mode would've caught this)

server/utils/placementRequests/adminIdentityBar.ts
```diff
@@ -23,18 +28,16 @@ export const adminActions = (placementRequest: PlacementRequestDetail): Array<Id
...
-      items,
+      href: '#',
+      text: 'Create placement',
     },
   ]
 }
+
+export const title = (placementRequest: PlacementRequestDetail) => `
+<span class="govuk-caption-l">Placement request</span>
+<h1 class="govuk-heading-l">${placementRequest.person.name}</h1>
```
richpjames, line 42: It would be good to preempt the changes to the Person object here and add a case for where a Person doesn't have a name.

richpjames (commented):
Looks good, I have left some comments

richpjames (approved):

PR #967, opened 2023-07-19
(281) Add PIPE screen

richpjames:
# Context
We need to ask the user some more questions if the application is for PIPE. 
[Trello card](https://trello.com/c/xxAAUogk/281-add-pipe-assess-screen)

# Changes in this PR
- We add some 'more details' boxes to the RFAP page that were missed the first time
- We add the PIPE screen

RFAP and PIPE are exclusive and it seemed overkill to run through the Assess journey twice just to test both pages so there isn't an intergration test for this page. 
## Screenshots of UI changes
image


pezholio (approved):
LGTM 👍

PR #969, opened 2023-07-20
Fix: Assess CYA integration test

richpjames:
I spotted this test was passing but broken. 
As the Assess journey has been developed we haven't updated the test so it was showing an error screen after clicking submit on the Suitability Assessment page when we expected it to be showing the tasklist. 
I have updated the test to include the 'update' stub and to complete the Suitability Assessment task


pezholio (approved):
Well spotted!

PR #973, opened 2023-07-24
Fix backlinks when amending cancelling bookings from admin screen

pezholio:
To support two possible routes for cancelling and amending bookings, we set the `backLink` in the Cancellation and Date Change controllers to be the referrer. We then also add the original referrer as a hidden field, so we know where to take the user back to when successful.

## Screenshots

Placement Requests -- allows me to amend a booking

Placement Requests -- allows me to cancel a booking

### DIFF
--- a/integration_tests/pages/admin/placementApplications/showPage.ts
+++ b/integration_tests/pages/admin/placementApplications/showPage.ts
@@ -35,6 +35,16 @@ export default class ShowPage extends Page {
     cy.contains('.moj-button-menu__item', 'Create placement').click()
   }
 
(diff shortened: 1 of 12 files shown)


richpjames (approved):
I have found using the referrer as the backlink can be a bit unpredictable in Apply (and think we should remove it from there when we do more work on the pages that use it) but it seems appropriate here. 
I think the security warnings assume that the user is entering the input that goes into the `req.body` (how could it know it's coming from a hidden field?) and I guess they could edit it themselves but I struggle to see how this could cause a vuln

PR #992, opened 2023-07-27
Update placement requests page to include additional context

adamhughes86:
Provide more context to the placement applications screen to make it clear to users what they're looking at

### DIFF
--- a/server/views/match/placementRequests/index.njk
+++ b/server/views/match/placementRequests/index.njk
@@ -16,6 +16,10 @@
             <h1 class="govuk-heading-l">Placement Applications</h1>
             {% include "../../_messages.njk" %}
 
+            <p class="govuk-body">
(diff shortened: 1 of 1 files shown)


server/views/match/placementRequests/index.njk
```diff
@@ -16,6 +16,10 @@
             <h1 class="govuk-heading-l">Placement Applications</h1>
             {% include "../../_messages.njk" %}
 
+            <p class="govuk-body">
+                All additional placements applications that have been assessed as suitable, require matching to an AP, and that have been allocated to you are listed below
```
richpjames, line 20: Not sure we need the second 'that': 'All additional placements applications that have been assessed as suitable, require matching to an AP, and have been allocated to you are listed below' ready better to me

richpjames (approved):

PR #999, opened 2023-08-01
Set a CRU_DASHBOARD_DISABLED flag in prod

pezholio:
We accidentally deployed the CRU Dashboard changes. Rather than reverting the changes, we’ll set a feature flag in prod, so we keep the original functionality in prod, whilst still allowing us to test in the other environments.


server/controllers/match/placementRequestsController.test.ts
```diff
@@ -42,7 +44,8 @@ describe('PlacementRequestsController', () => {
   })
 
   describe('index', () => {
-    it('should render the placement requests template', async () => {
+    it('should render the placement requests template when cruDashboardDisabled is false', async () => {
+      config.flags.cruDashboardDisabled = false
```
richpjames, line 48: I think it would be worthwhile to reset the state of the `cruDashboardDisabled` boolean in an afterEach as it will hang around after the tests have run which could cause headaches

server/controllers/match/placementRequestsController.ts
```diff
@@ -16,10 +17,17 @@ export default class PlacementRequestsController {
     return async (req: Request, res: Response) => {
       const tasks = await this.taskService.getMatchTasks(req.user.token)
 
-      res.render('match/placementRequests/index', {
-        pageHeading: 'My Cases',
-        tasks,
-      })
+      if (!config.flags.cruDashboardDisabled) {
```
richpjames, line 20: I think avoiding the double negative here would make it a bit clearer: 
```js
if (config.flags.cruDashboardDisabled) { 
          res.render('match/placementRequests/cases/index', {
          pageHeading: 'My Cases',
          tasks,
        })
      } else {
      ... 
      }

File: server/utils/userUtils.ts [hunk 3]
```diff
@@ -40,6 +41,13 @@ export const sections = {
   },
   match: {
     id: 'match',
+    title: 'Match people to Approved Premises placements',
+    description: "Allocate cases to suitable Approved Premises based on a person's needs.",
```
richpjames, line 45: I don't think we need the `.` here

server/utils/userUtils.ts
```diff
@@ -40,6 +41,13 @@ export const sections = {
...
+    shortTitle: 'Match',
+    href: matchPaths.placementRequests.index({}),
```
richpjames, line 47: Just checking we want to have the same `href` as `cruDashboard`?

richpjames (approved):
I've left a few comments but will approve so you can get it in ASAP. 
I was thinking of asking whether an E2E would be useful but given a regression on this behaviour would be very unlikely I don't think it would be.

PR #1010, opened 2023-08-03
(381) Add pagination and sorting to CRU dashboard

pezholio:
This adds server-side pagination and sorting to the CRU dashboard, using the functionality added in https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/859

## Screenshot

Placement Requests -- supports sorting

### DIFF
--- a/server/data/placementRequestClient.ts
+++ b/server/data/placementRequestClient.ts
@@ -8,6 +8,7 @@ import {
   PlacementRequest,
   PlacementRequestDetail,
   PlacementRequestSortField,
+  SortDirection,
 } from '@approved-premises/api'
 import config, { ApiConfig } from '../config'
 import RestClient from './restClient'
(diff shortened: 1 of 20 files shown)


server/utils/pagination.ts
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * Produces parameters for GOV.UK Pagination component macro
+ * NB: `page` starts at 1
+ *
+ * Accessibility notes:
+ * - set `landmarkLabel` on the returned object otherwise the navigation box is announced as "results"
+ * - set `previous.attributes.aria-label` and `next.attributes.aria-label` on the returned object if "Previous" and "Next" are not clear enough
+ */
+export function pagination(page: number, pageCount: number, hrefPrefix: string): Pagination {
```
Harriethw, line 32: really minor, but wonder if `page` could be changed to `currentPage` in this function? found it quite hard to understand what's going on here 😅

server/data/placementRequestClient.ts
```diff
@@ -24,11 +29,25 @@ export default class PlacementRequestClient {
     >
   }
 
-  async dashboard(isParole: boolean): Promise<Array<PlacementRequest>> {
-    return (await this.restClient.get({
+  async dashboard(
+    isParole: boolean,
+    page: number = 1,
```
Harriethw, line 34: v minor but ESLint is complaining that there's no need to define a type here as it's inferred
```suggestion
    page = 1,
```

Harriethw (approved):
LGTM! interesting to see how pagination can be implemented on frontend, thanks

PR #1021, opened 2023-08-08
(389) Make manage read only for non CRU admins

pezholio:
This removes some of the functionality for non CRU admins to write data back to manage. As well as removing the links, I've also added a layer of middleware to check the user's roles for certain endpoints for added safety.


edavey (approved):
this looks clear to me -- though I haven't the context to pick up possible ramifications...

PR #1038, opened 2023-08-15
(382) Use Premises summary endpoint

richpjames:
# Context
The new `premises/summary` endpoint has performance benefits over the `premises` endpoint. 
We should use it when we're listing all premises.

[Trello](https://trello.com/c/E9mQB0ZY/382-removing-the-static-number-of-beds-value-from-dashboards)
# Changes in this PR
- First we address some linting errors in the E2E tests
- We update the endpoint for premises to use the new `summary` endpoint
- We update the mockApis for the integration tests


pezholio (approved):
LGTM - ta!

PR #1040, opened 2023-08-15
(390) Allow arrived bookings to reduce length of stay

richpjames:
# Context

As well as extending a booking, users need to be able to reduce the time that a PoP will stay in an AP.
Therefore we need to change the copy on and around the 'extend a booking' screen.

# Changes in this PR
1. We delete an unused method on the Premises service - unrelated to these changes.
2. We update the copy and filenames around the Booking Extension change
3. We update the service name from 'extendBooking' to 'changeDepartureDate'

### Before
bfore 1
bfore 2

### After
booking-extension-create-page-after
booking-extension-confirmation-page-after

### DIFF
--- a/integration_tests/tests/manage/changeDepartureDate.cy.ts
+++ b/integration_tests/tests/manage/changeDepartureDate.cy.ts
@@ -1,9 +1,9 @@
 import { bookingFactory, premisesFactory } from '../../../server/testutils/factories'
 
-import { BookingExtensionConfirmationPage, BookingExtensionCreatePage } from '../../pages/manage'
+import { DepartureDateChangeConfirmationPage, DepartureDateChangePage } from '../../pages/manage'
 import { signIn } from '../signIn'
 
-context('BookingExtension', () => {
+context('Departure date', () => {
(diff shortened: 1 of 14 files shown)


integration_tests/tests/manage/changeDepartureDate.cy.ts
```diff
@@ -23,14 +23,15 @@ context('BookingExtension', () => {
...
+    const page = DepartureDateChangePage.visit(premises.id, booking.id)
+    cy.screenshot('booking-extension-create-page-after')
     // And I fill in the extension form
     page.completeForm(newDepartureDate)
     page.clickSubmit()
 
     // Then I should be redirected to the confirmation page
-    const bookingConfirmationPage = new BookingExtensionConfirmationPage()
+    const bookingConfirmationPage = new DepartureDateChangeConfirmationPage()
+    cy.screenshot('booking-extension-confirmation-page-after')
```
pezholio, line 34: This too

pezholio (approved):
LGTM apart from some erroneous `cy.screenshot` calls

PR #1043, opened 2023-08-16
(350) Cancellation reasons to be updated

pezholio:
Now the "Booking successfully appealed" reason has been retired (https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/890), we don't need to show a warning. We also update the service to ensure only active reasons are returned.


pezholio:
[Cancellation reasons to be updated](https://trello.com/c/Xj5yxx99/350-cancellation-reasons-to-be-updated)

richpjames (approved):

PR #1044, opened 2023-08-16
(296) Github action for updating the bank holidays json

richpjames:
[Trello card](https://trello.com/c/lRTWjnm3/390-create-date-change-option-should-appear-for-arrived-bookings)

We want to use 'working days' in the calendar (and potentially in future elsewhere).
In order for us to do this we need to be able to include bank holidays. As bank holidays are liable to change at short notice (EG last year) we need to make sure that these are kept up to date. 

This PR adds a Github action that calls the [gov.uk bank holidays API](https://www.api.gov.uk/gds/bank-holidays/#bank-holidays) and updates a file containing the bank holidays in this repo every night at midnight, if they have changed.

### DIFF
--- a/.github/workflows/fetch-bank-holidays.yml
+++ b/.github/workflows/fetch-bank-holidays.yml
@@ -1,0 +1,41 @@
+name: Fetch bank holidays JSON from the gov.uk API
+
+on:
+  schedule:
+    - cron: '55 8 * * *'
+
+jobs:
+  call_api:
+    name: Call API
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Call API
+        id: call_api
+        run: |
+          curl -o bank-holidays.json https://www.gov.uk/bank-holidays.json
+          mv bank-holidays.json ./server/data/bankHolidays/bank-holidays.json
+  create_pr:
+    name: Create Pull Request
+    needs: call_api
+    runs-on: ubuntu-latest
+    steps:
+      - name: Generate Token
+        uses: tibdex/github-app-token@v1
+        id: generate-token
+        with:
+          app_id: ${{ secrets.APP_ID }}
+          private_key: ${{ secrets.APP_PRIVATE_KEY }}
+      - name: Check out code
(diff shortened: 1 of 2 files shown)


.github/workflows/fetch-bank-holidays.yml
```diff
@@ -0,0 +1,50 @@
...
+        id: call_api
+        run: |
+          curl -o bank-holidays.json https://www.gov.uk/bank-holidays.json
+          if diff ./server/data/bankHolidays/bank-holidays.json bank-holidays.json; then
+            echo "Files are the same"
+          else
+            echo "Files are different"
+            mv bank-holidays.json ./server/data/bankHolidays/bank-holidays.json
+            echo "CHANGES=true" >> $GITHUB_OUTPUT
+          fi
```
pezholio, line 26: I think you can get away with not doing this actually - if you look at the [Generate Types](https://github.com/ministryofjustice/hmpps-approved-premises-ui/blob/main/.github/workflows/generate-types.yml) action, and the [runs](https://github.com/ministryofjustice/hmpps-approved-premises-ui/actions/workflows/generate-types.yml), it runs every time it gets triggered, but only opens a PR if there are any changes

richpjames: that would make things a lot simpler. 
I read something in the docs that suggested Actions will create an empty PR but I think the `peter-evans/create-pull-request@v5` must be a bit smarter

.github/workflows/fetch-bank-holidays.yml
```diff
@@ -0,0 +1,41 @@
+name: Fetch bank holidays JSON from the gov.uk API
+
+on:
+  schedule:
+    - cron: '0 0 * * *'
```
pezholio, line 5: Is it worth setting this (at least temporarily) to `'0 8 * * *'` so it'll next run at 9am (our time)?

pezholio (approved):
LGTM - let's give it a whirl!

.github/workflows/fetch-bank-holidays.yml
richpjames: good idea

PR #1047, opened 2023-08-17
(296) Github action for updating the bank holidays json

richpjames:
The previous job failed when creating a branch because the state is lost between jobs. 
Here we consolidate the two jobs into one to retain state


pezholio (approved):
LGTM

PR #1076, opened 2023-09-01
(337) Search, list and edit users

richpjames:
# Context

~~**NB: the [API spec for the branch this PR builds off is currently incorrect](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/902#discussion_r1312896559) causing the userClient.test.ts to fail. Once the error is corrected CI should pass on this branch as all the other tests are passing locally**~~

Users need to be able to self-serve permissions for other users. 
This PR adds the first piece of functionality to allow them to do this. 
Firstly we add a screen to list all users. This will likely need pagination when it is available on the API endpoint. 
Next we add the ability to search available users. 
After this we add the screen to allow users to edit the roles and qualifications for a given user. 

[Trello](https://trello.com/c/WTrXhGUP/337-user-management)

# Changes in this PR
- Firstly we start with some commits tidying and arranging things in order for the rest of the changes to proceed smoothly
- We add a user management controller with a a method for listing users
- We build out the view for listing users
- We add search functionality to the list page
- We add the ability to view and edit a specific users roles and qualifications 

## Screenshots of UI changes
updated (1)
show (2)
list (2)
dashboard (1)

### DIFF
--- a/server/utils/users/homePageDashboard.ts
+++ b/server/utils/users/homePageDashboard.ts
@@ -66,6 +66,14 @@ export const sections = {
     shortTitle: 'Reports',
     href: adminPaths.admin.reports.new({}),
   },
+  userManagement: {
+    id: 'userManagement',
+    title: 'Manage user roles',
+    description:
+      'Manage user roles and permissions. Stop automatic allocations for assessments, matches, and placement requests.',
+    shortTitle: 'Users',
+    href: adminPaths.admin.userManagement.index({}),
+  },
 }
 
 export const hasRole = (user: UserDetails, role: UserRole): boolean => {
@@ -88,6 +96,11 @@ export const sectionsForUser = (user: UserDetails): Array<ServiceSection> => {
     items.push(sections.placementRequests)
     items.push(sections.cruDashboard)
     items.push(sections.reports)
+    items.push(sections.userManagement)
+  }
+
+  if (hasRole(user, 'role_admin')) {
+    items.push(sections.userManagement)
   }
 
   return items
--- a/integration_tests/tests/dashboard.cy.ts
+++ b/integration_tests/tests/dashboard.cy.ts
@@ -49,4 +49,16 @@ context('Dashboard', () => {
 
     dashboardPage.shouldNotShowCard('assess')
   })
+
+  it('displays the apply and user management services to users with "role_admin"', () => {
+    signInWithRoles(['role_admin'])
+
+    const dashboardPage = DashboardPage.visit()
+
+    dashboardPage.shouldShowCard('apply')
+    dashboardPage.shouldShowCard('userManagement')
+
+    dashboardPage.shouldNotShowCard('assess')
+    dashboardPage.shouldNotShowCard('manage')
+  })
 })
--- a/server/utils/users/homePageDashboard.test.ts
+++ b/server/utils/users/homePageDashboard.test.ts
@@ -44,6 +44,7 @@ describe('homePageDashboard', () => {
         sections.placementRequests,
         sections.cruDashboard,
         sections.reports,
+        sections.userManagement,
       ])
     })
 
@@ -64,7 +65,14 @@ describe('homePageDashboard', () => {
         sections.placementRequests,
         sections.cruDashboard,
         sections.reports,
+        sections.userManagement,
       ])
+    })
+
+    it('should return the user management section for a user with a role admin role', () => {
+      const user = userDetailsFactory.build({ roles: ['role_admin'] })
+
+      expect(sectionsForUser(user)).toEqual([sections.apply, sections.userManagement])
     })
   })
 })


selina-dxw:
Thanks for this, just a few things: 

Can 'placement application' be changed to 'placement request' allocations. 

I've updated the content for the third section on [Figma](https://www.figma.com/file/kP7jyX3L01CQap6ErAX648/Approved-Premises-Beta?type=design&node-id=2428%3A66540&mode=design&t=BSNwFgJGEzy8MhM5-1)

Update the copy on the dashboard tile to:
Manage user roles

Manage user roles and permissions. Stop automatic allocations for assessments, matches, and placement requests.

richpjames:
@selina-dxw thank you! I'll make those changes now

server/utils/users/homePageDashboard.ts
```diff
@@ -90,5 +98,9 @@ export const sectionsForUser = (user: UserDetails): Array<ServiceSection> => {
     items.push(sections.reports)
   }
 
+  if (hasRole(user, 'role_admin')) {
```
richpjames, line 101: @pezholio I'm not sure if this is correct but it is the role that made the most sense

server/views/admin/users/index.njk
```diff
@@ -0,0 +1,41 @@
...
+
+{% block content %}
+
+    <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds">
+            <h1 class="govuk-heading-l">{{ pageHeading }}</h1>
+
+            {{ govukTable({
+                attributes: {
+                    'data-module': 'moj-sortable-table'
```
pezholio, line 53: I think if we move to pagination, sorting client side may be moot, but let's leave it as is now

server/utils/users/index.test.ts
```diff
@@ -0,0 +1,51 @@
...
+        },
+      },
+      {
+        key: {
+          text: 'Phone number',
+        },
+        value: {
+          text: user?.telephoneNumber ?? 'No phone number available',
+        },
+      },
```
pezholio: I feel like we need to test what happens without a user email / phone number in a seperate test, rather than with the logic in the tests here (sure this is just a copy paste error though)

server/utils/users/roleCheckboxes/index.ts
```diff
@@ -0,0 +1,95 @@
...
+type UnusedRoles = 'applicant' | 'report_viewer'
+
+type RolesForCheckboxes = Exclude<UserRole, AllocationRole | UnusedRoles>
+
+export type RoleLabel = { label: string; hint?: string }
+
+export type RoleLabelDictionary = { [K in RolesForCheckboxes]: RoleLabel }
+
+export type AllocationRoleLabelDictionary = { [K in AllocationRolesForCheckboxes]: RoleLabel }
+type AllocationRolesForCheckboxes = Exclude<UserRole, RolesForCheckboxes | UnusedRoles>
```
pezholio, line 30: Some top notch type magic here! 🪄

server/controllers/admin/userManagementController.ts
```diff
@@ -11,6 +13,15 @@ export default class UserController {
       res.render('admin/users/index', { pageHeading: 'User management dashboard', users })
     }
   }
+
+  show(): TypedRequestHandler<Request, Response> {
```
pezholio, line 19: Pretty minor, but this feels like an edit action rather than a show action to me

server/controllers/admin/userManagementController.ts
```diff
@@ -22,6 +24,21 @@ export default class UserController {
...
+      const user = await this.userService.getUserById(req.user.token, req.params.id)
+
+      await this.userService.updateUser(req.user.token, {
+        ...user,
+        roles: [...flattenCheckboxInput(req.body.roles), ...flattenCheckboxInput(req.body.allocationPreferences)],
+        qualifications: flattenCheckboxInput(req.body.qualifications),
+      })
+
+      req.flash('success', 'User updated')
+      res.redirect(`${paths.admin.userManagement.show({ id: user.id })}?success=true`)
```
pezholio, line 38: What does the `success=true` arg do here?

integration_tests/tests/admin/userManagement.cy.ts
```diff
@@ -13,27 +14,70 @@ context('User management', () => {
...
-    page.clickUser(user.name)
+    listPage.clickUser(user.name)
 
     // Then I am taken to the user's permissions page
     const showPage = Page.verifyOnPage(ShowPage)
+
     showPage.checkForBackButton(paths.admin.userManagement.index({}))
     showPage.shouldShowUserDetails(user)
+
+    // Given I can see the user's roles and qualifications
```
pezholio, line 41: This `Given` seems redundant here. I'd probably go with:

```
When I update the user's roles and qualifications
```

integration_tests/pages/admin/userManagement/showPage.ts
```diff
@@ -17,4 +21,36 @@ export default class ShowPage extends Page {
   shouldShowUserDetails(user: User): void {
     this.shouldContainSummaryListItems(userSummaryListItems(user))
   }
+
+  selectRoles(roles: ReadonlyArray<UserRole>): void {
```
pezholio, line 25: I think it'd be good to deselect all the roles here, and only select the ones we specify, just to give us confidence that the only roles we're sending are the ones we specify.

server/utils/users/homePageDashboard.ts
pezholio: We're letting `workflow_manager` roles edit users too 👍

pezholio (approved):
LGTM - I'm a big fan of committing all the service, client and controller stuff in one big commit like you have. Even though it makes the commits bigger, it's actually easier to follow, and makes my eyes less likely to glaze over when reviewing small, seemingly workaday commits 👍

server/utils/users/index.test.ts
richpjames: oooops

server/controllers/admin/userManagementController.ts
richpjames: I left this in in error. 
In CAS3 they do this when the page that the user is directed to on success is the same as the one if there is a validation error for auditing purposes. I thought we would need it here but as there isn't any validation we didn't end up needing it

PR #1084, opened 2023-09-05
Flatten checkbox inputs

richpjames:
When express receives a single response from a checkbox input it 'helpfully' sends it as a string as opposed to the usual array. This means array methods can't be used and causes errors (EG https://ministryofjustice.sentry.io/issues/4395174101/events/f1d83babeea64b809e5b320b1884e2bb/?project=4503931667742720) on the Check Your Answers screen.


pezholio (approved):

richpjames:
@pezholio I have only just realised that we can prevent this issue by appending `[]` to the `name` of the input. Worth doing in future but not worth retrofitting IMO

pezholio:
@richpjames No way! Big TIL!

PR #1087, opened 2023-09-05
Pin e2e node version to 9.8.1

pezholio:
Following on from https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/1078, we also pin of the version NPM we use in the e2e tests too as NPM 10 isn't compatible with the current version of Node


richpjames (approved):

yndajas:
Interesting - were there errors in the E2E tests prompting this? Haven't seen any on our end yet

PR #1095, opened 2023-09-07
Update cypress to v13

richpjames:
We previously ignored the updated to V13 as `cypress-axe` hadn't upgraded but now it has we can update Cypress

### DIFF
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -35,7 +35,6 @@ export default defineConfig({
   reporterOptions: {
     configFile: 'reporter-config.json',
   },
-  videoUploadOnPasses: false,
   taskTimeout: 70000,
   e2e: {
     setupNodeEvents(on) {
--- a/package-lock.json
+++ b/package-lock.json
@@ -85,7 +85,7 @@
         "axe-core": "^4.7.0",
         "concurrently": "^8.0.0",
         "cookie-session": "^2.0.0",
-        "cypress": "^12.10.0",
+        "cypress": "^13.1.0",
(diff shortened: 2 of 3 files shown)


yndajas (commented):
Not sure if you're still working on this, but thought I'd add a note in case it helps - what I shared with you before isn't quite enough for Approved Premises

To retain the same behaviour post-upgrade, I think you might need to:
- (as mentioned before) make these changes: https://github.com/ministryofjustice/hmpps-accredited-programmes-ui/pull/160/commits/619e42d1bbe0f15ea368c685b188b31eb1fbfb80
- add this, which does what `videoUploadOnPasses: false` was meant to do: https://github.com/ministryofjustice/hmpps-accredited-programmes-ui/blob/main/cypress.config.ts#L52-L60. Adapted from [the Cypress docs](https://docs.cypress.io/guides/guides/screenshots-and-videos#Delete-videos-for-specs-without-failing-or-retried-tests)

richpjames:
Thanks @yndajas 
I'm not sure how @pezholio feels but I never really view the videos (screenshot and logs tend to be sufficient) so the videos feel a bit redundant

pezholio:
Videos can be useful in e2e, but I never use them locally.

yndajas:
> Thanks @yndajas I'm not sure how @pezholio feels but I never really view the videos (screenshot and logs tend to be sufficient) so the videos feel a bit redundant

In which case, probably fine just to remove the `videoUploadOnPasses` config option and leave it!

yndajas (approved):

PR #1098, opened 2023-09-07
Remove names from page titles

richpjames:
We need to remove people's names from the page titles and questions in the Apply stage due to data sharing agreements and privacy issues. This will become more important when we add analytics. 
We also need to remove the names from questions that are saved in the 'Translated response' . 

This allows us to remove the dependency of many Apply pages on the application which means we can rip out a bunch of code 🥳


pezholio (approved):
LGTM - thanks for putting the boring work in for this!

PR #1102, opened 2023-09-11
(337) Add and delete a user

richpjames:
# Context

As well as search for, listing and updating the permissions of users users also need to be able to add new users and remove access for users that are no longer needed.

[Trello ticket](https://trello.com/c/WTrXhGUP/337-user-management)
Previous related PRs: #1088 #1076 

# Changes in this PR
- Some rearranging of files to keep the subsequent commits easy to follow
- Add search functionality
- Add delete functionality

### Before
before 1
before 2

### After
index after page
user permissions after (1)
delete user page (1)
search page
error


pezholio (approved):
LGTM 👍

selina-dxw:
Looks good, we've discussed adding an 'Allocations' column. 

On the 'Confirm user's access to AP service should be removed' can the 'Remove access' red button be changed to a 'Confirm' green button?

And the hint text for 'Workflow manager' under 'Role' on Manage permissions should include 'view reports'. So: 
'Manage the allocation of assessments and matches to staff, and view reports'

PR #1105, opened 2023-09-12
Show due date in business days on CRU dashboard

richpjames:
# Context
As well as showing working days for the due date on the assessment dashboard (#1055) we also need to show the due date in working days on the CRU dashboard.

# Changes in this PR
- We move the `differenceInBusinessDays` function to the DateFormats class
- We change the difference in days function to use the `differenceInBusisnessDays` function instead of just difference in days


pezholio (approved):
LGTM 👍

PR #1108, opened 2023-09-12
(430) Add pagination to user response

pezholio:
This builds on https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/956 to support the new pagination arguments for users, which should speed up the response times, as well as fix the strange sorting issues we had when using the client side sort.

## Screenshot
User management -- supports sorting (1)

### DIFF
--- a/server/controllers/admin/userManagementController.test.ts
+++ b/server/controllers/admin/userManagementController.test.ts
@@ -5,8 +5,10 @@ import { UserService } from '../../services'
 
 import UserManagementController from './userManagementController'
 import { qualifications, roles } from '../../utils/users'
-import { userFactory } from '../../testutils/factories'
+import { paginatedResponseFactory, userFactory } from '../../testutils/factories'
 import paths from '../../paths/admin'
+import { PaginatedResponse } from '../../@types/ui'
+import { ApprovedPremisesUser } from '../../@types/shared'
 
 describe('UserManagementController', () => {
   const token = 'SOME_TOKEN'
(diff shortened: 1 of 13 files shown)


server/controllers/admin/userManagementController.test.ts
```diff
@@ -25,19 +27,53 @@ describe('UserManagementController', () => {
...
   describe('index', () => {
+    const hrefPrefix = `${paths.admin.userManagement.index({})}?`
+
     it('renders the index page with all the users', async () => {
       const users = userFactory.buildList(1)
 
-      userService.getUsers.mockResolvedValue(users)
+      const paginatedResponse = paginatedResponseFactory.build({
+        data: users,
+      }) as PaginatedResponse<ApprovedPremisesUser>
```
richpjames, line 37: It's a bit annoying having to cast this but I've had a go and I can't work out a way around it

server/controllers/admin/userManagementController.ts
```diff
@@ -4,15 +4,28 @@ import { qualifications, roles } from '../../utils/users'
...
 import { flattenCheckboxInput } from '../../utils/formUtils'
+import { SortDirection, UserSortField } from '../../@types/shared'
 
 export default class UserController {
   constructor(private readonly userService: UserService) {}
 
   index(): TypedRequestHandler<Request, Response> {
     return async (req: Request, res: Response) => {
-      const users = await this.userService.getUsers(req.user.token)
+      const pageNumber = req.query.page ? Number(req.query.page) : undefined
```
richpjames, line 14: I think a `req.query?.page` would be safer (same for lines 15 and 16)

richpjames (approved):
Looks good, have left one minor thing to address

PR #1113, opened 2023-09-13
(438) Ensure assessment  due date is in working days

richpjames:
# Context
Previously (#1055) we did work to ensure that the due date shown on the assessment dashboard was in business days. 
However the due date was still calculated in calendar days meaning users had less time to complete assessments. 
This PR ensures that the due date is calculated in working days so that the due date is accurate.

[Trello card](https://trello.com/c/7GhcnItr/438-countdown-timer-allowing-less-then-the-10-working-days-for-assessments)


pezholio (approved):
LGTM 👍

PR #1114, opened 2023-09-13
Use correct sort field for arrival date

pezholio:
I’ve also added a generic to ensure that the `targetField` and `currentSortField` arguments are typed correctly.

### DIFF
--- a/server/utils/sortHeader.ts
+++ b/server/utils/sortHeader.ts
@@ -3,10 +3,10 @@ import { SortDirection } from '../@types/shared'
 import { TableCell } from '../@types/ui'
 import { createQueryString } from './utils'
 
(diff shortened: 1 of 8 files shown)


server/utils/sortHeader.ts
```diff
@@ -3,10 +3,10 @@ import { SortDirection } from '../@types/shared'
 import { TableCell } from '../@types/ui'
 import { createQueryString } from './utils'
 
-export const sortHeader = (
+export const sortHeader = <T>(
```
richpjames, line 6: If we do `<T extends string>` we can avoid casting on line 33

richpjames (approved):
LGTM, one minor comment

PR #1120, opened 2023-09-15
(337) User management design & content iterations

richpjames:
# Context

Some miscellaneous design and content iterations on the User management screens.

[Trello card](https://trello.com/c/WTrXhGUP/337-user-management)

# Changes in this PR
- On the user index page we split the 'roles' column into 'roles' and 'allocations' and add 'qualifications' to the 'allocations' column (previously not displayed on the index page). This mirrors what we do on the 'Edit' page
- Update the hint text for workflow manager to indicate they can view reports
- Update the colour of the Deletion confirmation button on the edit page to be green instead of red

### Before
befre 1
befre 2
after (3)

### After
after
User management -- allows searching for users (2)
after (2)

### DIFF
--- a/integration_tests/tests/admin/userManagement.cy.ts
+++ b/integration_tests/tests/admin/userManagement.cy.ts
@@ -74,7 +74,8 @@ context('User management', () => {
       const body = JSON.parse(requests[0].body)
 
       expect(body).to.deep.equal({
-        ...updatedUser,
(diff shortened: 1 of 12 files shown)


selina-dxw:
Looks great, thank you!

libuk (approved):
Looks good!

PR #1124, opened 2023-09-18
Mock response from DaysUntilDue

richpjames:
Now that we're using business days in the daysUntilDue function we can no longer predict how many days an assessment will have until it is due as there is an unknown number of working days between the current date and the due date. So, for test purposes, we have to mock the response from daysUntilDue. To make this easier we'll put it in its own file.

### DIFF
--- a/server/utils/assessments/daysUntilDue.test.ts
+++ b/server/utils/assessments/daysUntilDue.test.ts
@@ -1,0 +1,35 @@
+import { addDays, formatISO } from 'date-fns'
+import { assessmentSummaryFactory } from '../../testutils/factories'
+import { daysUntilDue } from './dateUtils'
+import { DateFormats, addBusinessDays } from '../dateUtils'
+
+jest.mock('../dateUtils')
+
+describe('daysUntilDue', () => {
+  it('calls the functions it contains with the correct arguments and returns the result of differenceInBusinessDays', () => {
+    const newDate = new Date()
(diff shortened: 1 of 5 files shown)


server/utils/assessments/daysUntilDue.test.ts
```diff
@@ -0,0 +1,21 @@
...
+import { daysUntilDue } from './dateUtils'
+
+jest.mock('../dateUtils', () => ({
+  DateFormats: {
+    isoToDateObj: jest.fn().mockImplementation(date => new Date(date)),
+    dateObjToIsoDate: jest.fn().mockImplementation(date => formatISO(date)),
+    dateObjToIsoDateTime: jest.fn().mockImplementation(date => formatISO(date)),
+    differenceInBusinessDays: jest.fn().mockReturnValue(1),
+  },
+  addBusinessDays: jest.fn().mockImplementation((date, days) => addBusinessDaysDateFns(date, days)),
```
pezholio, line 12: I feel like this might be easier to follow if we did the mocking in the test itself, rather than the setup here. Would also be good to confirm that `isoToDateObj` and `addBusinessDays` were called with the arguments we expect too.

richpjames: Yeah good shout. I don't know if it's easier to follow but it's definitely a more thorough test now.

pezholio (approved):

PR #1130, opened 2023-09-20
(427) Apply navigation fixes and non-statutory sentence type changes

richpjames:
We have previously had a high number of applications where the sentence type was 'Non statutory'. 
This should be a very rare occurence and after post-hoc inspection each of these cases was not a valid non-statutory case. 
In order to better guide the user to the correct sentence type we add some guidance about what is a valid non-statutory case and if non-statutory is selected we take the user to a page where they're asked to confirmed if the user is managed by MAPPA (as they must be if it is a valid non-stat case).

With the current integration test setup we have this is quite hard to test as it would require adding a fair amount of logic to the 'completeBasicInformation' method in Apply helpers so intead I suggest we test this behavio

As well as this addition I have correct the behaviour of the 'Back' button on a couple of pages of the Apply journey.

[Trello card](https://trello.com/c/DmBrYFNZ/427-design-screens-for-non-stat-sentence-type)

### Before
afte (1)

### After
afte

image

### DIFF
--- a/integration_tests/pages/apply/rehabilitativeInterventions.ts
+++ b/integration_tests/pages/apply/rehabilitativeInterventions.ts
@@ -1,6 +1,7 @@
 import { ApprovedPremisesApplication } from '@approved-premises/api'
 
 import ApplyPage from './applyPage'
+import paths from '../../../server/paths/apply'
 
 export default class RehabilitativeInterventions extends ApplyPage {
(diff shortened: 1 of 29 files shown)


selina-dxw:
LGTM thanks!

pezholio (approved):
LGTM - some nice quality of life improvements too!

PR #1133, opened 2023-09-20
Link to an assessment from an application page

pezholio:
This allows users to quickly view an assessment when an application has been assessed.

I’ve also broken the apply tests up a bit for readability.

# Screenshot

image

### DIFF
--- a/server/views/applications/show.njk
+++ b/server/views/applications/show.njk
@@ -2,6 +2,7 @@
 {% from "govuk/components/back-link/macro.njk" import govukBackLink %}
 {% from "govuk/components/button/macro.njk" import govukButton %}
 {% from "../partials/personDetails.njk" import personDetails %}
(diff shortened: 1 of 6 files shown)


pezholio:
[Access to rejected assessments](https://trello.com/c/17K73GvK/453-access-to-rejected-assessments)

server/views/applications/show.njk
```diff
@@ -42,6 +43,19 @@
         </div>
       </div>
 
+      {% if application.assessmentDecision %}
+        {% set html %}
+        <p>Application was {{ application.assessmentDecision }} on {{ formatDate(application.assessmentDecisionDate) }}</p>
```
richpjames, line 48: does this need `govuk-body`? I'm not 100% when the class should be use but I tend to put it on every `<p>` but I am not sure why it isn't just automatically applied

richpjames (approved):
LGTM

PR #1146, opened 2023-09-29
(165) Apply - Use Delius details if the user doesnt update fields

richpjames:
Previously if the user said they didn't want to update any of the details then we didn't pull through the email, name and phoneNumber from Delius. Now we do.
We have to map from the Delius names for fields to ones we can easily lowercase and still have them make sense.
We also do a check to make sure the detailsToUpdate array exists before we map over it. We also return 'None' for the fields the user would like to update if there aren't any they would like to update so that when someone reads the CYA or read-only version it's a bit more explicit.
 
 This should allow this PR to pass (once this PR is deployed to dev): https://github.com/ministryofjustice/hmpps-approved-premises-e2e/pull/45

### DIFF
--- a/integration_tests/fixtures/applicationData.json
+++ b/integration_tests/fixtures/applicationData.json
@@ -4,8 +4,13 @@
       "emailAddress": "bob@test.com",
       "name": "Bob Smith",
       "phoneNumber": "01234567890",
+      "detailsToUpdate": ["phoneNumber", "emailAddress", "name"],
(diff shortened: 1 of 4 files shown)


Harriethw (approved):
LGTM, though as an outsider it's a bit hard to follow all the different possible shapes of the data! according to the coverage report it looks like there's a few lines not covered by the unit tests so might be worth checking all the bases are covered

PR #1151, opened 2023-10-02
Add paragraphs to text with linebreaks

pezholio:
On some questions, particularly Oasys responses, text that has line breaks is played back to users as a single block of text, making it really hard to read. The adds a `linebreaksToParagraphs` helper which we use in the `taskResponsesAsSummaryListItems` helper to replace single linebreaks with `<br/>` HTML elements and double linebreaks with new `<p>` paragraph elements.

### Before

image

### After 

image


pezholio:
[Paragraph separation of OASys screens](https://trello.com/c/9bMCInQ2/462-paragraph-separation-of-oasys-screens)

libuk (approved):

PR #1153, opened 2023-10-04
Show expected arrival/departure on booking page

pezholio:
This adds a new `bookingInformationRows` helper which combines all the information about a booking into one definition list. If the booking has been marked as arrived, we show the actual arrival details, otherwise we show the expected arrival date. Similarly, if the booking has been marked as departed, we show the departure details, otherwise we show the expected departure date.

## Screenshots

image

image

image


libuk (approved):
Looks good 👍🏾

PR #1155, opened 2023-10-05
(Fix) Authorisation issues

pezholio:
This fixes when an unauthorised user tries to access the system - currently the autherror template tries to render the navigation, which requires a user to be present. As we don't get to that bit of middleware the user gets an error.

I've also added a check for if a user is active. If they are inactive, they see a authentication error, rather than the current (unexpected) behaviour of them being able to access the system, but without any of the roles.

### DIFF
--- a/server/middleware/populateCurrentUser.test.ts
+++ b/server/middleware/populateCurrentUser.test.ts
@@ -1,0 +1,109 @@
+import type { NextFunction, Request, Response } from 'express'
+import { DeepMocked, createMock } from '@golevelup/ts-jest'
+
+import { UserService } from '../services'
+import populateCurrentUser from './populateCurrentUser'
+import { userDetailsFactory } from '../testutils/factories'
+import logger from '../../logger'
+
+jest.mock('../../logger')
+
+describe('populateCurrentUser', () => {
+  const token = 'SOME_TOKEN'
+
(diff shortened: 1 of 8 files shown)


Harriethw (approved):
interesting! In what circumstances would a user be inactive? Seems like something we should have in CAS2 as well. Do you think this should be a default in the Typescript template?

pezholio:
@Harriethw This approach didn't actually work 😊 - I've tweaked where it happens in the middleware now, and added additional tests.

server/middleware/populateCurrentUser.test.ts
```diff
@@ -0,0 +1,109 @@
...
+  it('should catch an error when an error is raised', async () => {
+    const err = new Error()
+
+    ;(userService.getActingUser as jest.Mock).mockImplementation(() => {
+      throw err
+    })
+
+    const middleware = populateCurrentUser(userService)
+
+    await middleware(request, response, next)
```
Harriethw, line 94: minor: I don't think we need to `await` the middleware

Harriethw: oh my bad, we do on some of them!

Harriethw (approved):
LGTM still!

PR #1162, opened 2023-10-10
(454) Performance enhancements on premises page

pezholio:
The premises view page is one of our poorest performing endpoints, as we're calling multiple APIs that request far more information than we need. This PR calls a new Summary endpoint, which is customised to return all the information the premises page needs in one call. We also do a chunk of moving around to move a lot of the view-level functionality to utils, to match the approach we use everywhere else.


pezholio:
[Performance enhancements on /premises/{premisesId}/bookings](https://trello.com/c/dzgfdmMs/454-performance-enhancements-on-premises-premisesid-bookings)

libuk (approved):
Nice work 👍🏾

PR #1165, opened 2023-10-12
(488) Tasklist link returns the first completed section

pezholio:
Currently the `taskLink` helper returns the first question in the task. This is causing users some confusion, as the first question in the Basic Information task is telling the user that the Tier is not applicable, which is only shown when the user has an ineligible tier.

To get around this, rather than blindly returning the first question in the task, and the user has started the task, we return the first question they have completed within that task.

This is not ideal, and we’ll need to iterate on this, for example if we add a mandatory question to the start of a section, but this will at least resolve some of the confusion.

### DIFF
--- a/server/utils/taskListUtils.test.ts
+++ b/server/utils/taskListUtils.test.ts
@@ -28,6 +28,28 @@ describe('taskListUtils', () => {
         )
       })
 
+      it('should return a link to the first completed question page if it exists', () => {
+        task.status = 'in_progress'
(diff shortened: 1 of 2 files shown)


pezholio:
[Clicking on the Basic Information task takes the user to the invalid Tier question](https://trello.com/c/3mxXmGSj/488-clicking-on-the-basic-information-task-takes-the-user-to-the-invalid-tier-question)

PR #1173, opened 2023-10-17
(466) Information not received flow 

richpjames:
# Context

Previously if a user marked that they did not receive the information they requested from the probation practicioner then are not asked the Suitability Assessment questions. 
We need to ask these questions in order to have more detail on why the case has been rejected. 
However it doesn't make sense to ask the user questions about 'Required Actions' needed to support the PiP in an AP as they haven't been accepted into an AP and these questions are part of the 'Assess Application' section. Therefore we only want to show the first task for this section and not the second.

[Trello card](https://trello.com/c/xOGF8miz/466-user-flow-following-information-not-received-to-go-straight-to-reject-the-application)

### Before
after (1)

### After
after

### DIFF
--- a/integration_tests/pages/taskListPage.ts
+++ b/integration_tests/pages/taskListPage.ts
@@ -9,6 +9,10 @@ export default class TaskListPage extends Page {
     cy.get('.app-task-list').should('not.contain', section)
   }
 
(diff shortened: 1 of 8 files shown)


pezholio (approved):
LGTM 👍

PR #1179, opened 2023-10-19
(466) Information not received flow fix

richpjames:
This PR fixes the `informationSetAsNotReceived` function so that it works properly with the `filterSectionTasks` function. This means that a user will be able to mark that they have not received the information required for the assessment then go back and say that they have and see the 'Required Actions' task that they need to complete. Previously they would not have seen the 'Required Actions' task

Previously `informationSetAsNotReceived` checked if the response to the 'informationReceived' question was 'no' AND if the assessment status was 'awaiting_response'.
`informationSetAsNotReceived` is used in two places: a) in the assessmentsController.show method and, b)
in the filterSections function.
In a) we want to check the assessment status is 'awaiting_response' but we do this already in the controller
method so we don't need to do it again in the util.
In b) we don't care about the assessment status and checking that it is 'awaiting_response' before checking
the answer to the question means that if a user changes their mind about the whether they have received
sufficient information they won't be able to complete the 'required-actions' task.


pezholio (approved):
LGTM 👍

PR #1181, opened 2023-10-19
(495) Show taks due dates in business days

pezholio:
This was causing some confusion as the due dates were different on the task list and the assessment list. After some investigation, I found that the `daysUntilDueCell` was counting the difference in days between the due date and the current date in calendar days, not working days. This updates the utils to use the working days count to be more consistent.

### Before

(Task view)

image

(Assessment view)

image

### After

(Task view)

image

(Assessment view)

image


pezholio:
[Days until due should be in working days on Tasklist](https://trello.com/c/D3yqUEEX/495-days-until-due-should-be-in-working-days-on-tasklist)

richpjames (approved):

PR #1198, opened 2023-10-30
Improve types

richpjames:
# Context

[The template that this repo builds on](https://github.com/ministryofjustice/hmpps-template-typescript) does not have [typescript strict mode](https://www.typescriptlang.org/tsconfig#strict) enabled. A lot of the benefit of typechecking is lost without this but unfortunately we realised too late in the development process to turn it on without causing a lot of extra work. Running the TS compiler with strict mode now results in >700 violations compared to running it without strict mode. Despite this I think it would be worthwhile to enable it.

There are a few approaches I can think of to enable it: 
1. Turn it on, fix all the errors. Given an infinite amount of time this is option I would choose. The downsides of this option are: we don't have enough time and it would generate a PR so unwieldy that it would be difficult to review. 
2. Incrementally increase our typesafety. We can run TS in strict mode when we have time to fix some errors and hopefully get to a point where we can turn on strict mode permanently. The downside to this is that new changes would not be tested in strict mode and we would have to remember to find the time to improve the typesafety. 
3. Use [typescript-strict-plugin](https://www.npmjs.com/package/typescript-strict-plugin) or similar to gradually introduce strictness. Currently TS only supports having strict mode on for all files or none. typescript-strict-plugin allows you to gradually introduce strictness by commenting which files the compiler should not check in strict mode. This seems like a good option but it involves adding an additional dependency, commenting on every file to be ignored and running the plugin as well as running the usual typescript compiler. 

Of the options 2 and 3 seem most viable. I suggest we opt for 2 for the time being but if we don't make progress towards being able to turn strict mode on *or* we introduce bugs that would have been caught by strict mode being on then we should progress with 3.  

# Changes in this PR 
This PR contains various fixes that error when TS is run with `strict: true`. The behaviour should be largely unchanged except:
 - some places where we were returning a falsy value that is better implemented by a more accurate falsy value. For instance we use `null`s in a lot of places that makes `strict` error whereas `undefined` doesn't. 
- some places we assumed a value would exist when this might not be the case so we need to handle the cases where the value doesn't exist

### DIFF
--- a/server/config.ts
+++ b/server/config.ts
@@ -7,7 +7,7 @@ const production = process.env.NODE_ENV === 'production'
 
 function get<T>(name: string, fallback: T, options = { requireInProduction: false }): T | string {
   if (process.env[name]) {
-    return process.env[name]
+    return process.env[name] as string
   }
   if (fallback !== undefined && (!production || !options.requireInProduction)) {
     return fallback
@@ -54,7 +54,7 @@ export default {
   },
   redis: {
     host: get('REDIS_HOST', 'localhost', requiredInProduction),
-    port: parseInt(process.env.REDIS_PORT, 10) || 6379,
+    port: parseInt(process.env.REDIS_PORT as string, 10) || 6379,
     password: process.env.REDIS_AUTH_TOKEN,
     tls_enabled: get('REDIS_TLS_ENABLED', 'false'),
   },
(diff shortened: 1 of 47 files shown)


pezholio (approved):
LGTM - thanks for doing this!

PR #1204, opened 2023-10-31
Select index offence when creating a booking

pezholio:
When creating an ad-hoc booking, if an offender has multiple index offences against their name, the user should be prompted to select the index offence. This is needed to ensure that domain events get created in Delius.

This needs to be deployed along with https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1084

image

### DIFF
--- a/integration_tests/tests/manage/booking.cy.ts
+++ b/integration_tests/tests/manage/booking.cy.ts
@@ -1,4 +1,5 @@
 import {
+  activeOffenceFactory,
   applicationFactory,
   assessmentFactory,
   bedDetailFactory,
@@ -36,6 +37,7 @@ context('Booking', () => {
     applicationId: application.id,
     assessmentId: assessment.id,
   })
+  const offences = activeOffenceFactory.buildList(1)
 
   beforeEach(() => {
     cy.task('reset')
@@ -44,6 +46,7 @@ context('Booking', () => {
     cy.task('stubApplicationGet', { application })
     cy.task('stubAssessment', assessment)
     cy.task('stubPremisesSummary', premises)
+    cy.task('stubPersonOffences', { person, offences })
   })
 
   it('should shown an error if I search for an LAO', () => {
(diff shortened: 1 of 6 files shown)


server/controllers/manage/bookingsController.ts
```diff
@@ -37,12 +37,14 @@ export default class BookingsController {
 
       if (crnArr.length) {
         const person = await this.personService.findByCrn(req.user.token, crnArr[0])
+        const offences = await this.personService.getOffences(req.user.token, person.crn)
```
richpjames, line 40: should we perform some validation in the `create` method to ensure an index offence is selected?

pezholio: The validation occurs on the API side 👍

richpjames: Ah of course, I forgot that's how we do it for Manage

integration_tests/tests/manage/booking.cy.ts
```diff
@@ -141,6 +145,55 @@ context('Booking', () => {
...
+    bookingConfirmationPage.verifyBookingIsVisible(booking)
+
+    // And the booking should be created in the API
+    cy.task('verifyBookingCreate', { premisesId: premises.id }).then(requests => {
+      expect(requests).to.have.length(1)
+      const requestBody = JSON.parse(requests[0].body)
+
+      expect(requestBody.arrivalDate).equal(booking.arrivalDate)
+      expect(requestBody.departureDate).equal(booking.departureDate)
+      expect(requestBody.eventNumber).equal(multipleOffences[1].deliusEventNumber)
```
richpjames, line 196: I think it would be worth updating the Booking error test below this and the stub to include a missing `eventNumber`

pezholio: Good plan!

richpjames (approved):

PR #1213, opened 2023-11-02
Check release date in relation to the created date

pezholio:
Fixes https://ministryofjustice.sentry.io/issues/4595156572/?alert_rule_id=14512560&alert_type=issue&notification_uuid=ba2799d8-4dde-4071-b49a-34af020f6ef0&project=4503931667742720&referrer=slack

Because we run validations when showing the application, the release date will be in the past for applications where the release date has already passed, so we get inconsistent behaviour. We fix this by checking the release date in relation to when the application was created. There are still some edge cases where the user may go back and edit the repsonse, but this at least fixes the error in prod that we’re seeing.

### DIFF
--- a/server/form-pages/apply/reasons-for-placement/basic-information/placementDate.test.ts
+++ b/server/form-pages/apply/reasons-for-placement/basic-information/placementDate.test.ts
@@ -1,4 +1,4 @@
-import { add, isFuture, sub } from 'date-fns'
+import { add, sub } from 'date-fns'
 import { itShouldHaveNextValue, itShouldHavePreviousValue } from '../../../shared-examples'
 import { noticeTypeFromApplication } from '../../../../utils/applications/noticeTypeFromApplication'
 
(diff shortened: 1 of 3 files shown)


server/form-pages/apply/reasons-for-placement/basic-information/placementDate.test.ts
```diff
@@ -27,14 +25,14 @@ describe('PlacementDate', () => {
   })
 
   describe('constructor', () => {
-    describe('when the release date is in the future', () => {
+    describe('when the release date is after the application is created', () => {
```
tahb, line 28: Suggestion: Do we need another test here to cover the case when the created at and released at are exactly the same? To cover this logic: `|| isSameDay(applicationDate, releaseDateObj)`

server/form-pages/apply/reasons-for-placement/basic-information/placementDate.test.ts
```diff
@@ -27,14 +25,14 @@ describe('PlacementDate', () => {
...
       it('sets the title and body correctly', () => {
```
tahb, line 29: Suggestion: This is me being new to the code base but what does it meant to be set correctly here? Is it: "it sets the title and body with the release date as the default start date"?

Same comment on the next test if so

tahb (approved):
LGTM though I suspect we might need a new test for sameDay?

PR #1222, opened 2023-11-09
APS 14 - Withdraw a Placement Application

richpjames:
# Context
Users need to be able to withdraw placement applications (PAs). 
To do this we need to list all PAs and add a 'Withdraw' link to each one.

[Jira ticket/card](https://dsdmoj.atlassian.net/jira/software/c/projects/APS/boards/1328?selectedIssue=APS-14)

# Changes in this PR
- The majority of the commits in this PR are fixing bugs and issues I spotted along the way as well as refactoring to 'make the hard work easy'.
- The final two commits are where the bulk of the action happens

### Before
before

### After
after-1
after-2
after-3

### DIFF
--- a/integration_tests/mockApis/applications.ts
+++ b/integration_tests/mockApis/applications.ts
@@ -134,6 +134,22 @@ export default {
         jsonBody: args.timeline,
       },
     }),
+  stubApplicationPlacementRequests: (args: {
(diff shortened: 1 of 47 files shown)


pezholio (approved):
LGTM 👍

PR #1234, opened 2023-11-15
APS 14  - Withdraw placement applications

richpjames:
Tying up some odds an ends related to withdrawing placement applications

### DIFF
--- a/integration_tests/pages/apply/showPage.ts
+++ b/integration_tests/pages/apply/showPage.ts
@@ -4,6 +4,7 @@ import type {
   PlacementApplication,
   TimelineEvent,
 } from '@approved-premises/api'
+import { fromPartial } from '@total-typescript/shoehorn'
 import { DateFormats } from '../../../server/utils/dateUtils'
 
 import { summaryListSections } from '../../../server/utils/applications/summaryListUtils'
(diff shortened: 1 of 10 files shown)


integration_tests/pages/apply/showPage.ts
```diff
@@ -82,7 +83,7 @@ export default class ShowPage extends Page {
   }
 
   clickRequestAPlacementTab() {
-    cy.get('a').contains('Request a placement').click()
+    cy.get('a').contains('Placement Requests').click()
```
pezholio, line 86: I think this is a casing thing causing the tests to fail

server/controllers/apply/applicationsController.test.ts
```diff
@@ -135,13 +137,70 @@ describe('applicationsController', () => {
...
+        applicationService.getPlacementApplications.mockResolvedValue(placementApplications)
+
+        await requestHandler({ ...request, query: { tab: 'placementRequests' } }, response, next)
+
+        expect(response.render).toHaveBeenCalledWith('applications/show', {
+          application,
+          referrer,
+          tab: 'placementRequests',
+          placementApplications,
+          pageHeading: 'Application',
```
pezholio, line 165: I think this needs changing to `'Approved Premises application'`

server/controllers/apply/applicationsController.test.ts
```diff
@@ -158,6 +217,7 @@ describe('applicationsController', () => {
         application,
         referrer,
         tab: 'application',
+        pageHeading: 'Application',
```
pezholio, line 220: Same here - should be `'Approved Premises application'`

pezholio (approved):
LGTM - just a couple of issues with copy causing test failures

PR #1235, opened 2023-11-15
(APS 16) View all applications

pezholio:
This adds the controllers, views, services etc to allow viewing all applications as a paginated list. I've included sorting for now, but we'll be able to filter by CRN / Name and status soon too.

## Screenshot

All applications -- lists all applications with pagination

### DIFF
--- a/integration_tests/mockApis/applications.ts
+++ b/integration_tests/mockApis/applications.ts
@@ -1,9 +1,11 @@
 import { SuperAgentRequest } from 'superagent'
 
 import type {
+  ApplicationSortField,
   ApprovedPremisesApplication,
   ApprovedPremisesApplicationSummary,
   ApprovedPremisesAssessment,
+  SortDirection,
   TimelineEvent,
 } from '@approved-premises/api'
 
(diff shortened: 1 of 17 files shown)


server/utils/applications/utils.ts
```diff
@@ -41,22 +41,32 @@ import ReasonForPlacement, {
...
+    createNameAnchorElement(application.person, application.id),
+    textValue(application.person.crn),
+    htmlValue(getTierOrBlank(application.risks?.tier?.value?.level)),
+    textValue(getArrivalDateorNA(application.arrivalDate)),
+    textValue(DateFormats.isoDateToUIDate(application.createdAt, { format: 'short' })),
+    htmlValue(getStatus(application)),
+  ])
 }
 
+const getTierOrBlank = (tier: string | null) => (tier == null ? '' : tierBadge(tier))
```
richpjames, line 90: would it be safer to check if tier is any falsy value?

integration_tests/mockApis/applications.ts
```diff
@@ -23,6 +23,49 @@ export default {
...
+    page = '1',
+  }: {
+    applications: Array<ApprovedPremisesApplicationSummary>
+    page: string
+  }): SuperAgentRequest => {
+    const queryParameters = {
+      page: {
+        equalTo: page,
+      },
+    } as Record<string, unknown>
```
richpjames, line 49: Is this casting neccessary?

richpjames (approved):
LGTM. A couple of minor comments

PR #1236, opened 2023-11-16
Use dynamic value for todayAtMidnight

richpjames:
We were using a constant for todayAtMidnight which meant that it would always point
to the date it was first used. If the UI wasn't deployed daily this would mean that the date
would be incorrect.
We not return the result of a function so it should be kept up to date


pezholio (approved):
LGTM - nice catch!

PR #1238, opened 2023-11-16
(APS-7) Allow any user to create a placement application

pezholio:
We don’t have any safeguards in the API, so we can remove this check to allow anyone to make applications. The only slight downside is that any booking confirmations go to the original applicant, but I think this is fine for now.


richpjames (approved):
Looks good, do we have a ticket for making the change to the Notify code?

pezholio:
Not yet, I think we'll cross that bridge when we come to it :)

PR #1242, opened 2023-11-17
(APS 16) Allow searching on application dashboard

pezholio:
This extends the applications dashboard to add searching functionality by name/CRN and status, as well as uses the new statuses returned from the API.

This shouldn't be merged until https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1159 is merged.

## Screenshot

image

### DIFF
--- a/server/testutils/factories/applicationSummary.ts
+++ b/server/testutils/factories/applicationSummary.ts
@@ -17,5 +17,5 @@ export default Factory.define<ApplicationSummary>(() => ({
   arrivalDate: DateFormats.dateObjToIsoDate(faker.date.soon()),
   risks: risksFactory.build(),
   createdByUserId: faker.string.uuid(),
-  status: 'inProgress',
+  status: 'started',
 }))
--- a/server/utils/getSearchOptions.ts
+++ b/server/utils/getSearchOptions.ts
@@ -1,0 +1,14 @@
+import type { Request } from 'express'
+
+export const getSearchOptions = <T>(request: Request, keys: Array<string>): T => {
+  const searchOptions = {}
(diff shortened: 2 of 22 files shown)


server/utils/getSearchOptions.ts
```diff
@@ -0,0 +1,14 @@
+import type { Request } from 'express'
+
+export const getSearchOptions = <T>(request: Request, keys: Array<keyof T>): T => {
+  const searchOptions = {} as T
```
richpjames, line 4: If we're casting on line 4 then I think we're best off having `keys` as an `Array<string>` which would mean we don't have to cast on line 7 and 9

script/test
```diff
@@ -10,7 +10,7 @@ npm run lint
 
 echo "==> Pulling the latest type from the API repo"
 
-npm run generate-types
+# npm run generate-types
```
richpjames, line 13: think this snuck in

server/testutils/factories/applicationSummary.ts
```diff
@@ -17,5 +17,5 @@ export default Factory.define<ApplicationSummary>(() => ({
   arrivalDate: DateFormats.dateObjToIsoDate(faker.date.soon()),
   risks: risksFactory.build(),
   createdByUserId: faker.string.uuid(),
-  status: 'inProgress',
+  status: 'started',
```
richpjames, line 20: I think this commit would make more sense towards the start of the PR

richpjames (commented):
Looks good, I have a left a couple of comments

richpjames (approved):

PR #1245, opened 2023-11-20
APS-16: Remove link from All Applications dashboard for in progress applications

richpjames:
If an application is in progress (EG it has the 'started' status) then we don't want
people to be able to click on it and start editing it from the 'All applications' (dashboard)
view. So we pass in a boolean from the view which, when negative,  ensures if an application
isnt linked.


pezholio (approved):
LGTM 👍

PR #1251, opened 2023-11-22
(APS 62) Filter Assessments server side

pezholio:
We've noticed some performance issues with the assessments listing page as we're loading a ton of assessments in memory and filtering client side. As a temporary solution, we're using the status filtering functionality already introduced by CAS3 to do the filtering server side and hopefully cut down on the sheer number of assessments we load from the API. 

## Screenshot

image


server/controllers/assess/assessmentsController.ts
```diff
@@ -15,12 +15,18 @@ export default class AssessmentsController {
 
   index(): RequestHandler {
     return async (req: Request, res: Response) => {
-      const assessments = await this.assessmentService.getAll(req.user.token)
+      const { activeTab } = req.query
+      const statuses =
+        activeTab === 'awaiting_assessment' || !activeTab
+          ? (['in_progress', 'not_started', 'in_review'] as Array<AssessmentStatus>)
```
richpjames, line 21: I think it would be worth defining this array as a variable, typing it and then using it here and in the tests. Then we will be alerted if the `AssessmentStatus` type changes and we only have to update it one place

richpjames (approved):
I've left one small comment but looks good otherwise

server/controllers/assess/assessmentsController.ts
pezholio: Good plan!

PR #1258, opened 2023-11-28
(APS-70) Send sentence type when submitting application

pezholio:
This builds on https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1184 to send the sentence type when submitting an application. We'll be using it in domain events in the future.


server/utils/applications/adjacentPageFromSentenceType.ts
```diff
@@ -1,3 +1,5 @@
+import { SentenceTypeOption } from '../../@types/shared'
+
 export const sentenceTypes = {
```
richpjames, line 3: it would be better to type this rather than cast it

server/utils/applications/getApplicationData.test.ts
```diff
@@ -12,12 +12,13 @@ jest.mock('./utils')
 describe('getApplicationData', () => {
   describe('getApplicationSubmissionData', () => {
     const releaseType = 'license' as ReleaseTypeOption
+    const sentenceType = 'standardDeterminate' as SentenceTypeOption
```
richpjames, line 15: This can be typed as well (as can the `releaseType` above if you have time)

server/utils/applications/getApplicationData.test.ts
```diff
@@ -47,14 +49,15 @@ describe('getApplicationData', () => {
...
+        sentenceType,
         targetLocation,
         arrivalDate,
         isEmergencyApplication: true,
         isEsapApplication: false,
       })
     })
 
-    it('handles when a release type is missing', () => {
+    it('handles when a release type  is missing', () => {
```
richpjames, line 60: there's a double space here

richpjames (commented):
I've left a couple comments

PR #1262, opened 2023-11-29
(APS 72) Record situation when application submitted

pezholio:
This builds on https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1192 and sends an additional `situation` field to the API when the application is submitted. This will allow the correct Delius referral type to be recorded when a booking is made.


richpjames (approved):
LGTM 
I think we may have to update the cypress tests where we verify the payload sent on submission of an application?

PR #1272, opened 2023-12-05
APS 27 - Placement Applications - Multiple ROTL dates

richpjames:
# Context

ROTL cases often have multiple release dates. 
Previously users would have to complete a Placement Application for each release which causes more work and friction. 
This change enables users to add as many Dates Of Placement as they need in one application.

[JIRA Card](https://dsdmoj.atlassian.net/jira/software/c/projects/APS/boards/1328?selectedIssue=APS-27)

# Changes in this PR

### Before

Placement Applications -- allows me to complete form if the reason for placement is ROTL (3)

### After
image

Placement Applications -- allows me to complete form if the reason for placement is ROTL (2)

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,6 +1,6 @@
 # Context
 
-<!-- Is there a Trello ticket you can link to? -->
+<!-- Is there a Jira ticket you can link to? -->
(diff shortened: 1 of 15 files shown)


vubalasu (approved):

pezholio (approved):
LGTM - I can see why this has been such a pain to implement!

PR #1289, opened 2023-12-13
Throw error if OASys summary can't be found for CYA page

richpjames:
We assume that when a summary can't be found on the Check Your Answers pageit's because the OASys pages haven't been completed as we'd expect. Users can click through the tabs instead of clicking the 'Save and continue' button at the bottom of the page which is harder to find.
The result of this is an error on the check your answers screen. In order to see a more helpful error in the logs I think it is worth handling this error.


pezholio (approved):
Good plan!

PR #1290, opened 2023-12-13
Fix/support legacy date fields in placement applications

pezholio:
Before we added support for multiple dates in placement applications, we had different field names. As some placement applications still have this data, we need to support both kinds. If `datesOfPlacement` is not found, then we attempt to fetch from the legacy fields.

I’ve also removed the calls to `addResponseToFormArtifact`, as we’re already mocking  `pageDataFromApplicationOrAssessment`, so adding the responses doesn’t actually do anything.


server/utils/placementRequests/placementApplicationSubmissionData.ts
```diff
@@ -49,14 +52,49 @@ export const durationAndArrivalDateFromRotlPlacementApplication = (dateOfPlaceme
...
+    placementApplication,
+    DatesOfPlacement,
+    'datesOfPlacement',
+  )
+
+  if (datesOfPlacement) {
+    return datesOfPlacement
+  }
+
+  const dateOfPlacement = {}
```
richpjames, line 68: I think this would be better as a `let` without an initializer and then we can avoid casting on line 87

richpjames (approved):
I've left some comments but none of them are essential

server/utils/placementRequests/placementApplicationSubmissionData.ts
```diff
@@ -49,14 +52,49 @@ export const durationAndArrivalDateFromRotlPlacementApplication = (dateOfPlaceme
...
+  const legacyProperties = [
```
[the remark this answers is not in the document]
pezholio: I did wonder that, but I wanted to keep the typing the same, and avoid any nasty surprises

server/utils/placementRequests/placementApplicationSubmissionData.ts
pezholio: Ah, this didn't work 😢

server/utils/placementRequests/placementApplicationSubmissionData.ts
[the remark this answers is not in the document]
richpjames: yeah fair enough

server/utils/placementRequests/placementApplicationSubmissionData.ts
richpjames:
```   
let dateOfPlacement: DateOfPlacement

  const legacyProperties = [
    'arrivalDate',
    'arrivalDate-day',
    'arrivalDate-month',
    'arrivalDate-year',
    'duration',
    'durationDays',
    'durationWeeks',
  ]

  legacyProperties.forEach(property => {
    dateOfPlacement[property] = retrieveQuestionResponseFromFormArtifact(
      placementApplication,
      DatesOfPlacement,
      property,
    )
  })

  return [dateOfPlacement] 
  ``` 
  works for me, but not a big deal if we have to cast

PR #1304, opened 2023-12-19
(AP6 86 / APS 152) Normalise CRN when searching

pezholio:
This removes all trailing/leading spaces and capitalises when searching for a CRN.  When searching for applications and placement requests, the param is `crnOrName`, but when we search for names in the API, these are capitalised anyway, so this won’t make any difference.

### DIFF
--- a/integration_tests/tests/admin/searchPlacementRequests.cy.ts
+++ b/integration_tests/tests/admin/searchPlacementRequests.cy.ts
@@ -2,6 +2,7 @@ import SearchPage from '../../pages/admin/placementApplications/searchPage'
 
 import { placementRequestFactory } from '../../../server/testutils/factories'
 import { PlacementRequestDashboardSearchOptions } from '../../../server/@types/ui'
+import { normaliseCrn } from '../../../server/utils/normaliseCrn'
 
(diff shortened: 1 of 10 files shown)


pezholio:
I've had a change of heart - I think this is better handled on the API side

integration_tests/tests/admin/searchPlacementRequests.cy.ts
```diff
@@ -23,7 +24,11 @@ context('Search placement Requests', () => {
     cy.signIn()
 
     cy.task('stubPlacementRequestsSearch', { placementRequests })
-    cy.task('stubPlacementRequestsSearch', { placementRequests: searchResults, ...searchQuery })
+    cy.task('stubPlacementRequestsSearch', {
+      placementRequests: searchResults,
+      crnOrName: normaliseCrn(searchQuery.crnOrName),
```
richpjames, line 29: I think we want to swap lines 29 and 30 otherwise the non-normalized `crnOrName` will overwrite the normalized `crnOrName`

pezholio: Good call!

PR #1310, opened 2023-12-21
APS-123: Ensure that today is a valid placement date

richpjames:
We need to allow a placement date of today, previously this errored.


pezholio (approved):
Nice! 👍

PR #1359, opened 2024-01-12
(APS-82) Group premises by region

pezholio:
This adds optgroups to the premises select options in the CRU dashboard and the Preferred APs screen in Apply.

To get around the usability issues with optgroups (and also maintain functionality without JS) I've added some frontend JS to allow regions to be selected which then populate only the APs in that region.

### DIFF
--- a/assets/js/flattenPremisesOptions.js
+++ b/assets/js/flattenPremisesOptions.js
@@ -1,0 +1,108 @@
+function flattenPremisesOptions(select, index) {
+  var formGroup = select.parentNode
+  var formElement = formGroup.parentNode
+  var optgroups = select.getElementsByTagName('optgroup')
+  var selectedRegion
+
+  function hideSelect() {
+    formGroup.classList.add('govuk-visually-hidden')
+
+    var options = select.getElementsByTagName('option')
+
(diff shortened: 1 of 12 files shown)


assets/js/flattenPremisesOptions.js
```diff
@@ -0,0 +1,109 @@
+function flattenPremisesOptions(select, index) {
+  var formGroup = select.parentNode
```
richpjames, line 2: I don't think there's much benefit in changing it but for future ref I think it's safe to use `let`/`const` for variables in clientside JS since we no longer support IE11

assets/js/flattenPremisesOptions.js
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  var regions = getRegions()
+  flattenOptions()
+
+  if (regions.length > 1) {
+    if (select.value.length === 0) {
+      hideSelect()
+    }
+    console.log(selectedRegion)
```
richpjames, line 100: oops

richpjames (approved):

richpjames (approved):

PR #1363, opened 2024-01-15
Refactor Assessment integration tests

richpjames:
Previously our Assessment integration (Cypress) tests were stubbed statelessly. 
This meant we were unable to add stateful questions (ones that depend on the answer to a previous question) easily. 
This PR refactors the assessment tests to use the same approach that we use for the Apply integration tests which will make adding stateful behaviour easier in future.


pezholio (approved):
Looks good from a cursory glance. Like you say, it's a tricky one to review, but given the tests pass, and this has been something I've been keen to get in, I'm gonna go ahead and approve this 👍

PR #1375, opened 2024-01-17
(APS-209) Fix incorrect lost beds count

pezholio:
This adds a `lostBedsCountForToday` helper to gather and count all the beds that are currently lost on a given day. I’ve also added a `getRandomInt` helper to build a random number of lost beds in the tests to make them a bit more robust.

### DIFF
--- a/server/views/lostBeds/index.njk
+++ b/server/views/lostBeds/index.njk
@@ -30,7 +30,7 @@
                 text: "Out of service"
               },
               value: {
-                text: numberOfLostBeds
+                text: LostBedUtils.lostBedsCountForToday(lostBeds)
               }
             }
           ]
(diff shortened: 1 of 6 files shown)


server/views/lostBeds/index.njk
```diff
@@ -30,7 +30,7 @@
                 text: "Out of service"
               },
               value: {
-                text: numberOfLostBeds
```
richpjames, line 33: can we remove the numberOfLostBeds param passed into the view ?

richpjames (approved):

server/views/lostBeds/index.njk
pezholio: Good call!

PR #1386, opened 2024-01-19
(APS 83) - Add drop down clusters in manage to allow user to select preferred region

richpjames:
# Context
When we go multi-region users will want to be able to filter the APs by region. 
[Jira](https://dsdmoj.atlassian.net/jira/software/c/projects/APS/boards/1328?selectedIssue=APS-83)

### Before
after (2)

### After
after (1)

# Release checklist

confirm:

## Pre merge checklist

      advance?

## Post merge checklist

considering it done.

fix it as soon as possible.

### DIFF
--- a/server/data/premisesClient.ts
+++ b/server/data/premisesClient.ts
@@ -20,8 +20,12 @@ export default class PremisesClient {
     this.restClient = new RestClient('premisesClient', config.apis.approvedPremises as ApiConfig, token)
   }
 
-  async all(): Promise<Array<PremisesSummary>> {
-    return (await this.restClient.get({ path: paths.premises.index({}) })) as Array<PremisesSummary>
+  async all(apAreaId: string = ''): Promise<Array<PremisesSummary>> {
+    const query = apAreaId ? { apAreaId } : {}
+    return (await this.restClient.get({
+      path: paths.premises.index({}),
+      query,
+    })) as Array<PremisesSummary>
   }
 
   async find(id: string): Promise<ApprovedPremises> {
--- a/server/data/referenceDataClient.ts
+++ b/server/data/referenceDataClient.ts
@@ -1,6 +1,7 @@
 import type { ReferenceData } from '@approved-premises/ui'
 import RestClient from './restClient'
 import config, { ApiConfig } from '../config'
+import { ApArea, ProbationRegion } from '../@types/shared'
 
 export default class ReferenceDataClient {
   restClient: RestClient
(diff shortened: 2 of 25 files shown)


server/data/premisesClient.ts
```diff
@@ -20,8 +20,12 @@ export default class PremisesClient {
     this.restClient = new RestClient('premisesClient', config.apis.approvedPremises as ApiConfig, token)
   }
 
-  async all(): Promise<Array<PremisesSummary>> {
-    return (await this.restClient.get({ path: paths.premises.index({}) })) as Array<PremisesSummary>
+  async all(probationRegionId: string = ''): Promise<Array<PremisesSummary>> {
```
pezholio, line 23: Do we need this, given we're filtering on the frontend?

server/routes/manage.ts
```diff
@@ -27,6 +27,7 @@ export default function routes(controllers: Controllers, router: Router, service
   } = controllers
 
   get(paths.premises.index.pattern, premisesController.index(), { auditEvent: 'LIST_PREMISES' })
+  post(paths.premises.index.pattern, premisesController.index(), { auditEvent: 'LIST_PREMISES' })
```
pezholio, line 30: Should this auditEvent have a name like `FILTER_PREMISES`?

server/services/premisesService.ts
```diff
@@ -14,9 +14,9 @@ import { mapApiOccupancyToUiOccupancy } from '../utils/premisesUtils'
 export default class PremisesService {
   constructor(private readonly premisesClientFactory: RestClientBuilder<PremisesClient>) {}
 
-  async getAll(token: string): Promise<Array<ApprovedPremisesSummary>> {
+  async getAll(token: string, probationRegionId = ''): Promise<Array<ApprovedPremisesSummary>> {
     const premisesClient = this.premisesClientFactory(token)
-    const premises = await premisesClient.all()
+    const premises = await premisesClient.all(probationRegionId)
```
pezholio, line 19: We don't seem to use this anywhere?

pezholio (commented):
Just a coupla comments. Looks good overall though

server/services/premisesService.ts
richpjames: yeah my bad, I'll use the API side filtering

server/data/referenceDataClient.ts
```diff
@@ -12,4 +13,8 @@ export default class ReferenceDataClient {
   async getReferenceData(objectType: string): Promise<Array<ReferenceData>> {
     return (await this.restClient.get({ path: `/reference-data/${objectType}` })) as Array<ReferenceData>
   }
+
```
vubalasu, line 16: is there any specific reason for adding getProbationRegions since for rest of reference data we are directly calling getReferenceData

server/services/premisesService.ts
richpjames: I've deleted the Cypress test for the filtering (that I had previously added) but will write an E2E for it

server/data/referenceDataClient.ts
richpjames: there is a note in the commit message: 
"Probation Regions have a different type signature to the rest of the Reference data.
Rather than having to handle this in the consuming code it makes sense that they have their
own method". 
If we use the same endpoints for both types of data we have to check which type of data it is in all places we call `getReferenceData`

pezholio (approved):
LGTM 👍

vubalasu (approved):

PR #1405, opened 2024-01-23
APS-184 - Fix - Add getter for area

richpjames:
If the area property isnt in the body then we want to use the version from Delius. We need this because the 'area' is a first class field which is sent to the API in getApplicationData function


pezholio (approved):
LGTM 👍

PR #1412, opened 2024-01-25
APS-199 - Ensure that the status filter is retained when another filter is applied

richpjames:
The status filter on the admin Placement Request dashboard (CRU dashboard) is applied using tabs. When the user is on a given tab and they click to apply additional filters to the tab they would not expect the tab to then revert to default (notMatched). We need to retain the tab filter by passing it through the template as a hidden input.

### Before
output2

### After
https://github.com/ministryofjustice/hmpps-approved-premises-ui/assets/50752284/b03333c9-3e9e-4fd5-bd27-7972b24ccff8

# Release checklist

As part of our continuous deployment strategy we must ensure that this work is
ready to be released at any point. Before merging to `main` we must first
confirm:

## Pre merge checklist

- [ ] Have all changes to any dependencies been deployed to both `preprod` and
      `production` in a backwards compatible way? (This includes our API,
      infrastructure, third-party integrations and any secrets or environment variables)
- [ ] Has a required data migration been included in this change or been done in
      advance?

## Post merge checklist

Once we've merged we now need to release this through to production before
considering it done.

[Find the build-and-deploy job in CircleCI](https://app.circleci.com/pipelines/github/ministryofjustice/hmpps-temporary-accommodation-ui)
and work through the following steps:

- [ ] Has the product manager asked to provide approval for this feature?
  - [ ] Has the product manager approved?
- [ ] Manually approve release to preprod
- [ ] Verify change released to preprod
- [ ] Manually approve release to prod
- [ ] Verify change released to production

Should a release fail at any step, you as the author should now lead the work to
fix it as soon as possible.

PR #1421, opened 2024-01-30
(Fix) Populate and set user's AP Area

pezholio:
I made an incorrect assumption that a user’s area already existed in the session, and it was an ID. This PR adds the Ap Area to the User object in the session and ensures it's set correctly in the controller / view.

I've also updated the CRU dashboard to do the same thing in this PR, as well as got rid of some logic that we no longer needed following the changes in #1412

### DIFF
--- a/server/controllers/admin/placementRequests/placementRequestsController.ts
+++ b/server/controllers/admin/placementRequests/placementRequestsController.ts
@@ -1,6 +1,6 @@
 import type { Request, Response, TypedRequestHandler } from 'express'
 import { ApAreaService, PlacementRequestService } from '../../../services'
-import { PlacementRequestSortField } from '../../../@types/shared'
+import { PlacementRequestRequestType, PlacementRequestSortField, PlacementRequestStatus } from '../../../@types/shared'
 import paths from '../../../paths/admin'
 import { PlacementRequestDashboardSearchOptions } from '../../../@types/ui'
 import { getPaginationDetails } from '../../../utils/getPaginationDetails'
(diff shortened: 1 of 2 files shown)


server/views/tasks/index.njk
```diff
@@ -26,8 +26,7 @@
             </div>
           </div>
           <div class="govuk-grid-row">
-            <div class="govuk-grid-column-one-third">
-              {{ govukSelect({
+            <div class="govuk-grid-column-one-third">              {{ govukSelect({
```
lookupdaily, line 29: could you revert this formatting so that govukSelect is on a new line?

server/controllers/admin/placementRequests/placementRequestsController.ts
```diff
@@ -14,13 +14,9 @@ export default class PlacementRequestsController {
...
   index(): TypedRequestHandler<Request, Response> {
     return async (req: Request, res: Response) => {
-      const [apAreaId, requestType] = ['apArea', 'requestType'].map(filter =>
-        req.query[filter] ? req.query[filter] : req.body[filter],
-      )
-      let status = req.query.status ? req.query.status : req.body.status
-      if (status === undefined) {
-        status = 'notMatched'
-      }
+      const status = (req.query.status ? req.query.status : 'notMatched') as PlacementRequestStatus
```
lookupdaily, line 17: 👍

lookupdaily (approved):
I think this looks good, just a small formatting comment :)

server/views/tasks/index.njk
pezholio: 👍

PR #1448, opened 2024-02-02
Withdrawals fixes

richpjames:
# Changes
1. Show the withdrawables select screen when there is only an Application to withdraw
2. Open booking/placement requests views in new tabs
3. Bookings without application IDs use the old withdrawables flow


pezholio (approved):
Approving this in Rich's absence. Most of the stuff I added was grunt work to fix the tests

PR #1452, opened 2024-02-02
Feature flag withdrawals behaviour

pezholio:
There are a few issues with the Referrals behaviour, and we need to ensure that the changes are communicated before we go live. This adds a `NEW_WITHDRAWALS_FLOW_DISABLED` feature flag, which, if present, links directly to the withdrawal/cancellation pages, skipping the interstitial page that prompts the user to select a withdrawable item.


helm_deploy/values-prod.yaml
```diff
@@ -17,6 +17,7 @@ generic-service:
     COMMUNITY_ACCOMMODATION_API_SERVICE_NAME: approved-premises
     COMMUNITY_ACCOMMODATION_API_TIMEOUT_RESPONSE: 30000
     COMMUNITY_ACCOMMODATION_API_TIMEOUT_DEADLINE: 30000
+    NEW_WITHDRAWLS_FLOW_DISABLED: 1
```
richpjames, line 20: I'm happy to be wrong here, but is it not 'withdrawal'?

richpjames (approved):

helm_deploy/values-prod.yaml
pezholio: Oh man, I struggle so much with that word!

helm_deploy/values-prod.yaml
```diff
@@ -17,6 +17,7 @@ generic-service:
     COMMUNITY_ACCOMMODATION_API_SERVICE_NAME: approved-premises
     COMMUNITY_ACCOMMODATION_API_TIMEOUT_RESPONSE: 30000
     COMMUNITY_ACCOMMODATION_API_TIMEOUT_DEADLINE: 30000
+    NEW_WITHDRAWALS_FLOW_DISABLED: 1
```
tahb, line 20: I might consider putting this into every environment so your e2e tests will run against it and any manual QA you do will have high parity

pezholio: Yeah, I considered that, but that means we'd have to change the e2e tests

PR #1454, opened 2024-02-03
Fix withdrawal reasons

pezholio:
Some of these reasons weren’t matching up with the ones in the API, so we were getting Bad Request errors. This tightens up the reasons so they’re more strongly typed, so we know the API is going to accept them.


libuk (approved):

richpjames (approved):
There's a typo in the withdrawalReasons.ts file name but other than that this looks good

pezholio:
@richpjames Why can I not spell "withdrawal"?!? 😆

PR #1457, opened 2024-02-05
DRY up `filterByType` lookups

pezholio:
We had `filterByType` in two places, as we’ll probably be using this elsewhere, it makes sense to have the code in one place.


server/utils/utils.ts
```diff
@@ -180,3 +180,18 @@ export const linebreaksToParagraphs = (text: string) =>
...
   return Math.floor(Math.random() * (Math.floor(max) - Math.ceil(min) + 1)) + Math.ceil(min)
 }
+
+/**
+ * Filters a record with the given keys from a master list of records
+ * @param keys An array of keys to return from `lookup`
+ * @param lookup A key/value of records
+ * @returns A key/value list with the given keys in `keys`
+ */
+export const filterByType = <T extends `${string}`>(
```
richpjames, line 190: I think `T extends ${string}` is the same as `T extends string` ?

richpjames (approved):

server/utils/utils.ts
pezholio: Ah, you're right 👍

PR #1458, opened 2024-02-05
Add link to Withdraw application from the application show page

richpjames:
This was in [the mocks](https://www.figma.com/file/kP7jyX3L01CQap6ErAX648/Approved-Premises-Beta?type=design&mode=design&t=JvIQmfnWt288NG7I-0) but missed on implementation.

### Before
image

### After
image

### DIFF
--- a/server/views/applications/show.njk
+++ b/server/views/applications/show.njk
@@ -5,6 +5,7 @@
 {% from "./partials/_readonly-application.njk" import applicationReadonlyView %}
 {% from "./partials/_timeline.njk" import timeline %}
 {% from "./partials/_request-a-placement.njk" import requestAPlacement %}
+{%- from "moj/components/identity-bar/macro.njk" import mojIdentityBar -%}
 {%- from "govuk/components/tag/macro.njk" import govukTag -%}
 
 {% extends "../partials/layout.njk" %}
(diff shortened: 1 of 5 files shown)


server/views/applications/show.njk
```diff
@@ -60,17 +68,45 @@
...
-      {% if tab === ApplyUtils.applicationShowPageTabs.timeline %}
-        {{ timeline(mapTimelineEventsForUi(timelineEvents), application, csrfToken) }}
-      {% elif tab === ApplyUtils.applicationShowPageTabs.placementRequests %}
-        {{ requestAPlacement(placementApplications, application, user, csrfToken) }}
-      {% else %}
-        {{ applicationReadonlyView(application) }}
-      {% endif %}
+      {% if tab === ApplyUtils.applicationShowPageTabs.application %}
+        <div class="govuk-grid-column-one-quarter">
+          {{ mojButtonMenu({
```
pezholio, line 78: I think we should be using the `mojIdentityBar` helper here

server/views/applications/show.njk
```diff
@@ -60,17 +68,50 @@
...
-        {{ timeline(mapTimelineEventsForUi(timelineEvents), application, csrfToken) }}
-      {% elif tab === ApplyUtils.applicationShowPageTabs.placementRequests %}
-        {{ requestAPlacement(placementApplications, application, user, csrfToken) }}
-      {% else %}
-        {{ applicationReadonlyView(application) }}
-      {% endif %}
+      {% if tab === ApplyUtils.applicationShowPageTabs.application %}
+        <div class="govuk-grid-column-one-quarter govuk-!-padding-left-0 govuk-!-padding-right-0">
+
+          {{ mojIdentityBar({
```
pezholio, line 79: If you look at the bookings/show.njk template, the actions are to the right of the heading e.g:

image

I think we should be consistent with that approach. Happy to be wrong though!

richpjames: OK I get you. 
It's not what the mocks look like but it would be much simpler for us so I think we should go with it for now

pezholio: Cool - yeah, I think it makes sense to lean on existing patterns that are in use elsewhere 👍

pezholio (approved):
LGTM 👍

PR #1463, opened 2024-02-05
APS 282 - filter users when viewing a task

lookupdaily:
# Context

This adds the option to filter users by AP Area and/or qualification on a task allocations page.
As we are filtering by AP Area, have also changed the 'region' column to show 'AP Area'.

[Jira](https://dsdmoj.atlassian.net/browse/APS-282?atlOrigin=eyJpIjoiM2NmN2VmNmQ0NDU1NDA4N2FjMmQyYWU2Yzc0NDQ0MzUiLCJwIjoiamlyYS1zbGFjay1pbnQifQ)

# Changes in this PR

- Change 'Region' column to AP Area
- Add a filter form to the view to match other table filters
- Update tasksService to filter users associated with a task by new filter parameters
- Add a method to taskUtils class to return the task parameters with the correct kebab case for task type
- Add integration tests for different filter scenarios

### Before

User table on task allocations page with no filters:

### After

Now able to filter the user table on task allocations page:

https://github.com/ministryofjustice/hmpps-approved-premises-ui/assets/50752284/c9e99a97-e1a9-434c-ac71-b4448368493c

# Release checklist

confirm:

## Pre merge checklist

      advance?

## Post merge checklist

considering it done.

fix it as soon as possible.

### DIFF
--- a/server/services/taskService.test.ts
+++ b/server/services/taskService.test.ts
@@ -1,5 +1,5 @@
 import { CategorisedTask, PaginatedResponse } from '@approved-premises/ui'
-import { Task } from '../@types/shared'
+import { Task, UserQualification } from '../@types/shared'
 import TaskClient from '../data/taskClient'
 import {
   paginatedResponseFactory,
@@ -7,6 +7,7 @@ import {
   placementRequestTaskFactory,
   taskFactory,
   taskWrapperFactory,
+  userWithWorkloadFactory,
 } from '../testutils/factories'
 import TaskService from './taskService'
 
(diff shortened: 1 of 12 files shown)


server/services/taskService.test.ts
```diff
@@ -103,6 +104,50 @@ describe('taskService', () => {
...
       expect(taskClient.find).toHaveBeenCalledWith(applicationId, 'assessment')
     })
+
+    it('filters users result by AP area and qualification', async () => {
+      const apArea = {
+        id: '0544d95a-f6bb-43f8-9be7-aae66e3bf244',
+        name: 'Midlands',
+      }
+
+      const qualification = 'womens' as UserQualification
```
richpjames, line 114: `const qualification: UserQualification = 'womens' as const` 
      is preferable as it retains type safety

server/services/taskService.ts
```diff
@@ -65,10 +66,23 @@ export default class TaskService {
...
+  async find(
+    token: string,
+    premisesId: string,
+    taskType: string,
+    userFilters?: { apAreaId?: string; qualification?: UserQualification },
+  ): Promise<TaskWrapper> {
     const taskClient = this.taskClientFactory(token)
 
     const task = await taskClient.find(premisesId, taskType)
+    if (userFilters) {
```
richpjames, line 78: This will always be true if `userFilters` is an object as even an empty object is truthy.

richpjames (commented):
Good work as always Liz! 
I've left a couple of minor comments

server/services/taskService.ts
lookupdaily: Ah but it will not be falsy if no argument provided? As it is not a required parameter

richpjames: It would be if but I *think* the only place we're calling it is in the tasksController show method where we're passing in an object. The object properties could be undefined but the object itself will always be defined. Ignore me if we're calling this somewhere else or if I missed something!

lookupdaily: Ah but it will not be falsy if no argument provided? As it is not a required parameter

If I remove the if statement, I will always need to provide an object

lookupdaily: yea - fair enough! Shall amend to make it required :)

lookupdaily: have kept an if statement for each of the properties to prevent an unnecessary loop - what do you think?

richpjames: yeah that's a good idea

richpjames (approved):

PR #1466, opened 2024-02-06
(APS-240) Record appeal decision

pezholio:
This adds logic to record an appeal decision against a rejected application. Only appeal managers should see the link to do this, and it should only show when the application has been rejected

## Screenshot

image

image

### DIFF
--- a/server/controllers/apply/appealsController.test.ts
+++ b/server/controllers/apply/appealsController.test.ts
@@ -1,0 +1,143 @@
+import { NextFunction, Request, Response } from 'express'
+import { DeepMocked, createMock } from '@golevelup/ts-jest'
+import type { ErrorsAndUserInput } from '@approved-premises/ui'
+
+import AppealService from '../../services/appealService'
+import AppealsController from './appealsController'
+import { catchValidationErrorOrPropogate, fetchErrorsAndUserInput } from '../../utils/validation'
+import paths from '../../paths/apply'
+import { AppealDecision } from '../../@types/shared'
+import { ApplicationService } from '../../services'
+import { applicationFactory } from '../../testutils/factories'
+
+jest.mock('../../utils/validation')
+
+describe('AppealsController', () => {
+  const token = 'SOME_TOKEN'
+  const application = applicationFactory.build()
+
+  const request: DeepMocked<Request> = createMock<Request>({ user: { token } })
+  const response: DeepMocked<Response> = createMock<Response>({})
+  const next: DeepMocked<NextFunction> = createMock<NextFunction>({})
+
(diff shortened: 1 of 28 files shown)


server/data/appealClient.test.ts
```diff
@@ -0,0 +1,44 @@
...
+import describeClient from '../testutils/describeClient'
+import paths from '../paths/api'
+
+describeClient('AppeasClient', provider => {
+  let appealsClient: AppealClient
+
+  const token = 'token-1'
+
+  beforeEach(() => {
+    config.flags.oasysDisabled = false
```
richpjames, line 13: do we need this?

server/data/appealClient.test.ts
```diff
@@ -0,0 +1,44 @@
+import AppealClient from './appealClient'
+import config from '../config'
+import { appealFactory, newAppealFactory } from '../testutils/factories'
+import describeClient from '../testutils/describeClient'
+import paths from '../paths/api'
+
+describeClient('AppeasClient', provider => {
```
richpjames, line 7: typo

server/controllers/apply/appealsController.test.ts
```diff
@@ -0,0 +1,143 @@
...
+      })
+    })
+  })
+
+  describe('create', () => {
+    it.each([
+      ['accepted', 'Assessment reopened'],
+      ['rejected', 'Appeal marked as rejected'],
+    ])(
+      'creates an nonArrival with a decision of %s and redirects to the premises page',
```
richpjames, line 82: this test description needs updating

server/utils/applications/applicationIdentityBar.ts
```diff
@@ -29,15 +32,30 @@ export const applicationMenuItems = (application: Application): Array<IdentityBa
         'data-cy-withdraw-application': application.id,
       },
     },
-  ]
+  ] as Array<IdentityBarMenuItem>
```
richpjames, line 35: this is better typed than cast

richpjames (commented):
Looks good, thanks for this parting gift.

server/data/appealClient.test.ts
lookupdaily: I don't believe we do.. it doesn't seem to be used in the appealsClient, and if I remove it the tests pass

server/data/appealClient.test.ts
lookupdaily, line 1: Should we name this 'appealsClient' to match 'appealsController' ?

richpjames: controllers are plural, clients and services are singular

edavey (approved):
Nice work team!

PR #1478, opened 2024-02-12
Fix - Allow a null to be passed in the selectedList array

richpjames:
We are seeing 'null' being passed in as an array value in production. Whilst we track down the cause of this we can handle the error by only checking the 'section' if the OASysSection exists. This will allow users to proceed without having the page error.


richpjames:
Closing as this doesnt address the route cause of the issue

PR #1487, opened 2024-02-15
Revert "Fix/APS-291 add application to withdrawable radio options"

richpjames:
This erroneously added applications to the withdrawable selection view


edavey (approved):
Is it worth tweaking the integration test to verify that there are *only* the expected radio options on the "What do you want to withdrawal" form?

richpjames:
@edavey I made this PR in haste hoping to fix the issue before demoing but I think it's better I take a more thorough approach. I think we will need to make this change but also others too.

PR #1491, opened 2024-02-15
Fix/ Remove applications from selectable withdrawables

richpjames:
Since applications were added to the [withdrawable.type enum](https://github.com/ministryofjustice/hmpps-approved-premises-ui/blob/83eab29d000ae1c8b0dfde2b73e213d491ca3246/server/%40types/shared/models/WithdrawableType.ts#L5) they can be passed into the [withdrawableRadioOptions](https://github.com/ministryofjustice/hmpps-approved-premises-ui/blob/83eab29d000ae1c8b0dfde2b73e213d491ca3246/server/utils/applications/withdrawables/index.ts#L59) function.
It doesn't make sense to show them when selecting a withdrawable to withdraw as I did in [#1477](https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/1477) so I reverted it in [#1487](https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/1487) but this meant that the new withdrawals journey errored. Now, we filter out the application from the list of possible withdrawables which should prevent the error and prevent applications showing up as a selectable withdrawable.

### DIFF
--- a/server/utils/applications/withdrawables/index.test.ts
+++ b/server/utils/applications/withdrawables/index.test.ts
@@ -113,5 +113,20 @@ describe('withdrawableTypeRadioOptions', () => {
         },
       ])
     })
+
+    it('filters out applications', () => {
+      const paWithdrawable = withdrawableFactory.build({ type: 'placement_application' })
+      const applicationWithdrawable = withdrawableFactory.build({ type: 'application' })
+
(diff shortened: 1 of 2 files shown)


edavey (approved):
👍 thks for the fix

PR #1494, opened 2024-02-16
Fix/only allow selected withdrawable types to be selectable

richpjames:
# Changes in this PR
Following #1491 #1487 this PR ensures that the 'Select your placement'/'Select your placement request' screens only show the withdrawables of the type the user has shown on the preceding screen. 
Previously we would show all the possible withdrawables. Now if a use selects 'placement' on the 'What do you want to withdraw?' screen they will only see placements on the following screen. 
This slipped through the net as we were only testing that the withdrawables we wanted to see were in the view but now we also test that they withdrawables that shouldn't be visible aren't there also.

### DIFF
--- a/integration_tests/pages/apply/newWithdrawal.ts
+++ b/integration_tests/pages/apply/newWithdrawal.ts
@@ -1,10 +1,35 @@
 import { Withdrawable } from '../../../server/@types/shared'
 import matchPaths from '../../../server/paths/match'
+import { SelectedWithdrawableType } from '../../../server/utils/applications/withdrawables'
 import Page from '../page'
 
 export default class SelectWithdrawableTypePage extends Page {
   constructor(heading: 'What do you want to withdraw?' | 'Select your placement') {
     super(heading)
+  }
+
+  shouldShowWithdrawableTypes(types: Array<SelectedWithdrawableType>) {
+    types.forEach(type => {
+      cy.get(`input[value="${type}"]`).should('exist')
+    })
+  }
+
+  shouldNotShowWithdrawableTypes(types: Array<SelectedWithdrawableType>) {
+    types.forEach(type => {
+      cy.get(`input[value="${type}"]`).should('not.exist')
+    })
+  }
+
+  shouldShowWithdrawables(withdrawables: Array<Withdrawable>) {
+    withdrawables.forEach(withdrawable => {
+      cy.get(`input[value="${withdrawable.id}"]`).should('exist')
+    })
+  }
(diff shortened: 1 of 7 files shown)


server/controllers/apply/withdrawablesController.test.ts
```diff
@@ -50,21 +53,21 @@ describe('withdrawablesController', () => {
...
         })
       })
     })
 
     describe('Bookings', () => {
-      it(`renders the view`, async () => {
+      it(`renders the view, calling the booking service to retrieve bookings`, async () => {
         const selectedWithdrawableType = 'placement'
-        const withdrawables = withdrawableFactory.buildList(2, { type: 'booking' })
+        const bookingWithdrawables = withdrawableFactory.buildList(2, { type: 'booking' })
```
yndajas, line 64: `placementWithdrawables`?

integration_tests/pages/apply/newWithdrawal.ts
```diff
@@ -1,12 +1,37 @@
...
+  shouldNotShowWithdrawableTypes(types: Array<SelectedWithdrawableType>) {
+    types.forEach(type => {
+      cy.get(`input[value="${type}"]`).should('not.exist')
+    })
+  }
+
+  shouldShowWithdrawables(withdrawables: Array<Withdrawable>) {
+    withdrawables.forEach(type => {
+      cy.get(`input[value="${type.id}"]`).should('exist')
+    })
```
yndajas, line 26: `type` -> `withdrawable` (here and below)?

integration_tests/tests/withdrawals/withdrawals.cy.ts
```diff
@@ -50,12 +56,20 @@ context('Withdrawals', () => {
 
     // Then I am asked what I want to withdraw
     const newWithdrawalPage = new NewWithdrawalPage('What do you want to withdraw?')
+
+    // And I am shown the correct withdrawables
+    newWithdrawalPage.shouldShowWithdrawableTypes(['placementRequest', 'placement'])
```
yndajas, line 61: `placementRequest` here includes placement applications?

yndajas (approved):
I think it makes sense - just a few typos

integration_tests/tests/withdrawals/withdrawals.cy.ts
richpjames: Yeah in the withdrawables journey, for the user, they are the same thing

PR #1495, opened 2024-02-16
Move E2Es into UI repo

richpjames:
Following from CAS2 teams approach: https://github.com/ministryofjustice/hmpps-community-accommodation-tier-2-ui/pull/482 

It’s important to note that e2e tests are required to pass for deploying the frontend to preprod and prod.

We found that it can be difficult to coordinate the frontend changes (a pull request in 1 repo) alongside the changes to the e2e tests (a pull request in another repo).

Both must be worked on and tested locally from different repos. We also have double the admin of proposing and reverting two pull requests, rather than just one.

We’d like to move these tests from a separate repo to the same repo.
Changes

### Changes

 - We install playwright and copy over the e2e tests from the separate repo.


playwright-report/index.html
```diff
@@ -0,0 +1,69 @@
+
+
+<!DOCTYPE html>
```
pezholio, line 3: This should probably be `.gitignore`d

playwright.config.ts
```diff
@@ -0,0 +1,74 @@
+import 'dotenv/config'
```
pezholio, line 1: Can this sit in the `e2e` dir?

server/@types/shared/models/ApprovedPremisesApplication.ts
```diff
@@ -2,31 +2,29 @@
...
-    status: ApprovedPremisesApplicationStatus;
-    assessmentId?: string;
-    assessmentDecision?: AssessmentDecision;
-    assessmentDecisionDate?: string;
-    submittedAt?: string;
-    personStatusOnSubmission?: PersonStatus;
-    apArea?: ApArea;
-});
-
+import type { AnyValue } from './AnyValue'
```
pezholio, line 5: How come this file has changed?

pezholio (approved):
LGTM - I think this is deffo worth doing. Just one thing on a folder that I don't think needs committing.

We'll also need to make sure the API is updated too once this is live

e2e/playwright/.auth/user.json
```diff
@@ -0,0 +1,25 @@
+{
```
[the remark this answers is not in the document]
richpjames: oo yeah

tahb:
[Recently discovered that I forgot to add additional folder ignores for tsconf](https://mojdt.slack.com/archives/C05GEPLBRN3/p1714121815149429?thread_ts=1714121651.920579&cid=C05GEPLBRN3) which you may also need.

richpjames:
@tahb good shout, thanks

PR #1507, opened 2024-02-21
Tweak application tag styling

pezholio:
Looking at the styling for the application tag made my teeth itch a bit, so I had a little play with some of the CSS to make the tag show inline a bit more nicely. It’s not pixel perfect by any stretch of the imagination, but it’s a bit of an improvement.

# Screenshots

## Before

image

## After

image


richpjames (approved):
Nice!

PR #1508, opened 2024-02-21
Filter out application withdrawables from sorted withdrawables

richpjames:
Applications won't have start dates so it doesn't make sense to sort them. Applications won't be shown on the 'Withdrawable select' screen.


server/utils/applications/withdrawables/sortWithdrawables.test.ts
```diff
@@ -6,6 +6,14 @@ import { sortWithdrawables } from './sortWithdrawables'
 describe('sortWithdrawables', () => {
   const { endDate } = withdrawableFactory.build().dates[0]
 
+  it('should filter out withdrawables with ', () => {
```
yndajas, line 9: `with type "application"`?

server/utils/applications/withdrawables/sortWithdrawables.test.ts
```diff
@@ -6,6 +6,14 @@ import { sortWithdrawables } from './sortWithdrawables'
 describe('sortWithdrawables', () => {
```
yndajas, line 6: So we never want applications to be selectable?

Given the sorted withdrawables are currently always filtered, would it make more sense for the included types to be an argument to the method so you can do a single filter:

```ts
const sortAndFilterWithdrawables = (
  withdrawables: Array<Withdrawable>,
  typesToInclude: Array<Withdrawable['type']>
) => {
  return withdrawables.filter(withdrawable => typesToInclude.includes(withdrawable.type)
  ...
}
```

Otherwise you're filtering out applications, then sorting the remaining withdrawables, then filtering some more types out in the controller

yndajas (approved):
This should work, but I think there's an opportunity to simplify

server/utils/applications/withdrawables/sortWithdrawables.test.ts
richpjames: yeah this is a much better approach, I was going for something quick but it's better to try and improve readability if we can

PR #1509, opened 2024-02-22
Add back button to Withdrawable select page

richpjames:
This will help the user navigate back if they change their mind about the type of withdrawable they need to select

### Before
before

### After
after


pezholio (approved):
LGTM 👍

PR #1511, opened 2024-02-22
(APS-383) Make appeal manager the arbitrator of an appeal

pezholio:
This builds on https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1518 to remove the `Reviewer` field from the appeals form, as well as return the name of the user who created the appeal when viewing the appeal.


edavey (approved):
👍 

see https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1518#discussion_r1499143532 for my related comment about the naming of `createdBy`.

PR #1513, opened 2024-02-22
(Chore) Run e2es in parallel

pezholio:
This follows the instructions in:

https://playwright.dev/docs/ci#circleci

To avoid stamping on user’s permissions, we also have a user for each shard, with the login details stored as with a prefix
(so `HMPPS_AUTH_USERNAME_1`, `HMPPS_AUTH_PASSWORD_1` etc etc), we then get the auth details for each shard in the setup process with a bit of Bash jiggery pokery and then run a selection of tests on each shard. This reduces the total runtime to around 3 minutes!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -182,6 +182,7 @@ jobs:
   e2e_environment_test:
     docker:
       - image: mcr.microsoft.com/playwright:v1.41.2-focal
+    parallelism: 4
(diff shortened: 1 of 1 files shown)


pezholio:
The tests are failing at the because the flash messages aren't appearing as expected. Possibly because the same user is logged in to multiple sessions.

richpjames (approved):

PR #1515, opened 2024-02-22
APS-317: Task Allocation region filter isnt sticking

vubalasu:
# Context

# Changes in this PR
Issue with the task allocation screen: it is defaulting back to home area when clicking on further pages or sorting in pre prod and the live service.  At the minute we have Midlands staff working in North East so this issue is preventing them from doing their work around task allocations.
### Before
image

### After
Screenshot 2024-02-22 at 15 39 09

### DIFF
--- a/server/controllers/tasksController.test.ts
+++ b/server/controllers/tasksController.test.ts
@@ -72,7 +72,7 @@ describe('TasksController', () => {
         hrefPrefix: paginationDetails.hrefPrefix,
         sortBy: 'createdAt',
         sortDirection: 'asc',
-        selectedArea: apArea.id,
+        apArea: apArea.id,
       })
 
       expect(taskService.getAll).toHaveBeenCalledWith({
@@ -114,7 +114,7 @@ describe('TasksController', () => {
         hrefPrefix: paginationDetails.hrefPrefix,
         sortBy: paramPaginationDetails.sortBy,
         sortDirection: paramPaginationDetails.sortDirection,
-        selectedArea: apAreaId,
+        apArea: apAreaId,
       })
       expect(getPaginationDetails).toHaveBeenCalledWith(unallocatedRequest, paths.tasks.index({}), {
         allocatedFilter: 'unallocated',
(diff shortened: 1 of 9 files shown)


server/controllers/tasksController.test.ts
```diff
@@ -72,7 +72,7 @@ describe('TasksController', () => {
         hrefPrefix: paginationDetails.hrefPrefix,
         sortBy: 'createdAt',
         sortDirection: 'asc',
-        selectedArea: apArea.id,
+        apArea: apArea.id,
```
yndajas, line 75: Is there a reason for changing from `selectedArea` to `apArea`?

server/views/tasks/_navigation.njk
```diff
@@ -1,19 +1,19 @@
...
+          href: (paths.tasks.index({})+ "?allocatedFilter=allocated&area="+apArea if apArea else paths.tasks.index({})+ "?allocatedFilter=allocated")
         },
         {
           text: 'Unallocated',
-          href: paths.tasks.index({}) + "?allocatedFilter=unallocated",
-          active: (activeTab === 'unallocated')
+          active: (activeTab === 'unallocated'),
+          href: (paths.tasks.index({})+ "?allocatedFilter=unallocated&area="+apArea if apArea else paths.tasks.index({})+ "?allocatedFilter=unallocated")
         }
       ]
```
yndajas, line 18: This is getting a little complicated - is there a suitable utils file this could be done in instead? It might be worth moving the whole `items` array generation into a utils method and then pass that to the template from the controller

server/views/tasks/_navigation.njk
```diff
@@ -1,19 +1,19 @@
...
   {{
     mojSubNavigation({
       label: 'Sub navigation',
       items: [
         {
           text: 'Allocated',
-          href: paths.tasks.index({})+ "?allocatedFilter=allocated",
-          active: (activeTab === 'allocated' or activeTab|length === 0)
+          active: (activeTab === 'allocated' or activeTab|length === 0),
+          href: (paths.tasks.index({})+ "?allocatedFilter=allocated&area="+apArea if apArea else paths.tasks.index({})+ "?allocatedFilter=allocated")
```
yndajas, line 11: Minor thing but might be good to retain the spaces around the `+`s if this stays in the template

yndajas (approved):
Looks like this does the job, but it might be worth a refactor to reduce the amount of logic implemented in the template (and allow proper unit testing) if you have the time

PR #1520, opened 2024-02-26
APS-314: Request a placement from Apply dashboard

vubalasu:
# Context

# Changes in this PR
Request a placement from Apply dashboard
### Before
image

### After
Screenshot 2024-02-26 at 08 05 28
Screenshot 2024-02-26 at 09 08 31

### DIFF
--- a/server/utils/applications/utils.test.ts
+++ b/server/utils/applications/utils.test.ts
@@ -21,6 +21,7 @@ import { DateFormats } from '../dateUtils'
 import { isApplicableTier, isFullPerson, nameOrPlaceholderCopy, tierBadge } from '../personUtils'
 
 import {
+  actionsCell,
   appealDecisionRadioItems,
   applicationStatusSelectOptions,
   applicationTableRows,
@@ -28,6 +29,7 @@ import {
   dashboardTableRows,
   eventTypeTranslations,
   firstPageOfApplicationJourney,
+  getAction,
   getApplicationType,
   getSections,
   getStatus,
@@ -35,7 +37,6 @@ import {
   lengthOfStayForUI,
   mapPlacementApplicationToSummaryCards,
   mapTimelineEventsForUi,
-  withdrawCell,
   withdrawnStatusTag,
 } from './utils'
 import { journeyTypeFromArtifact } from '../journeyTypeFromArtifact'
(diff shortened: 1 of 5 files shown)


server/utils/applications/utils.ts
```diff
@@ -131,8 +147,21 @@ const createNameAnchorElement = (
     : textValue(`LAO CRN: ${person.crn}`)
 }
 
-export const withdrawCell = (applicationId: string) => {
-  return htmlValue(linkTo(paths.applications.withdraw.new, { id: applicationId }, { text: 'Withdraw' }))
+export const actionsCell = (application: ApplicationSummary | Application) => {
+  const actionItems: Array<string> = []
+  const withDrawLink = linkTo(paths.applications.withdraw.new, { id: application.id }, { text: 'Withdraw' })
+  actionItems.push(withDrawLink)
```
yndajas, line 153: `withdrawLink` (casing)

server/utils/applications/utils.test.ts
```diff
@@ -603,10 +692,32 @@ describe('utils', () => {
...
+      actionItems.push(withDrawLink)
+      const requestForPlacementLink = linkTo(
+        placementApplicationPaths.placementApplications.create,
+        { id: application.id },
+        { text: 'Request for placement' },
+      )
+      actionItems.push(requestForPlacementLink)
+      const actionLinks = `<ul class="govuk-list">${actionItems.map(actionItem => `<li>${actionItem}</li>`).join('')}</ul>`
+      expect(actionsCell(application)).toEqual({
+        html: actionLinks,
```
yndajas, line 720: Sorry, 'granular' probably wasn't the right word in my earlier comment! The problem now is that this is essentially reimplementing the logic of the actual method in the test, and testing the logic/implementation rather than the outcome

For a utils method like this, it's probably best to use an explicit object literal in the expected value, rather than relying on the same logic to build the expected value and the output. Something like this for the two tests would avoid these issues and document the conditional expectations:

```ts
  describe('actionsCell', () => {
    it('returns a link to withdraw the application', () => {
      const application = applicationFactory.build({ id: 'an-application-id' })

      expect(actionsCell(application)).toEqual({
        html: '<ul class="govuk-list"><li><a href="/applications/an-application-id/withdrawals/new"  >Withdraw</a></li></ul>',
      })
    })

    it('returns a link to withdraw and request for placement of the application', () => {
      const application = applicationFactory.build({ id: 'an-application-id', status: 'awaitingPlacement' })

      expect(actionsCell(application)).toEqual({
        html: '<ul class="govuk-list"><li><a href="/applications/an-application-id/withdrawals/new"  >Withdraw</a></li><li><a href="/placement-applications"  >Request for placement</a></li></ul>',
      })
    })
  })
```

vubalasu: yah we already have tests for linkTo thats why just added for actions

yndajas (approved):

PR #1522, opened 2024-02-26
Withdrawals demo feedback 3

richpjames:
# Context
[Jira](https://dsdmoj.atlassian.net/jira/software/c/projects/APS/boards/1328?assignee=62a050ba9f5d480069c97f49&selectedIssue=APS-397) - this completes as many items on the ticket as possible at this time

# Changes in this PR
- update the placement applications GET request to include the `includeInitialRequestForPlacement` query param 
and: 
### Before
image

### After
image

### Before
before (1)

### After 
after (2)

### Before
image

### After 
image

### Before 
before-1
before-2

### After
after-1
after-2
before-3 (1)

### Before 
image

### After 
image


server/controllers/admin/placementRequests/withdrawalsController.ts
```diff
@@ -21,21 +23,31 @@ export default class WithdrawalsController {
...
+        if (!reason) {
           throw new ErrorWithData({ 'invalid-params': [{ propertyName: `$.reason`, errorType: 'empty' }] })
         }
 
         await this.placementRequestService.withdraw(req.user.token, req.params.id, req.body.reason)
 
-        req.flash('success', 'Placement request withdrawn successfully')
+        req.flash(
+          'success',
+          `Placement request for ${placementLength(Number(duration))} starting on ${DateFormats.isoDateToUIDate(expectedArrival, { format: 'short' })} withdrawn successfully`,
```
pezholio, line 49: Might be better to have this in a util to keep logic out of the controller, but this is fine for now

pezholio (approved):
LGTM - one tiny point about presentation logic in the controller, but no biggy

server/controllers/admin/placementRequests/withdrawalsController.ts
richpjames: Good idea, I'm working on doing the same thing for placement applications in a subsequent PR so will fix there

server/data/applicationClient.test.ts
```diff
@@ -454,28 +454,67 @@ describeClient('ApplicationClient', provider => {
...
+          },
+        })
+
+        await applicationClient.placementApplications(applicationId)
       })
+    })
+
+    describe('with NEW_WITHDRAWALS_FLOW_DISABLED turned off', () => {
+      it('calls the placement applications endpoint with the application ID and includeInitialRequestForPlacement=true', async () => {
+        process.env.NEW_WITHDRAWALS_FLOW_DISABLED = ''
```
yndajas, line 490: Not really for this PR, but is there a reason for using `'1'` and `''` instead of `true` and `false`?

yndajas (commented):
Looks like this has already been merged, but I was basically there with the review anyway (nothing significant!)

server/views/cancellations/new.njk
```diff
@@ -107,7 +107,7 @@
         {% endif %}
 
         {{ govukButton({
-            text: "Submit"
+            text: "Withdraw"
```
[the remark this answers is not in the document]
richpjames: Yeah that is the long term aim but not for now

PR #1530, opened 2024-02-27
Add translations to timeline URLs

pezholio:
Most of the time, the URL types translate in a human-readable fashion, but the newly-added `assessmentAppeal` type needs a bit of attention before it gets exposed to the user. This adds a translation method to `mapTimelineEventsForUi` to translate the url type to something human readable.

### Before

image

### After

image

### DIFF
--- a/server/utils/applications/utils.ts
+++ b/server/utils/applications/utils.ts
@@ -23,7 +23,9 @@ import type {
   PlacementType,
   SortDirection,
   TimelineEvent,
+  TimelineEventAssociatedUrl,
   TimelineEventType,
+  TimelineEventUrlType,
   User,
 } from '@approved-premises/api'
 import MaleAp from '../../form-pages/apply/reasons-for-placement/basic-information/maleAp'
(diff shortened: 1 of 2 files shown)


edavey (approved):
👍 thks - could you just tweak the commit msg to include the revised name of `mapTimelineUrlsForUi()` rather than `mapTimelineEventsForUi()` ?

server/utils/applications/utils.ts
```diff
@@ -271,6 +273,20 @@ const mapTimelineEventsForUi = (timelineEvents: Array<TimelineEvent>): Array<UiT
...
+
+const urlTypeForUi = (type: TimelineEventUrlType) => {
+  const translations: Record<TimelineEventUrlType, string> = {
+    application: 'application',
+    assessment: 'assessment',
+    booking: 'booking',
+    assessmentAppeal: 'appeal',
+  }
+  return translations[type]
+}
```
yndajas, line 288: Given this is only tested through `mapTimelineUrlsForUi`, does it need to be a function rather than a POJO (whether that POJO is inside or outside of the other function)?

PR #1531, opened 2024-02-27
APS-304:Investigate why placement requests are now appearing on Tasks…

vubalasu:
… screen

# Context

# Changes in this PR
Investigate why placement requests are now appearing on Tasks screen

### DIFF
--- a/server/controllers/match/placementRequestsController.test.ts
+++ b/server/controllers/match/placementRequestsController.test.ts
@@ -84,7 +84,7 @@ describe('PlacementRequestsController', () => {
       expect(getPaginationDetails).toHaveBeenCalledWith(request, paginationDetails.hrefPrefix)
       expect(taskService.getAll).toHaveBeenCalledWith({
         token,
-        taskType: 'PlacementApplication',
+        taskTypes: ['PlacementApplication'],
         page: 1,
         allocatedFilter: 'allocated',
         sortBy: 'createdAt',
--- a/server/data/taskClient.test.ts
+++ b/server/data/taskClient.test.ts
@@ -27,7 +27,7 @@ describeClient('taskClient', provider => {
 
       const apAreaId = 'ap-area-id'
       const userId = 'user-id'
-      const taskType: TaskType = 'PlacementRequest'
+      const taskTypes: Array<TaskType> = ['PlacementApplication', 'Assessment']
 
       provider.addInteraction({
         state: 'Server is healthy',
@@ -42,7 +42,7 @@ describeClient('taskClient', provider => {
             page: '1',
             sortDirection: 'asc',
             sortBy: 'createdAt',
-            taskType,
+            types: 'PlacementApplication,Assessment',
           },
           headers: {
             authorization: `Bearer ${token}`,
@@ -67,7 +67,7 @@ describeClient('taskClient', provider => {
         page: 1,
         sortDirection: 'asc',
         sortBy: 'createdAt',
-        taskType,
+        taskTypes,
       })
 
       expect(result).toEqual({
--- a/server/data/taskClient.ts
+++ b/server/data/taskClient.ts
@@ -27,7 +27,7 @@ export default class TaskClient {
     page,
     sortDirection,
     sortBy,
-    taskType,
+    taskTypes,
   }: {
     allocatedFilter: AllocatedFilter
     apAreaId: ApArea['id']
@@ -35,12 +35,16 @@ export default class TaskClient {
     page: number
     sortDirection: string
     sortBy: TaskSortField
-    taskType?: TaskType
+    taskTypes?: Array<TaskType>
   }): Promise<PaginatedResponse<Task>> {
+    const filters = {} as Record<string, string>
+    if (taskTypes.length) {
+      filters.types = taskTypes.join(',')
+    }
     return this.restClient.getPaginatedResponse({
       path: paths.tasks.index.pattern,
       page: page.toString(),
-      query: { allocatedFilter, allocatedToUserId, apAreaId, sortBy, sortDirection, taskType },
+      query: { ...filters, allocatedFilter, allocatedToUserId, apAreaId, sortBy, sortDirection },
     })
   }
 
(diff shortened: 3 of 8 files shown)


server/controllers/match/placementRequestsController.test.ts
```diff
@@ -84,7 +84,7 @@ describe('PlacementRequestsController', () => {
       expect(getPaginationDetails).toHaveBeenCalledWith(request, paginationDetails.hrefPrefix)
```
yndajas, line 84: Would be useful to update the first commit message so that the title describes the change rather than the ticket, and then in the body you can explain how this achieves that/what the cause of the issue was. Something like this:

```
Fix `taskType` query param

Placement requests were showing up on the tasks screens. We were
using `TaskType` as a query param rather than `Array<TaskType>`.
This fixes the typing of the query param and specifies the task types
we want to see on the relevant screens
```

server/data/taskClient.ts
```diff
@@ -35,12 +35,16 @@ export default class TaskClient {
...
+    const filters = {} as Record<string, string>
+    if (taskType.length) {
+      filters.types = taskType.join(',')
+    }
     return this.restClient.getPaginatedResponse({
       path: paths.tasks.index.pattern,
       page: page.toString(),
-      query: { allocatedFilter, allocatedToUserId, apAreaId, sortBy, sortDirection, taskType },
+      query: { ...filters, allocatedFilter, allocatedToUserId, apAreaId, sortBy, sortDirection },
     })
```
yndajas, line 48: Rather than adding a `filters` object that just contains one query param and not the rest, perhaps it would be cleaner to move the `query` object up and then conditionally add `taskType`?

```ts
    const query = { allocatedFilter, allocatedToUserId, apAreaId, sortBy, sortDirection }
    if (taskType.length) {
      query.types = taskType.join(',')
    }
    return this.restClient.getPaginatedResponse({
      path: paths.tasks.index.pattern,
      page: page.toString(),
      query,
    })
```

server/data/taskClient.test.ts
```diff
@@ -42,7 +42,7 @@ describeClient('taskClient', provider => {
             page: '1',
```
yndajas, line 42: Is the second commit something that was missed in 314, or is it taken directly from there? I'm a little confused about why the field name changes only in the Wiremock stub and not in the client itself or elsewhere?

yndajas (commented):
Generally makes sense but I'm a little confused by the second commit, and I have a couple of other suggestions that would be good to address

server/data/taskClient.ts
vubalasu: instead of amending the existing one better to create new one for immutability

yndajas: The `filters` variable is mutated in the current implementation, so I'm not clear what the immutability benefit is here?

server/data/taskClient.test.ts
yndajas: The second question still applies here after squashing the two commits

server/data/taskClient.ts
vubalasu: here it is new object   query: { ...filters, allocatedFilter, allocatedToUserId, apAreaId, sortBy, sortDirection } but here we are changing existing if (taskType.length) {
      query.types = taskType.join(',')
    }

yndajas: It's just defined the line above in `const query`, so it's isolated to this method

server/controllers/match/placementRequestsController.test.ts
yndajas: The commit message still needs revising ([see here for suggestion](https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/1531/files#r1504206883))

server/data/taskClient.ts
```diff
@@ -27,20 +27,24 @@ export default class TaskClient {
...
+    if (taskTypes.length) {
+      filters.types = taskTypes.join(',')
+    }
     return this.restClient.getPaginatedResponse({
       path: paths.tasks.index.pattern,
       page: page.toString(),
-      query: { allocatedFilter, allocatedToUserId, apAreaId, sortBy, sortDirection, taskType },
+      query: { ...filters, allocatedFilter, allocatedToUserId, apAreaId, sortBy, sortDirection },
     })
   }
```
yndajas, line 49: I think we suggested moving the query out and then spreading it below when we discussed [this suggestion](https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/1531#discussion_r1504217526) earlier, but I've just had a quick look at this locally and I think I see the typing issue you were coming up against. The code below is an alternative that avoids having a `filters` object containing only one of the multiple filters

```ts
    const query = {
      allocatedFilter,
      allocatedToUserId,
      apAreaId,
      sortBy,
      sortDirection,
      types: taskTypes.join(',') || undefined,
    }

    return this.restClient.getPaginatedResponse({
      path: paths.tasks.index.pattern,
      page: page.toString(),
      query,
    })
```

There are few other ways this could be done without introducing the partial `filters` object, but what do you think about this approach?

In any case, this is less important than the commit message, so if you're able to sort the commit message out I'm happy to leave this one as is for now

yndajas (commented):
Happy to approve this once the commit message is sorted out. Would be nice to sort out the filters/query thing too, but that's less important

server/data/taskClient.ts
vubalasu: with this approach we are sending undefined for types to the api and we need to remove again that

yndajas: Yeah I wondered is that was an issue 😅

This is probably the simplest change to improve the clarity of labels while avoiding sending `undefined` over then? Essentially moving all the filters into the `filters` object

```ts
    const filters = { allocatedFilter, allocatedToUserId, apAreaId } as Record<string, string>
    if (taskTypes.length) {
      filters.types = taskTypes.join(',')
    }
    return this.restClient.getPaginatedResponse({
      path: paths.tasks.index.pattern,
      page: page.toString(),
      query: { ...filters, sortBy, sortDirection },
    })
```

vubalasu: the existing one is a bit cleaner i suppose

yndajas: Though I think it would be better for all filters to go in the `filters` object rather than just `filtersThatNeedFurtherManipulation`/`justTypeFilters`, this is a minor thing so I'm happy to leave it for now

yndajas (approved):

PR #1532, opened 2024-02-27
(APS-358) Prepopulate single room requirement

yndajas:
# Context

[JIRA](https://dsdmoj.atlassian.net/browse/APS-358)

# Changes in this PR

Prepopulate the single room placement requirement on the matching information Assess screen based on the various fields from the application

This one is a little more complicated than the others, in that it looks at multiple fields and if any of them have a value of `'yes'`, the requirement is set to `'essential'`, otherwise `'notRelevant'`

I've taken this opportunity to introduce a `TaskListPageYesNoField` interface that holds information about the fields we're checking: the name of the field, the page it appears on, and optionally the value it should be set to and whether it's an optional field. This allows us to use iterators in a few places to keep the code DRY as the number of fields we check and mock increase

### Before

image

### After

image

### DIFF
--- a/server/form-pages/utils/defaultMatchingInformationValues.test.ts
+++ b/server/form-pages/utils/defaultMatchingInformationValues.test.ts
@@ -5,10 +5,12 @@ import {
 } from '../../utils/retrieveQuestionResponseFromFormArtifact'
 import { assessmentFactory } from '../../testutils/factories'
 import { MatchingInformationBody } from '../assess/matchingInformation/matchingInformationTask/matchingInformation'
-import { defaultMatchingInformationValues } from './defaultMatchingInformationValues'
+import { TaskListPageYesNoField, defaultMatchingInformationValues } from './defaultMatchingInformationValues'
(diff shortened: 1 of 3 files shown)


server/form-pages/utils/defaultMatchingInformationValues.test.ts
```diff
@@ -42,17 +44,23 @@ describe('defaultMatchingInformationValues', () => {
...
+      { name: 'riskToOthers', page: RoomSharing },
+      { name: 'riskToStaff', page: RoomSharing },
+      { name: 'sharingConcerns', page: RoomSharing },
+      { name: 'traumaConcerns', page: RoomSharing },
+    ]
+
+    yesNoFieldsToMock.forEach(({ name, page, value, optional }) =>
+      when(optional ? retrieveOptionalQuestionResponseFromFormArtifact : retrieveQuestionResponseFromFormArtifact)
+        .calledWith(assessment.application, page, name)
+        .mockReturnValue(value || 'yes'),
```
pezholio, line 62: I ❤️ this!

pezholio (approved):
This is really neat! Thanks! 👍

PR #1539, opened 2024-02-28
APS-314: Fix the link issue for Request a placement on Apply dashboard

vubalasu:
# Context

# Changes in this PR
Fix the link issue not opening on "Request a placement" from Apply dashboard

### DIFF
--- a/server/utils/applications/utils.test.ts
+++ b/server/utils/applications/utils.test.ts
@@ -706,7 +706,7 @@ describe('utils', () => {
       const application = applicationFactory.build({ id: 'an-application-id', status: 'awaitingPlacement' })
 
       expect(actionsCell(application)).toEqual({
-        html: '<ul class="govuk-list"><li><a href="/applications/an-application-id/withdrawals/new"  >Withdraw</a></li><li><a href="/placement-applications"  >Request for placement</a></li></ul>',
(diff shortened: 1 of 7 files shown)


server/utils/applications/utils.test.ts
```diff
@@ -706,7 +706,7 @@ describe('utils', () => {
       const application = applicationFactory.build({ id: 'an-application-id', status: 'awaitingPlacement' })
 
       expect(actionsCell(application)).toEqual({
-        html: '<ul class="govuk-list"><li><a href="/applications/an-application-id/withdrawals/new"  >Withdraw</a></li><li><a href="/placement-applications"  >Request for placement</a></li></ul>',
+        html: '<ul class="govuk-list"><li><a href="/applications/an-application-id/withdrawals/new"  >Withdraw</a></li><li><a href="/placement-applications?id=an-application-id"  >Request for placement</a></li></ul>',
```
richpjames, line 709: Given this test was passing before I think we'll need more confidence that the change has fixed the issue in question. Could you write an integration test for this functionality?   
IE:  
- Given I have an application in an 'awaiting placement' state 
- When I click 'Request for placement' 
- Then I am taken to the 'Create a request for placement' form

richpjames (approved):
thanks Bala

PR #1541, opened 2024-02-28
Fix/Only show the withdraw button to users who made the application

richpjames:
Previously we gave the option to all users viewing an application to withdraw it. However the API only allows the user who made the application to withdraw it so when the user reached the 'what would you like to withdraw?' screen they weren't seeing any options to choose. This change hides the button so they can't reach that screen


pezholio (approved):
LGTM 👍

PR #1548, opened 2024-02-28
Set flash messages on res.locals

pezholio:
While looking at https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/1513 I noticed some issues with Flash messages not appearing, I had a look and found https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/557 which seems to address the issues we're having.

This updates app.ts so that flash messages are set on res.locals rather than res.app.locals. This ensures their scope is a single request, rather than the application as a whole.


richpjames (approved):
Good spot!

PR #1554, opened 2024-02-29
APS-389: Update link to feedback survey from Beta banner

vubalasu:
# Context

# Changes in this PR
Update link to feedback survey from Beta banner
### Before
image

### After
image


richpjames (commented):
I don't think there's any need to retain `phaseBannerUtils` given all it does now is return one piece of markup. We might as well just add the markup to the view and remove it.

richpjames (approved):
Thanks Bala

PR #1555, opened 2024-02-29
Remove notes and date inputs on booking withdrawal screen

richpjames:
These are no longer required for any users

### Before
before

### After
after


pezholio (approved):
LGTM - do we need to make API changes for this?

richpjames:
@pezholio thanks. We've already removed the validation in the API for the date

PR #1559, opened 2024-03-01
Add 'other' reason description in to application withdrawal reasons

richpjames:
# Context

# Changes in this PR


integration_tests/tests/apply/withdrawApplication.cy.ts
```diff
@@ -43,7 +44,7 @@ context('Withdraw Application', () => {
     const withdrawConfirmationPage = Page.verifyOnPage(WithdrawApplicationPage)
 
     // When I choose a reason and click submit
-    withdrawConfirmationPage.completeForm(withdrawalReason)
+    withdrawConfirmationPage.completeForm(withdrawalReason, reasonDescription)
```
yndajas, line 47: Maybe not worth having two integration tests for one conditional input, but is it worth noting somewhere that this hardcodes in an assumption that the user is selecting `'other'`, and that the reason shouldn't be included if not?

server/controllers/apply/applications/withdrawalsController.ts
```diff
@@ -50,6 +54,14 @@ export default class WithdrawalsController {
     return async (req: Request, res: Response) => {
       try {
         const body: NewWithdrawal = { reason: req.body.reason, otherReason: req.body.otherReason }
+
+        if (body.reason === 'other' && !body.otherReason) {
```
yndajas, line 58: Is it worth adding a cheeky `.trim` in here to check that the user doesn't just enter a space (intentionally or by mistake)?

server/utils/applications/withdrawalReasons.ts
```diff
@@ -13,10 +13,9 @@ export type NewApplicationToBeSubmittedReasons = Extract<
 export type ApplicationProblemReasons = Extract<WithdrawalReason, (typeof applicationProblemReasons)[number]>
 export type OtherReasons = Extract<WithdrawalReason, (typeof placementNoLongerNeededReasons)[number]>
 
-export const withdrawalReasons: Record<WithdrawalReason, string> = {
+export const withdrawalReasons: Record<Exclude<WithdrawalReason, 'other'>, string> = {
   death: 'The person died',
   other_accommodation_identified: 'Other accommodation has been identified',
-  other: 'The placement is not needed for another reason',
```
yndajas, line 19: I don't think this needs excluding here, since excluding `'other'` from `placementNoLongerNeededReasons` should do the job? Leaving it in here and then taking the value from here in `withdrawalRadioOptions` might be more transparent?

server/utils/applications/withdrawalReasons.ts
```diff
@@ -35,9 +34,16 @@ export const placementNoLongerNeededOptions = filterByType<OtherReasons>(
...
+export const withdrawalRadioOptions = (conditionalValue: string) => [
   { divider: 'The application is no longer needed' },
   ...convertKeyValuePairToRadioItems(placementNoLongerNeededOptions, undefined),
+  {
+    text: 'The placement is not needed for another reason',
+    value: 'other',
+    conditional: {
+      html: conditionalValue,
+    },
+  },
```
yndajas, line 46: I find the variable names in this file a little confusing now they contain just two of the three withdrawal reasons

You could probably add an optional parameter to `convertKeyValuePairToRadioItems` along the lines of `conditionals?: Record<keyof T, RadioItem['conditional']>` and add it to the relevant item there if present?

Alternatively a quicker solution would be to add `WithoutOther` to the two variable names (`placementNoLongerNeededReasonsWithoutOther`, `placementNoLongerNeededOptionsWithoutOther`, and `withdrawalReasons` if the change there isn't reverted) here might help?

yndajas (approved):
A few suggestions for clarity and to avoid an empty value, but generally looks good to me

integration_tests/tests/apply/withdrawApplication.cy.ts
richpjames: Sounds like a job for type inference: 
`completeForm<T extends WithdrawalReason>(withdrawalReason: T, otherReason: T extends 'other' ? string : never) `

server/utils/applications/withdrawalReasons.ts
richpjames: correct

richpjames:
thanks @yndajas - all good suggestions

PR #1562, opened 2024-03-04
(APS-443) Add text to withdrawal request reason divider

yndajas:
# Context

[JIRA](https://dsdmoj.atlassian.net/browse/APS-443)

# Changes in this PR

- Hide capacity reasons on reason for placement request withdrawal form page for non-CRU users

### Before

image

image

#### CRU

image

image

#### Non-CRU

image

### DIFF
--- a/integration_tests/pages/match/placementApplicationWithdrawalConfirmationPage.ts
+++ b/integration_tests/pages/match/placementApplicationWithdrawalConfirmationPage.ts
@@ -1,9 +1,30 @@
 import { WithdrawPlacementRequestReason } from '../../../server/@types/shared/models/WithdrawPlacementRequestReason'
 import Page from '../page'
+import {
+  noCapacityReasons,
+  placementNoLongerNeededReasons,
+  problemInPlacementReasons,
+} from '../../../server/utils/applications/withdrawables/withdrawalReasons'
 
 export default class ConfirmationPage extends Page {
   constructor() {
     super('Why is this request for placement being withdrawn?')
+  }
+
+  shouldShowAllReasons() {
+    const reasons = [placementNoLongerNeededReasons, noCapacityReasons, problemInPlacementReasons].flat()
+
+    cy.get('.govuk-radios__input').each((inputElement, inputElementIndex) =>
+      cy.wrap(inputElement).should('have.value', reasons[inputElementIndex]),
+    )
+  }
+
+  shouldShowReasonsUnrelatedToCapacity() {
+    const reasons = [placementNoLongerNeededReasons, problemInPlacementReasons].flat()
+
+    cy.get('.govuk-radios__input').each((inputElement, inputElementIndex) =>
+      cy.wrap(inputElement).should('have.value', reasons[inputElementIndex]),
+    )
   }
 
   selectReason(withdrawalReason: WithdrawPlacementRequestReason) {
--- a/server/controllers/placementApplications/withdrawalsController.ts
+++ b/server/controllers/placementApplications/withdrawalsController.ts
@@ -5,7 +5,7 @@ import { PlacementApplicationService } from '../../services'
 import { catchValidationErrorOrPropogate, fetchErrorsAndUserInput } from '../../utils/validation'
 
 import placementApplicationPaths from '../../paths/placementApplications'
-import { applicationShowPageTab } from '../../utils/applications/utils'
+import { applicationShowPageTab, placementApplicationWithdrawalReasons } from '../../utils/applications/utils'
 import { WithdrawPlacementRequestReason } from '../../@types/shared/models/WithdrawPlacementRequestReason'
 import { Application } from '../../@types/shared'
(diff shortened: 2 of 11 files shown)


edavey:
👍 this looks correct, but as the ticket doesn't explain the user need I wonder if we ought to be going a bit further here?

Maybe I'm getting the wrong end of the stick but to me it looks like ticket 443 might be proposing a less than reliable solution to the story:

- So that an applicant doesn't provide an inapplicable reason for withdrawing a placement request
- As an applicant
- I want to only see PR withdrawal reasons which are applicable to applicants

server/utils/applications/withdrawables/withdrawalReasons.ts
```diff
@@ -54,7 +54,7 @@ export const placementApplicationWithdrawalReasons = (
   return [
     { divider: 'The placement is no longer needed' },
     ...convertKeyValuePairToRadioItems(placementNoLongerNeededOptions, selectedReason),
-    { divider: 'The placement is unavailable' },
+    { divider: 'The placement is unavailable (CRU use only)' },
```
yndajas: Not sure if we still want this in addition to the filtering

integration_tests/pages/match/placementApplicationWithdrawalConfirmationPage.ts
```diff
@@ -1,11 +1,27 @@
...
 
+  shouldShowReasons(userRoles: Array<ApprovedPremisesUserRole>) {
+    const reasons = userRoles.includes('workflow_manager')
+      ? [placementNoLongerNeededReasons, noCapacityReasons, problemInPlacementReasons].flat()
+      : [placementNoLongerNeededReasons, problemInPlacementReasons].flat()
+
+    cy.get('.govuk-radios__input').each((inputElement, inputElementIndex) =>
+      cy.wrap(inputElement).should('have.value', reasons[inputElementIndex]),
+    )
+  }
```
edavey, line 23: hmm, I wonder if we could be a bit clearer here? e.g.

1) be explicit in the name of this assertion that we are expecting the list of available reasons to vary depending on the user's role e.g. `shouldShowReasonsApplicableToRole()`, or

2) have two separate assertion functions named explicitly for the conditional logic in play e.g.

- `shouldShowReasonsAvailableToAllRoles()` and 
- `shouldShowReasonsAvailableToWorkflowManagerRoleOnly()`

What do you think / prefer?

server/controllers/placementApplications/withdrawalsController.ts
```diff
@@ -21,12 +24,16 @@ export default class WithdrawalsController {
         req.params.id,
       )
 
+      const user = await this.userService.getActingUser(req.user.token)
```
richpjames, line 27: We have this value in the session (`req.session.user`) so it would be better if we read from that

server/services/userService.test.ts
```diff
@@ -35,7 +35,7 @@ describe('User service', () => {
     referenceDataClientFactory.mockReturnValue(referenceDataClient)
   })
 
-  describe('getUser', () => {
+  describe('getActingUser', () => {
```
edavey, line 38: nice, thanks

server/controllers/placementApplications/withdrawalsController.ts
yndajas: Ah nice - will update!

edavey (approved):
Thanks, this looks very good. A couple of suggested tweaks:

- as you suggest, I agree that "CRU use only" in the heading is now redundant
- I think we could be a little clearer in how in our integration test we describe the two sets of reasons which we offer up based on the presence/absence of the `WORKFLOW_MANAGER` role

integration_tests/pages/match/placementApplicationWithdrawalConfirmationPage.ts
yndajas: So the difficulty here is that the radio options end up in a flat array and the CRU-specific ones are in the middle, hence the use of index-matching in the assertions. It might be worth extending the radio options generator to allow a test ID attribute for easier selection, which should allow us to assert on each of the reason types individually

For now though, I've split it into two assertion functions: one for all reasons and one for non-capacity-related reasons. These could be renamed to refer to the role instead of the reason type, but I opted for the latter because the index matching approach would prevent `shouldShowReasonsAvailableToAllRoles` from working for workflow managers, which might be a little confusing given it sounds as though should apply to all roles, and `nonWorkflowManager` feels awkward

What do you think?

Fixup commit showing the change: https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/1562/commits/e1e2237f4302c399c7d2681d7e99bb17ca50d201

yndajas: (Could just as easily go with `shouldShowReasonsApplicableToRole` though!)

PR #1564, opened 2024-03-04
(APS-417) Filter out unselected date variables

pezholio:
We’ve had issues where a user inputs a date the the relevant dates section and then unchecks a checkbox. If the date is invalid, we don’t validate when updating the data, so when we return the response, we get an error. 

We need to:

a) Ensure we only return dates included in the `selectedDates` array when rendering the responses for that page; and
b) Ensure that we filter out any date inputs that have not been explicitly selected when updating the page


yndajas (approved):

edavey (approved):
👍

PR #1567, opened 2024-03-05
(APS 452) Change notice type rules

pezholio:
During the pilot, in the interests of stabilising things, we treated all applications with less than 28 days as emergency and all applications with less than 6 months notice as short notice. Now the rules are:

- Less than or equal to 7 calendar days between application and start date - emergency application
- Less than or equal to 28 calendar days between application and start date - short notice
- More than 28 calendar days between application and start date - standard


edavey (approved):
Thanks

PR #1571, opened 2024-03-05
Fix some strict mode errors

yndajas:
# Context

Back in June 2023, Rich flagged that the [HMPPS TypeScript template](https://github.com/ministryofjustice/hmpps-template-typescript) doesn't have strict mode on by default. You can turn it on by adding `"strict": true` to the `compilerOptions` in `tsconfig.json`

Strict mode tightens TypeScript's checks which in turn minimises risk of incorrectness, allows us to make safer assumptions about the data we're passing around (particularly around nullish data), and can improve the usefulness of in-editor type hints

There's a `tsconfig.json` file in both the root of the project directory and within the `integration_tests` subdirectory. Back in June, Rich reported over 600 errors, which were presumably just those outside of `integration_tests` (the `integration_tests` checks don't run unless there are no errors in the standard checks). There are now 766 errors outside of `integration_tests`, so the problem hasn't grown a great deal

In Accredited Programmes, thanks to Rich's flagging we turned this on much earlier in the project lifecycle, so we managed to fix the errors in [one (substantial) PR](https://github.com/ministryofjustice/hmpps-accredited-programmes-ui/pull/62). However, strict mode compliant typing works fine with strict mode off, so this feels like something we can tackle gradually when time allows, with the aim of eventually cutting down the errors to zero so that strict mode can be turned on permanently

# Changes in this PR

Address a few errors highlighted by turning on strict mode in the main `tsconfig.json`. This brings the total errors down:

| scope | before | after|
| --- | --- | ---|
| outside integration tests | 766 | 698 |
| inside integration tests | 776 | 754 |

Strict mode remains off

### DIFF
--- a/server/applicationVersion.ts
+++ b/server/applicationVersion.ts
@@ -7,7 +7,7 @@ function getBuild() {
   try {
     // eslint-disable-next-line import/no-unresolved,global-require
     return require('../build-info.json')
-  } catch (ex) {
+  } catch (error) {
     return null
   }
 }
--- a/server/controllers/admin/deliusUserController.ts
+++ b/server/controllers/admin/deliusUserController.ts
@@ -1,4 +1,5 @@
 import type { Request, Response, TypedRequestHandler } from 'express'
+import { HttpError } from 'http-errors'
 import { UserService } from '../../services'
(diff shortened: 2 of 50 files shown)


pezholio (approved):
LGTM - thanks for doing this!

PR #1573, opened 2024-03-06
(APS-455) Return correct due date for assessments

pezholio:
This allows us to return the new `dueAt` attribute for `AssessmentSummary` objects when listing assessments using a new shared `daysUntilDueCell` function which we use for both `Task` and `AssessmentSummary` objects.

I've also tidied up some unused code


edavey (approved):
👍 expanding / sharing the use of the this `daysUntilDueCell` should help, thks!

PR #1575, opened 2024-03-06
APS-460: Fix filters not retained when switching between tabs showing…

vubalasu:
… placement request in various statuses

# Context

https://dsdmoj.atlassian.net/browse/APS-460 

# Changes in this PR
filters not retained when switching between tabs showing placement request in various statuses
## Screenshots of UI changes
No UI changes just to retain filter applied on different tabs


server/utils/placementRequests/utils.ts
```diff
@@ -52,3 +54,30 @@ export const requestTypes = [
 
 export const withdrawalMessage = (duration: number, expectedArrivalDate: string) =>
   `Request for placement for ${placementLength(Number(duration))} starting on ${DateFormats.isoDateToUIDate(expectedArrivalDate, { format: 'short' })} withdrawn successfully`
+
+export const placementRequestTabItems = (activeTab?: string, apArea?: string, requestType?: string): Array<TabItem> => {
+  const hrefSuffix = createQueryString({ apArea, requestType })
```
richpjames, line 59: we should check if the return value from `createQueryString` is defined before appending it to the URL. That way we avoid adding `undefined`

richpjames (approved):

PR #1579, opened 2024-03-07
Don’t request apAreaId on placement request list

pezholio:
If people are allocated a placement request that isn’t in their area, then they currently won’t see placement requests in that view because we’re filtering by the logged in user’s AP area.


edavey (approved):
👍 For now we want all assessors to be able to see all placement requests which have been allocated to themselves, regardless of AP Area.

I'm not sure why were constraining to AP area?

PR #1584, opened 2024-03-07
APS-457: Workflow manager views timeline whilst allocating task

vubalasu:
# Context

# Changes in this PR
Workflow manager views timeline whilst allocating task
### Before
Screenshot 2024-03-07 at 18 14 27

### After
image

### DIFF
--- a/server/utils/tasks/index.ts
+++ b/server/utils/tasks/index.ts
@@ -15,6 +15,7 @@ import {
   unallocatedTableRows,
 } from './listTable'
 import { userTableHeader, userTableRows } from './usersTable'
+import paths from '../../paths/apply'
 
 type GroupedTasks = {
   allocated: Array<Task>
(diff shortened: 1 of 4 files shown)


server/utils/tasks/index.ts
```diff
@@ -70,6 +71,15 @@ const taskSummary = (task: Task, application: Application): Array<SummaryListIte
       value: {
         text: getApplicationType(application),
       },
+      actions: {
+        items: [
+          {
+            href: `${paths.applications.show({ id: application.id })}?tab=timeline`,
+            text: 'View timeline',
+            visuallyHiddenText: 'timeline',
```
pezholio, line 79: I don't think this line is needed - a screenreader user would head "View timeline timeline" which doesn't seem right

pezholio (changes requested):

pezholio (approved):
LGTM 👍

PR #1594, opened 2024-03-11
APS-451 - Seperate area from the person details on Confirm Your Details screen

richpjames:
The area here relates to the case and not the user (if the case manager is different then the area should be the same). We want to make this clearer through the UI so we now show two different Summary lists: one for the case and one for the user

[JIRA](https://dsdmoj.atlassian.net/jira/software/c/projects/APS/boards/1328?assignee=62a050ba9f5d480069c97f49&selectedIssue=APS-451)

### Before
before

### After
after (1)


pezholio (approved):
Good plan 👍

PR #1598, opened 2024-03-12
Fix Identity bar tags

pezholio:
These were showing strangely before because they were out of line with the titles

### Before

image
image

### After

image
image


richpjames (approved):

richpjames:
Thanks for this Stu!

PR #1608, opened 2024-03-14
(APS-337) Filter tasks on task allocation by qualification, allocated user and CRN / Name

pezholio:
This adds filtering to tasks by various fields, already implemented in the API

### Before

image

image

### After

image

image


server/data/userClient.ts
```diff
@@ -27,6 +27,10 @@ export default class UserClient {
     return (await this.restClient.get({ path: paths.users.profile({}) })) as User
   }
 
+  async getUserList(): Promise<Array<User>> {
+    return this.restClient.get({ path: paths.users.index({}) }) as Promise<Array<User>>
+  }
+
```
edavey, line 33: hmm, I don't think we want the list of all the users (applicants included) do we? 

If we pass in a list of roles I think we can limit this list to staff members to whom tasks can be allocated? e.g.  `roles = assessor, matcher`, which might be a lot shorter and more relevant?

edavey (approved):
👍 Looks good: just one suggestion -- to retrieve a list of assessors and matchers only, as we won't need to find tasks allocated to anyone else?

server/data/userClient.ts
pezholio: Sorted! 👍

PR #1610, opened 2024-03-14
(APS-315) Add created column to application lisitings

pezholio:
This adds a "Created" column to the listing of a user's applications to make it easier to identify which application is which if there are multiple users

### Before

image

### After

image


davidatkinsuk (approved):
LGTM 👍  i'm guessing this will apply to all tabs?

PR #1611, opened 2024-03-14
(APS-470) Tidy up download reports list

pezholio:
This removes some reports we no longer need, as well as updates the copy and ordering of the remaining reports, as well as adding hint text.

### Before

image

### After

image

### DIFF
--- a/server/utils/reportUtils.ts
+++ b/server/utils/reportUtils.ts
@@ -1,17 +1,28 @@
-export const reportNames = {
-  'daily-metrics': 'Daily Metrics',
-  'placement-metrics': 'Placement Metrics',
-  'referrals-by-ap-type': 'Referrals by AP Type',
-  'referrals-by-tier': 'Referrals by Tier',
-  applications: 'Combined applications and placement requests',
-  referrals: 'Applications',
-  'lost-beds': 'Lost Beds',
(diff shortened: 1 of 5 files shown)


server/utils/reportUtils.ts
```diff
@@ -1,17 +1,28 @@
...
+  },
+  'daily-metrics': { text: 'Daily metrics', hint: 'Counts of key actions across the service grouped by day.' },
 } as const
 
 export type ReportType = (keyof typeof reportNames)[number]
 
-export const reportOptions = Object.entries(reportNames)
-  .sort((a, b) => (a[1] > b[1] ? 1 : -1))
-  .map(e => ({ value: e[0], text: e[1] }))
+export const reportOptions = Object.entries(reportNames).map(e => ({
```
richpjames, line 24: Might just be Monday morning but I struggled to grok this.  
I think it would be clearer if `reportNames` was renamed to (something like) `reportInputLabels`. 
In the `.map` here we could deconstruct the parameters and name them:  
```
.map(([reportName, reportLabel]),  => ({
  value: reportName,
  text: reportLabel.text,
  hint: { text: reportLabel.hint })) 
  ```

richpjames (approved):
I've left a non-blocking comment, feel free to merge

PR #1612, opened 2024-03-14
(Fix) Preserve filter options between tabs on Task view

pezholio:
The new filter options on the task view were not being preserved when clicking on the Allocated / Unallocated tabs, which was already happening for the `apArea`

This PR makes the `tasksTabItems` generic to handle all query strings using the `hrefPrefix`, which we already use for pagination. 

As it doesn't make sense to filter unallocated tasks by who a task has been allocated, we strip that query option out.

### DIFF
--- a/server/utils/tasks/listTable.ts
+++ b/server/utils/tasks/listTable.ts
@@ -152,21 +152,32 @@ export type TabItem = {
   href: string
 }
 
-const tasksTabItems = (activeTab?: string, apArea?: string): Array<TabItem> => {
-  const hrefSuffix = apArea ? `&area=${apArea}` : ''
-  const allocatedHref = `${paths.tasks.index({})}?allocatedFilter=allocated${hrefSuffix}`
-  const unallocatedHref = `${paths.tasks.index({})}?allocatedFilter=unallocated${hrefSuffix}`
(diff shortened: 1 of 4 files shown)


server/utils/tasks/listTable.ts
```diff
@@ -152,21 +152,32 @@ export type TabItem = {
...
+  const urlParams = new URLSearchParams(query)
+
+  const allocatedParams = {
+    ...Object.fromEntries(urlParams),
+    allocatedFilter: 'allocated',
+  } as Record<string, string>
+  const { allocatedToUserId: _, ...unallocatedParams } = {
+    ...Object.fromEntries(urlParams),
+    allocatedFilter: 'unallocated',
+  } as Record<string, string>
```
edavey, line 166: this assigns to a const named `unallocatedParams`? I'm guessing this is idiomatic TS but ...

edavey (approved):
Thks for fixing this issue and improving the area generally.

server/utils/tasks/listTable.ts
pezholio: Yeah, it removes `allocatedToUserId` and assigns the rest to a const called `unallocatedParams`. It's not the prettiest, but it's the least faffy option I've found

PR #1622, opened 2024-03-18
APS 312 - Add sorting to Placement Requests table

richpjames:
# Context

[JIRA](https://dsdmoj.atlassian.net/jira/software/c/projects/APS/boards/1328?selectedIssue=APS-312)

# Changes in this PR
This allows the Placment Application table to be sortable following the established pattern used for the other tables in this template.
Whilst users can only sort by name for now it makes it trivial to add other sortable columns once the API allows

### Before
before (4)

### After
after (5)


server/utils/placementApplications/table.test.ts
```diff
@@ -15,6 +15,37 @@ describe('table', () => {
...
+      expect(placementApplicationsTable(tasks, 'person', 'asc', '/')).toEqual({
+        firstCellIsHeader: true,
+        head: [
+          {
+            attributes: {
+              'aria-sort': 'ascending',
+              'data-cy-sort-field': 'person',
+            },
+            html: '<a href="/?undefined">Name</a>',
+          },
```
pezholio, line 30: Might be better replacing this with `sortHeader<TaskSortField>('Name', 'person', 'person', 'asc', '/'),` - not sure where that `/?undefined` is coming from?

richpjames: oops, yeah

richpjames: I've used the function directly instead of using the result of it

pezholio (approved):

PR #1623, opened 2024-03-18
(APS-511) Ask timeliness question for all applications made less than 6 months before arrival date

pezholio:
This adds a new `startDateOutsideOfNationalStandardsTimescales` helper which is separate from the `noticeTypeFromApplication` helper. Now if an application is made outside the six month timescale, but not short notice / emergency, the applicant should be prompted to justify the reason, and the assessor should be able to state if they agree with the justification

### DIFF
--- a/server/utils/assessments/suitabilityAssessmentAdjacentPage.test.ts
+++ b/server/utils/assessments/suitabilityAssessmentAdjacentPage.test.ts
@@ -1,245 +1,222 @@
 /* eslint-disable consistent-return */
 
-import { upper } from 'case'
+import { when } from 'jest-when'
 import { assessmentFactory } from '../../testutils/factories'
-import { noticeTypeFromApplication } from '../applications/noticeTypeFromApplication'
 import {
   shouldShowContingencyPlanPartnersPages,
   shouldShowContingencyPlanQuestionsPage,
 } from '../applications/shouldShowContingencyPlanPages'
 import { retrieveOptionalQuestionResponseFromFormArtifact } from '../retrieveQuestionResponseFromFormArtifact'
-import { suitabilityAssessmentAdjacentPage } from './suitabilityAssessmentAdjacentPage'
+import { SuitabilityAssessmentPageName, suitabilityAssessmentAdjacentPage } from './suitabilityAssessmentAdjacentPage'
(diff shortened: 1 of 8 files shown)


server/utils/assessments/suitabilityAssessmentAdjacentPage.test.ts
```diff
@@ -1,247 +1,229 @@
...
+import Rfap from '../../form-pages/apply/risk-and-need-factors/further-considerations/rfap'
+import SelectApType from '../../form-pages/apply/reasons-for-placement/type-of-ap/apType'
+import { ApprovedPremisesAssessment } from '../../@types/shared'
 
 jest.mock('../retrieveQuestionResponseFromFormArtifact')
 jest.mock('../applications/shouldShowContingencyPlanPages')
-jest.mock('../applications/noticeTypeFromApplication')
+jest.mock('../applications/startDateOutsideOfNationalStandardsTimescales')
+
+const mockApplicationOfType = (apType: string, assessment: ApprovedPremisesAssessment) => {
```
richpjames, line 20: it's probably worth giving `apType` a type of `'rfap' | 'esap' | 'pipe'`

server/utils/assessments/suitabilityAssessmentAdjacentPage.test.ts
```diff
@@ -1,247 +1,229 @@
...
+        mockApplicationOfType(apType, assessment)
+      })
+
+      it('should return the correct next page', () => {
+        expect(suitabilityAssessmentAdjacentPage(assessment, 'suitability-assessment')).toEqual(`${apType}-suitability`)
+      })
+    })
+  })
+
+  describe.each(['rfap', 'esap', 'pipe'])('with an %s application', apType => {
```
richpjames, line 64: If we cast the `['rfap', 'esap', 'pipe']` array `as const` then on line 65 we can type the `pageName` instead of casting it. It might event be useful to save the array to a const and share it between the tests that are using it.

richpjames (approved):
LGTM. I've left some minor nit picky comments about types, as usual 😝

server/utils/assessments/suitabilityAssessmentAdjacentPage.test.ts
pezholio: Good call! Have done that

PR #1641, opened 2024-03-21
APS-522: Reinstate text box to record rationale for decision in assess

vubalasu:
# Context

# Changes in this PR
Reinstate text area to record rationale for decision in assess 
### Before
Screenshot 2024-03-21 at 13 33 45

### After
Screenshot 2024-03-21 at 10 35 42

### DIFF
--- a/integration_tests/fixtures/assessmentData.json
+++ b/integration_tests/fixtures/assessmentData.json
@@ -39,7 +39,7 @@
       "outlineOfDiscussion": "Some discussion"
     }
   },
-  "make-a-decision": { "make-a-decision": { "decision": "accept" } },
+  "make-a-decision": { "make-a-decision": { "decision": "accept", "decisionRationale": "some rejected comment" } },
   "matching-information": {
     "matching-information": {
       "apType": "normal",
--- a/integration_tests/pages/assess/makeADecisionPage.ts
+++ b/integration_tests/pages/assess/makeADecisionPage.ts
@@ -10,4 +10,8 @@ export default class MakeADecisionPage extends AssessPage {
   completeForm() {
     this.checkRadioButtonFromPageBody('decision')
   }
+
+  enterDecisionRationale() {
+    this.completeTextInputFromPageBody('decisionRationale')
+  }
 }
--- a/server/form-pages/assess/makeADecision/makeADecisionTask/makeADecision.test.ts
+++ b/server/form-pages/assess/makeADecision/makeADecisionTask/makeADecision.test.ts
@@ -15,9 +15,11 @@ describe('MakeADecision', () => {
     it('should set the body', () => {
       const page = new MakeADecision({
         decision: 'accommodationNeedOnly',
+        decisionRationale: 'decisionRationale',
       })
       expect(page.body).toEqual({
         decision: 'accommodationNeedOnly',
+        decisionRationale: 'decisionRationale',
       })
     })
   })
@@ -44,6 +46,28 @@ describe('MakeADecision', () => {
         decision: 'You must select one option',
       })
     })
+    it('should have an error if the decision is rejected and a reason has not been provided', () => {
+      const page = new MakeADecision({
+        decision: 'informationNotProvided',
+      })
+
+      expect(page.errors()).toEqual({
+        decisionRationale: 'You must provide the rationale for your decision',
+      })
+    })
+  })
+
+  describe('no errors', () => {
+    it.each(['accept', 'withdrawnByPp'])(
+      'should not raise error if the decision is %s and a reason has not been provided',
+      async decision => {
+        const page = new MakeADecision({
+          decision,
+        })
+
+        expect(page.errors()).toEqual({})
+      },
+    )
   })
 
   describe('response', () => {
--- a/server/form-pages/assess/makeADecision/makeADecisionTask/makeADecision.ts
+++ b/server/form-pages/assess/makeADecision/makeADecisionTask/makeADecision.ts
@@ -7,7 +7,7 @@ import TasklistPage from '../../../tasklistPage'
 
 @Page({
   name: 'make-a-decision',
-  bodyProperties: ['decision'],
+  bodyProperties: ['decision', 'decisionRationale'],
 })
 export default class MakeADecision implements TasklistPage {
   name = 'make-a-decision'
@@ -43,6 +43,7 @@ export default class MakeADecision implements TasklistPage {
   constructor(
     public body: {
       decision: string
+      decisionRationale?: string
     },
   ) {}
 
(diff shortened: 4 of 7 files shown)


integration_tests/fixtures/assessmentData.json
```diff
@@ -39,7 +39,7 @@
       "outlineOfDiscussion": "Some discussion"
     }
   },
-  "make-a-decision": { "make-a-decision": { "decision": "accept" } },
+  "make-a-decision": { "make-a-decision": { "decision": "accept", "rejectReason": "some rejected comment" } },
```
yndajas, line 42: Given the decision here is "accept", I'd guess the reason should be `undefined` or an empty string?

server/form-pages/assess/makeADecision/makeADecisionTask/makeADecision.ts
```diff
@@ -65,6 +66,8 @@ export default class MakeADecision implements TasklistPage {
     const errors: TaskListErrors<this> = {}
 
     if (!this.body.decision) errors.decision = 'You must select one option'
+    if (this.body.decision && this.body.decision !== 'accept' && !this.body.rejectReason)
```
yndajas, line 69: I think this can be simplified slightly?
```suggestion
    if (this.body.decision !== 'accept' && !this.body.rejectReason)
```

server/form-pages/assess/makeADecision/makeADecisionTask/makeADecision.ts
```diff
@@ -65,6 +66,8 @@ export default class MakeADecision implements TasklistPage {
...
+      errors.rejectReason = 'You must enter reject reason'
```
yndajas, line 70:
```suggestion
      errors.rejectReason = 'You must enter a reason for rejection'
```

server/views/assessments/pages/make-a-decision/make-a-decision.njk
```diff
@@ -37,4 +37,18 @@
...
       {%endcall%}
     </div>
+    {{
+        formPageTextarea(
+          {
+            fieldName: "rejectReason",
+            type: "textarea",
+            spellcheck: false,
+            label: {
+              text: "Reject reason",
```
yndajas, line 47:
```suggestion
              text: "Reason for rejection",
```

server/form-pages/assess/makeADecision/makeADecisionTask/makeADecision.test.ts
```diff
@@ -44,6 +46,15 @@ describe('MakeADecision', () => {
         decision: 'You must select one option',
       })
     })
+    it('should have an error if the decision is not accepted and not provided reject reason', () => {
```
yndajas, line 49:
```suggestion
    it('should have an error if the decision is not "accepted" and a reason has not been provided', () => {
```

yndajas (commented):
Generally looks good - just a few very minor wording suggestions and I wonder if some of the test data needs clarifying?

server/form-pages/assess/makeADecision/makeADecisionTask/makeADecision.ts
vubalasu: if i say this.body.decision !== 'accept' then it will give error for no decision

integration_tests/fixtures/assessmentData.json
vubalasu: this is the testdata being used for all test cases accept and reject

server/form-pages/assess/makeADecision/makeADecisionTask/makeADecision.ts
edavey: Apologies for the late contribution but how about "You must provide the rationale for your decision"?

server/views/assessments/pages/make-a-decision/make-a-decision.njk
edavey: or perhaps "Rationale for your decision"

integration_tests/fixtures/assessmentData.json
edavey: does it help if we name the "rationale" property in a decision-agnostic way e.g.

```
"make-a-decision": { "make-a-decision": { "decision": "accept", "decisionRationale": "some rationale for the decision" } 
```

server/form-pages/assess/makeADecision/makeADecisionTask/makeADecision.ts
yndajas: I'm not sure I understand: there should be an error if there's no decision based on the line above this one. As far as I can tell, this can follow the same pattern as other pages like [type of accommodation](https://github.com/ministryofjustice/hmpps-approved-premises-ui/blob/main/server/form-pages/apply/move-on/typeOfAccommodation.ts#L58-L64), [plans in place](https://github.com/ministryofjustice/hmpps-approved-premises-ui/blob/main/server/form-pages/apply/move-on/plansInPlace.ts#L80-L92), or [placement duration](https://github.com/ministryofjustice/hmpps-approved-premises-ui/blob/main/server/form-pages/apply/move-on/placementDuration.ts#L71-L87) where the `undefined` error is handled first, then you don't need to check the value exists for the second error

integration_tests/fixtures/assessmentData.json
yndajas: That probably makes more sense given you can still provide a rationale when accepting a placement

server/form-pages/assess/makeADecision/makeADecisionTask/makeADecision.ts
vubalasu: the rejectReason is only needed for rejection if you accept that is optional
so if we submit without selecting any option then you will see 2 error
1)You must select one option
2)You must enter a reason for rejection
since this.body.decision undefined and this.body.decision !== 'accept' becomes true and raise both errors
which i dont want to show thats why added this.body.decision check in the if block

server/form-pages/assess/makeADecision/makeADecisionTask/makeADecision.ts
```diff
@@ -65,6 +66,8 @@ export default class MakeADecision implements TasklistPage {
     const errors: TaskListErrors<this> = {}
 
     if (!this.body.decision) errors.decision = 'You must select one option'
+    if (this.body.decision && this.body.decision !== 'accept' && !this.body.decisionRationale)
```
richpjames, line 69: I think we should add a case for `this.body.decision !== 'withdrawnByPp'` too

integration_tests/pages/assess/makeADecisionPage.ts
```diff
@@ -10,4 +10,8 @@ export default class MakeADecisionPage extends AssessPage {
   completeForm() {
     this.checkRadioButtonFromPageBody('decision')
   }
+
+  enterDecisionRationale(reason: string) {
```
richpjames, line 14: I think it would be best if we used `this.completeTextInputFromPageBody('decisionRationale')` here to follow the format that the other `AssessPage`s use

richpjames (approved):
LGTM, thanks for addressing my comments

PR #1645, opened 2024-03-22
Add missing args to CircleCI config

pezholio:
We were missing `pipeline_id` and `pipeline_number` args, which are needed for the jira integration to add build status to tickets


Harriethw (approved):
cool, we'll introduce this to CAS2 as well 👍 💯

PR #1647, opened 2024-03-22
(APS-533) Add Jira update to prod and preprod pipelines

pezholio:
This ensures we update a Jira ticket's status when we push to prod and preprod too


davidatkinsuk (approved):
LGTM 👍

PR #1649, opened 2024-03-25
APS-512 - Show both the applicant and the case manager details on the assess screen

richpjames:
# Context 

We need to the user the details for both the case manager and the applicant. Often this is the same person so we only show one set of details and indicate in the heading that they're the same person. Otherwise we show both sets of details. As the backfill has now been completed in the API to fill the 'caseManagerIsNotApplicant', 'applicantUserDetails' and 'caseManagerUserDetails' we can remove the code responsible for fetching this information from the application itself and just use the first class properties instead. 

[JIRA](https://dsdmoj.atlassian.net/jira/software/c/projects/APS/boards/1328?assignee=62a050ba9f5d480069c97f49&selectedIssue=APS-512) 

# Changes in this PR

### Before
before

#### When the case manager and applicant are different people
after (1)
#### When the case manager and applicant are the same person
after

### DIFF
--- a/integration_tests/pages/assess/clarificationNoteConfirmPage.ts
+++ b/integration_tests/pages/assess/clarificationNoteConfirmPage.ts
@@ -12,25 +12,32 @@ export default class SufficientInformationPage extends Page {
   }
 
   confirmUserDetails(application: ApprovedPremisesApplication) {
-    cy.get('dl').within(() => {
-      this.assertDefinition(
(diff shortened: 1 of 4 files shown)


pezholio (approved):
Looking good - the handling of each type of user is much tidier now too!

richpjames:
@pezholio yep, reads a lot easier now

PR #1651, opened 2024-03-25
(APS-535) Investigate Flipt for feature flags

pezholio:
This adds an ADR / config for the use of [Flipt](https://www.flipt.io) for easy use of feature flags. This is already used by the integrations team in production, and all the necessary infrastructure is in place. 

Assuming we decide to go with this, I have another PR waiting in the wings as a proof of concept.

### DIFF
--- a/doc/architecture/decisions/0012-handling-feature-flags-with-flipt.md
+++ b/doc/architecture/decisions/0012-handling-feature-flags-with-flipt.md
@@ -1,0 +1,59 @@
+# 12. Handling feature flags with Flipt
+
+Date: 2024-03-25
+
+## Status
+
+Accepted
+
+## Context
+
+As the service gets used by more people, and communicating new features / changes with
+users becomes more essential, it's important that sometimes we hide certain features
+behind a feature flag to ensure we can roll out features in a controlled manner, whilst
+not slowing down the development process.
+
+Previously we've handled feature flags in an adhoc way, by use of environment variables,
+but this causes a couple of issues, most importantly, to roll back or introduce a feature
+we need to start a new deploy, which can take time, as we need to go through the whole
+deployment process, from PR, through to E2E tests and deployment to various different
+environments. We also need to have a lot of code that handles fetching these variables,
+adding conditionals etc, which gives us more code to maintain.
+
+## Decision
+
+The integrations team have recently deployed an instance of [Flipt](https://www.flipt.io/),
+which is a simple service that allows us to manage feature flags and turn them on and off
+at will via a UI, without having to redeploy the service.
+
+There are instances for each environment, Dev, Prod and Prepod, meaning we can have
+different values for feature flags in each environment.
+
+To manage feature flags, we will add a new Feature Flag service, which (initially at least)
+will accept the name of a feature flag and return a boolean value indicating whether the
+feature flag is enabled or not.
+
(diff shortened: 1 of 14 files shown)


richpjames:
I've used a similar FS solution in the past and it was really useful so I think it would be good to use one here. 
How do you plan that we host Flipt (assuming we're not using the managed version)? could you link me to where the integration team have it setup?

server/services/featureFlagService.ts
```diff
@@ -0,0 +1,42 @@
...
+    if (config.fliptEnabled) {
+      this.fliptClient = new FliptClient({
+        url: config.apis.flipt.url,
+        authenticationStrategy: new ClientTokenAuthentication(config.apis.flipt.token),
+      })
+      this.namespaceKey = config.apis.flipt.namespace
+    }
+  }
+
+  async getBooleanFlag(flag: string): Promise<boolean> {
```
richpjames, line 20: When we have flags in the service we will want to make sure this argument is strongly typed

doc/architecture/decisions/0012-handling-feature-flags-with-flipt.md
```diff
@@ -0,0 +1,59 @@
...
+adding conditionals etc, which gives us more code to maintain.
+
+## Decision
+
+The integrations team have recently deployed an instance of [Flipt](https://www.flipt.io/),
+which is a simple service that allows us to manage feature flags and turn them on and off
+at will via a UI, without having to redeploy the service.
+
+There are instances for each environment, Dev, Prod and Prepod, meaning we can have
+different values for feature flags in each environment.
```
edavey, line 30: will we be able to expand this to our 4th environment `test` which we may be repurposing on the UI side as a `qa`?

helm_deploy/values-test.yaml
```diff
@@ -19,7 +19,9 @@ generic-service:
     COMMUNITY_ACCOMMODATION_API_SERVICE_NAME: approved-premises
     COMMUNITY_ACCOMMODATION_API_TIMEOUT_RESPONSE: 30000
     COMMUNITY_ACCOMMODATION_API_TIMEOUT_DEADLINE: 30000
-
+    FLIPT_ENABLED: true
+    FLIPT_URL: 'https://feature-flags-dev.hmpps.service.justice.gov.uk/'
```
edavey, line 23: so for now our `test` and `dev` environments will always have their flags toggled in unison?

doc/how-to/add-a-feature-flag.md
```diff
@@ -0,0 +1,65 @@
...
+
+You can then call the `getBooleanFlag` method to return a boolean value depending on if
+the flag is enabled or not, e.g (where `FLAG_NAME` is a string specifying the key of the
+flag you created in step 1):
+
+```typescript
+const flagEnabled = await this.featureFlagService.getBooleanFlag(FLAG_NAME)
+```
+
+You can then pass that boolean value to wherever it is required in your code to.
```
edavey, line 52: I don't think you want/need that final "to"

edavey (approved):
Very thorough, thanks for the clear documentation and developer guidance

doc/architecture/decisions/0012-handling-feature-flags-with-flipt.md
pezholio: Possibly, but we'd have to deploy another instance of the Flipt infrastructure

PR #1657, opened 2024-03-26
(APS-546) Fix weeks and days being displayed incorrectly

pezholio:
This fixes days and weeks being incorrectly flipped when playing back a placement request response for an additional request for placement.

## Before

image

## After

image


davidatkinsuk (approved):
LGTM 👍

PR #1660, opened 2024-03-26
(APS-334) Show requested and actual arrival dates

pezholio:
There was some degree of confusion on the CRU dashboard where users assumed the “Arrival Date” against match requests would change when a booking was made. This updates the table to show both the requested aand actual arrival dates to avoid confusion.

This is behind a new `show-both-arrival-dates` feature flag . Ordinarily this wouldn’t be behind a feature flag, but this is just a proof of concept

## Before

image

## After

image

### DIFF
--- a/server/services/featureFlagService.ts
+++ b/server/services/featureFlagService.ts
@@ -2,6 +2,8 @@ import { ClientTokenAuthentication, FliptClient } from '@flipt-io/flipt'
 import config from '../config'
 import logger from '../../logger'
 
+type FeatureFlag = 'show-both-arrival-dates'
+
 export default class FeatureFlagService {
   fliptClient: FliptClient | null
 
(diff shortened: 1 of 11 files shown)


server/services/featureFlagService.ts
```diff
@@ -17,7 +19,7 @@ export default class FeatureFlagService {
     }
   }
 
-  async getBooleanFlag(flag: string): Promise<boolean> {
+  async getBooleanFlag(flag: FeatureFlag): Promise<boolean> {
```
richpjames, line 22: 👏

server/utils/placementRequests/table.ts
```diff
@@ -35,7 +35,22 @@ export const tableRows = (tasks: Array<PlacementRequestTask>): Array<TableRow> =
 export const dashboardTableRows = (
   placementRequests: Array<PlacementRequest>,
   status: PlacementRequestStatus | undefined,
+  showRequestedAndActualArrivalDates: boolean = false,
```
richpjames, line 38: could we have this third param as an object? 
I know it should be short lived but I find it a lot clearer to read in the consuming code than just passing a boolean

richpjames (approved):

PR #1661, opened 2024-03-26
(APS-550) Add other reason to Application Timeliness page

pezholio:
If the reason ID is `other` then we fetch the “other” reason from the ReasonForShortNotice page and return that

## Before

image

## After 

image

### DIFF
--- a/server/form-pages/assess/assessApplication/suitablityAssessment/applicationTimeliness.ts
+++ b/server/form-pages/assess/assessApplication/suitablityAssessment/applicationTimeliness.ts
@@ -64,9 +64,13 @@ export default class ApplicationTimeliness implements TasklistPage {
     )
     const arrivalDate = arrivalDateFromApplication(this.assessment.application)
 
-    const lateApplicationReason = lateApplicationReasonId
-      ? shortNoticeReasons[lateApplicationReasonId]
-      : 'None supplied'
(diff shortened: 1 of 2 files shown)


server/form-pages/assess/assessApplication/suitablityAssessment/applicationTimeliness.ts
```diff
@@ -64,9 +64,17 @@ export default class ApplicationTimeliness implements TasklistPage {
     )
     const arrivalDate = arrivalDateFromApplication(this.assessment.application)
 
-    const lateApplicationReason = lateApplicationReasonId
-      ? shortNoticeReasons[lateApplicationReasonId]
-      : 'None supplied'
+    let lateApplicationReason: string
+
+    if (lateApplicationReasonId === 'other') {
+      lateApplicationReason = retrieveOptionalQuestionResponseFromFormArtifact(
```
richpjames, line 70: Is it worth providing a fallback here for backwards compatibility?

richpjames (approved):

server/form-pages/assess/assessApplication/suitablityAssessment/applicationTimeliness.ts
pezholio: Good call 👍

PR #1668, opened 2024-03-27
(APS-564) Use `document` when rendering submitted applications

pezholio:
When viewing applications in the read only version, or when assessing, we still build up the response using the `data` response and the form classes. This causes issues if the schema changes, as we will get errors. This was one of the reasons why we have a persisted `document` field in an application, but this wasn’t getting used for some reason

This adds a new `SumbmittedApplicationSummaryCards` class to get all the sections for a form (unfortunately this is not yet persisted) and walk through each task and page and return each question in each page as a summary list.

### DIFF
--- a/server/utils/applications/submittedApplicationSummaryCards.ts
+++ b/server/utils/applications/submittedApplicationSummaryCards.ts
@@ -1,0 +1,77 @@
+import { ApprovedPremisesApplication as Application } from '../../@types/shared'
+import { FormSections, SummaryListItem, UiTask } from '../../@types/ui'
+import Apply from '../../form-pages/apply'
+import { documentsFromApplication } from '../assessments/documentUtils'
+import { getActionsForTaskId } from '../assessments/getActionsForTaskId'
+import { linebreaksToParagraphs } from '../utils'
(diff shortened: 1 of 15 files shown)


server/utils/applications/submittedApplicationSummaryCards.ts
```diff
@@ -0,0 +1,77 @@
...
+
+type PageResponse = Record<string, QuestionResponse>
+
+type TaskResponse = Array<PageResponse>
+
+export class SumbmittedApplicationSummaryCards {
+  constructor(
+    private readonly application: Application,
+    private readonly assessmentId?: string,
+    private readonly sections: FormSections = Apply.sections.slice(0, -1), // Defaults to all Apply sections except the final "Check your answers" section
```
richpjames, line 19: Not for now but I think we're doing `Apply.sections.slice(0, -1)` in a couple places, we could save it as a `const` and use that instead

richpjames (approved):
LGTM, thanks for doing this

richpjames:
Merging this in Stu's absence to unblock us

PR #1669, opened 2024-03-27
Make a decision page to show previously selected value

richpjames:
We weren't showing the value the user previously selected when the page was rendered even though it was saved
[Jira](https://dsdmoj.atlassian.net/jira/software/c/projects/APS/boards/1328?assignee=62a050ba9f5d480069c97f49&selectedIssue=APS-522)


edavey (approved):
👍 nasty UX bug!

PR #1670, opened 2024-03-27
APS 557 - RFAP details page to AP type task

richpjames:
# Context
[JIRA](https://dsdmoj.atlassian.net/jira/software/c/projects/APS/boards/1328?assignee=62a050ba9f5d480069c97f49&selectedIssue=APS-557)

# Changes in this PR


edavey (approved):
It would be good to understand more clearly which in-flight applications will be affected by this change -- and how we might handle those scenarios gracefully...

Is it the scenario where I've supplied the RFAP details within the 'further considerations' section of my `data` blob?

Are you planning to explore this further once the code has been promoted to preprod?

richpjames:
@edavey I've looked through the places where I'd expect the code to break when there are changes to the application and I think it will work ok given the changes here as it uses the `sections` defined [here](https://github.com/ministryofjustice/hmpps-approved-premises-ui/blob/991855b44bda34d49503ab84b07177e873920ec5/server/form-pages/apply/index.ts#L11) to calculate which sections are complete.  There however could be some unexpected issues once it reaches preprod.
My main concern is that if a user starts their application *before* this change is deployed: 
1. They enter that they want a 'standard'/'normal' AP 
2. They reach the 'Further considerations' section
3. They request an RFAP and explain why 
4. This change is deployed which removes the RFAP pages they have completed from the 'Further Considerations' section
5. *Then* they submit the application. In this case I think their response to the RFAP questions will be lost and this might not reach the assessor.  

I'm not too sure how best to handle this. We could communicate when this change is deployed to prod that any users requiring an RFAP will have to select this from the AP Type page... Not exactly graceful.

PR #1671, opened 2024-03-28
Handle withdrawn unsubmitted applications

richpjames:
#1668 switched to using the `application.document` for showing applications in the read only view. This won't work for unsubmitted withdrawn applications as they don't have the 'document' property. For now we revert to the legacy way of showing applications if they are withdrawn and unsubmitted. In future we will want to address this in a more changeproof manner but this unblocks us for now.

[JIRA](https://dsdmoj.atlassian.net/browse/APS-564)


froddd (approved):
LGTM

PR #1682, opened 2024-04-04
Temporarily remove MHAPs

richpjames:
We don't want to deploy the MHAP work yet but we need to unblock the pipeline.
In order to do this we remove them from the 'apTypes' array which will prevent them being shown on the AP type page in Apply and on the Matching Information page in Assess

### Before
before
before (1)

### After
after
after (1)

### DIFF
--- a/integration_tests/pages/assess/matchingInformationPage.ts
+++ b/integration_tests/pages/assess/matchingInformationPage.ts
@@ -7,6 +7,11 @@ export default class MatchingInformationPage extends AssessPage {
     super('Matching information', assessment, 'matching-information', 'matching-information', '')
   }
 
+  shouldNotShowMentalHealthAps() {
+    cy.get(`input[name="apType"][value="isMHAPElliottHouse"]`).should('not.exist')
+    cy.get(`input[name="apType"][value="isMHAPStJosephs"]`).should('not.exist')
+  }
+
   completeForm() {
     this.checkRadioButtonFromPageBody('apType')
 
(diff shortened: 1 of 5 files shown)


integration_tests/pages/assess/matchingInformationPage.ts
```diff
@@ -7,6 +7,11 @@ export default class MatchingInformationPage extends AssessPage {
     super('Matching information', assessment, 'matching-information', 'matching-information', '')
   }
 
+  shouldNotShowMentalHealthAps() {
+    cy.get(`input[name="apType"][value="mhapElliottHouse"]`).should('not.exist')
+    cy.get(`input[name="apType"][value="mhapStJosephs"]`).should('not.exist')
```
yndajas, line 12: Should this be?

```suggestion
    cy.get(`input[name="apType"][value="isMHAPElliottHouse"]`).should('not.exist')
    cy.get(`input[name="apType"][value="isMHAPStJosephs"]`).should('not.exist')
```

I think `placementCriteriaUtils` might need updating for this?

richpjames: Good spot with the integration tests.

 I believe `placementCriteriaUtils` is driven by the same apTypes we use in the `apType` page

yndajas: Ah so it is!

richpjames: I changed it whilst you were on leave  👍

yndajas (approved):

PR #1686, opened 2024-04-04
APS-553: Remove short notice application title

vubalasu:
# Context

# Changes in this PR
Remove short notice application title
### Before
image

### After
image

### DIFF
--- a/server/form-pages/apply/reasons-for-placement/basic-information/reasonForShortNotice.test.ts
+++ b/server/form-pages/apply/reasons-for-placement/basic-information/reasonForShortNotice.test.ts
@@ -28,6 +28,14 @@ describe('ReasonForShortNotice', () => {
     expect(page.question).toEqual('Why is this application being submitted outside of National Standards timescales?')
   })
 
+  it('should return the question for standard applications and not the title', () => {
+    ;(noticeTypeFromApplication as jest.Mock).mockReturnValue('standard')
+    const page = new ReasonForShortNotice({}, application)
+
+    expect(page.title).toBeUndefined()
+    expect(page.question).toEqual('Why is this application being submitted outside of National Standards timescales?')
+  })
+
   it('should set the body', () => {
     const page = new ReasonForShortNotice(
       {
--- a/server/form-pages/apply/reasons-for-placement/basic-information/reasonForShortNotice.ts
+++ b/server/form-pages/apply/reasons-for-placement/basic-information/reasonForShortNotice.ts
@@ -31,11 +31,14 @@ export default class ReasonForShortNotice implements TasklistPage {
     public readonly body: Partial<ReasonForShortNoticeBody>,
     readonly application: Application,
   ) {
-    if (noticeTypeFromApplication(application) === 'emergency') {
+    const noticeType = noticeTypeFromApplication(application)
+    if (noticeType === 'emergency') {
       this.title = 'Emergency application'
       this.question = 'What was the reason for submitting this application 7 days or less before the AP is needed?'
-    } else {
+    } else if (noticeType === 'shortNotice') {
       this.title = 'Short notice application'
+      this.question = 'Why is this application being submitted outside of National Standards timescales?'
+    } else {
       this.question = 'Why is this application being submitted outside of National Standards timescales?'
     }
   }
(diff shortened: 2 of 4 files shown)


server/form-pages/apply/reasons-for-placement/basic-information/reasonForShortNotice.ts
```diff
@@ -35,7 +35,6 @@ export default class ReasonForShortNotice implements TasklistPage {
       this.title = 'Emergency application'
       this.question = 'What was the reason for submitting this application 7 days or less before the AP is needed?'
     } else {
-      this.title = 'Short notice application'
```
yndajas, line 38: Given the template is calling `this.title` in a `h1`, will this change lead to an empty `h1` for short notice applications?

You should be able to get round this by making a couple of changes in the template:
1. wrap the `h1` in a conditional block to check if there's a page title
2. add `isPageHeading: !this.title` (not 100% sure if that's Nunjucks-compliant logic) to the `legend`

server/form-pages/apply/reasons-for-placement/basic-information/reasonForShortNotice.ts
```diff
@@ -35,7 +35,6 @@ export default class ReasonForShortNotice implements TasklistPage {
       this.title = 'Emergency application'
```
yndajas, line 35: Is it worth confirming with Dan that we definitely want a separate title and question for emergency applications but not short notice ones? The divergent behaviour for two types of non-standard application might be unexpected?

server/views/applications/pages/basic-information/reason-for-short-notice.njk
```diff
@@ -23,7 +23,7 @@
         fieldset: {
           legend: {
             text: page.question,
-            classes: "govuk-fieldset__legend--m"
+            classes: "govuk-fieldset__legend--l"
```
yndajas, line 26: I think we only want this to be large when there isn't a page title (i.e. for short notice applications but not emergency applications if the ticket is correct), unless the emergency applications title also needs removing (in which case the ticket needs revising)? Could add a little ternary logic here to check if there's a page title and set it to `l` or `m` accordingly?

server/form-pages/apply/reasons-for-placement/basic-information/reasonForShortNotice.ts
richpjames: Did Dan confirm that there needs to be a seperate title and question for emergency applications but not for short notice applications @vubalasu?

server/form-pages/apply/reasons-for-placement/basic-information/reasonForShortNotice.test.ts
```diff
@@ -28,6 +28,14 @@ describe('ReasonForShortNotice', () => {
     expect(page.question).toEqual('Why is this application being submitted outside of National Standards timescales?')
   })
 
+  it('should return the question for standard applications and not the title', () => {
```
richpjames, line 31: Users with a standard application won't see this page. 
We need to remove the title for **short notice** applications

richpjames (changes requested):

server/form-pages/apply/reasons-for-placement/basic-information/reasonForShortNotice.ts
vubalasu: yes i had a chat with him and confirmed

server/form-pages/apply/reasons-for-placement/basic-information/reasonForShortNotice.test.ts
vubalasu: for short notice applications also it needs to see the title this.title = 'Short notice application' 
<=7 emergency, <=28 Short notice application and >29 and <=6 months it shouldnt have title as per Ed @richpjames

richpjames: Ah ok, apologies - this has changed since I last worked on it

server/views/applications/pages/basic-information/reason-for-short-notice.njk
```diff
@@ -23,7 +25,7 @@
         fieldset: {
           legend: {
             text: page.question,
-            classes: "govuk-fieldset__legend--m"
+            classes: "govuk-fieldset__legend--{{ 'm' if page.title else 'l' }}"
```
richpjames, line 28: This needs to have the `isPageHeading` property set to `true` when there isn't a title https://design-system.service.gov.uk/components/radios/

richpjames (approved):

PR #1697, opened 2024-04-09
Add user qualifications for Recovery-focused and Mental Health tasks

edavey:
See Jira [APS-617: Add mental health and recovery focused as user qualifications](https://dsdmoj.atlassian.net/browse/APS-617)

We add these two user qualifications to:

- the admin's "user management" section so that they can be:
 
    - added to and removed from users
    - used to filter the list of users by the qualification

- the "task allocation" section so that tasks can be filtered using these user qualifications

Note that there are two AP Types for the two different Mental Health Specialist APs:

- `MHAP_ST_JOSEPHS`
- `MHAP_ELLIOTT_HOUSE`

but only one user qualification `mental_health_specialist`. i.e. an assessor is qualified to asses applications to both St Joe's and Elliot House.

## Screenshots of UI changes

assign_rf_mh_user_qualifications
filter_users_ by_rf_mh_qualification
filter_tasks_by_mh_rf_qualification

## Dependencies

This should be merged with the corresponding API change: https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1725

### DIFF
--- a/integration_tests/tests/admin/userManagement.cy.ts
+++ b/integration_tests/tests/admin/userManagement.cy.ts
@@ -51,7 +51,7 @@ context('User management', () => {
         'excluded_from_match_allocation',
         'excluded_from_placement_application_allocation',
       ] as const,
-      qualifications: ['pipe', 'emergency', 'esap', 'lao'] as const,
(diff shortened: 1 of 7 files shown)


edavey:
I have done some local tests and I believe that it's safe to merge this either before or after the associated API work.

pezholio (approved):
LGTM 👍

PR #1699, opened 2024-04-09
(APS-569) Remove `getActingUser` HMPPS Auth call

pezholio:
This has been retired in favour of the manage users API. However, all of the profile information we need is fetched via the Approved Premises API, so this is a call that we no longer need to make.

I had to do a bit of jiggery pokery with the task integration tests, because they assumed a user wouldn’t have an AP Area by default, which is no longer the case anyway.

I've tested this to destruction in the local env and seems to work absolutely fine with the ap-tools setup. Will do some more robust testing in dev and preprod too

### DIFF
--- a/integration_tests/mockApis/auth.ts
+++ b/integration_tests/mockApis/auth.ts
@@ -1,10 +1,12 @@
 import jwt from 'jsonwebtoken'
 import { Response } from 'superagent'
-import { ApArea, ApprovedPremisesUserRole as UserRole } from '../../server/@types/shared'
+import { faker } from '@faker-js/faker'
+
+import { ApArea, User, ApprovedPremisesUserRole as UserRole } from '../../server/@types/shared'
 
 import { getMatchingRequests, stubFor } from './setup'
 import tokenVerification from './tokenVerification'
-import { apAreaFactory } from '../../server/testutils/factories'
+import { apAreaFactory, userFactory } from '../../server/testutils/factories'
 
 const createToken = () => {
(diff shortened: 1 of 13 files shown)


integration_tests/mockApis/auth.ts
```diff
@@ -120,37 +122,15 @@ const token = () =>
...
 
-const defaultProfile = {
-  id: defaultUserId,
-  roles: [],
-  isActive: true,
-  apArea: apAreaFactory.build(),
-}
-
-const stubProfile = (roles = [], userId = defaultUserId, isActive = true, apArea = undefined) =>
+const stubProfile = (
```
richpjames, line 127: I wonder if it's worth just passing in a `User` object here? rather than the properties individually?

integration_tests/mockApis/auth.ts
```diff
@@ -161,28 +141,13 @@ const stubProfile = (roles = [], userId = defaultUserId, isActive = true, apArea
       headers: {
         'Content-Type': 'application/json;charset=UTF-8',
       },
-      jsonBody: {
-        ...defaultProfile,
+      jsonBody: userFactory.build({
+        name,
```
richpjames, line 145: and then here we could do `userFactory.build({...user})`

richpjames (approved):
Makes sense!

integration_tests/mockApis/auth.ts
pezholio: Good plan!

PR #1702, opened 2024-04-10
APS 584- Add link to timeline in AP Type cell

richpjames:
# Context

[JIRA](https://dsdmoj.atlassian.net/jira/software/c/projects/APS/boards/1328?assignee=62a050ba9f5d480069c97f49&selectedIssue=APS-584)

# Changes in this PR


server/utils/placementRequests/adminSummary.test.ts
```diff
@@ -128,6 +129,14 @@ describe('adminSummary', () => {
           value: {
             text: apTypeLabels[apType],
           },
+          actions: {
```
pezholio, line 132: What's the reasoning behind having the link next the the AP Type?

richpjames: It's a good point - I'm not sure of the reasoning, I agree that it isn't intuitive as the timeline doesn't really relate to the AP type.  
Similar issue here: https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/1584 

@edavey and I met with Sarah last week to look into a more user friendly solution. 
We think it would be good to introduce three tabs on the page for viewing the assessment, application and timeline instead of linking out to each one... But I think I'd struggle to do that this sprint. 
image 

Another solution could be to add a 'View timeline' link underneath 'View Application' and 'View Assessment' but I assume this design has been approved by stakeholders etc...

pezholio (approved):

PR #1709, opened 2024-04-11
APS 538 - Add feature flag for sufficient information confirmation screen

richpjames:
# Context
- [JIRA](https://dsdmoj.atlassian.net/jira/software/c/projects/APS/boards/1328?assignee=62a050ba9f5d480069c97f49&selectedIssue=APS-591)
- [JIRA](https://dsdmoj.atlassian.net/jira/software/c/projects/APS/boards/1328?assignee=62a050ba9f5d480069c97f49&selectedIssue=APS-538)

We want to add a feature flag for the 'Sufficient Information Confirmation' screen to see if removing it would be a viable user experience. 
Having the screen in place leaves open the possibility of creating an infinite loop [detailed here](https://dsdmoj.atlassian.net/browse/APS-541) and we think the simplest approach may be to remove the confirmation screen.
GjLP1CYdOS(1)
By removing the screen the user may submit a 'clarification note' without properly considering how this will effect the assessment. 

# Technical considerations
In order to add feature flags to the Assessment journey I explored a number of approaches: 

1.  **Add a call to fetch the flags in the `initialize` method of the page class (`SufficientInformation`)**. This failed because although the `initialize` method is called when viewing the page in question it is not called in the `TasklistService` when determining if a task is complete. Therefore if the feature flag changes the logic around which page to display next (as ours does) then this will result in tasks showing as 'In progress' when the user has seen all the screens they are able to. This could be resolved by add the value to the `page` object, passing it through to the view and back to the `page` again via a hidden input but this is suboptimal
2. **Pass the feature flag values in to the `Page` class constructors**. This would work but there are many, many places we would have to make code changes including passing in the value of the feature flags to every view where we render Assessments, Applications and Placement Applications. I did attempt this but ultimately backed out because of the complexity it introduced to functions that shouldn't have to `know` about the feature flags - they were just passing the values along
3. **Introduce middleware to add the feature flag values to the `process.env`**. This allows us to access flags via `process.env` as we did before the introduction of flipit (#1651) but with the advantage of being able to toggle the flags without a deploy. However we now have to determine how often we want to call the feature flag service. Too often and it will cause performance issues in this app and potentially overload the feature flag service. Too little and there will be a delay in flipping the flag and the user seeing the desired changes. I have opted for a delay of 1.5 seconds to begin with which potentially tends towards the former but we can adjust as we go.

### DIFF
--- a/server/middleware/setupFeatureFlags.ts
+++ b/server/middleware/setupFeatureFlags.ts
@@ -1,0 +1,35 @@
+import { RequestHandler, Router } from 'express'
+import { throttle } from 'underscore'
+import { Services } from '../services'
+import FeatureFlagService, { FeatureFlag } from '../services/featureFlagService'
+
+export const featureFlagsToUse: Array<FeatureFlag> = ['allow-sufficient-information-request-without-confirmation']
+export const throttleTime = 15 * 1000
+
+export const setupFeatureFlags = ({ featureFlagService }: Services): Router => {
+  const router = Router()
+
+  router.use(featureFlagHandler(featureFlagService))
+  return router
+}
+
+export const featureFlagHandler = (featureFlagService: FeatureFlagService): RequestHandler => {
+  const retrieveFlagThrottled = throttle(retrieveFlags, throttleTime)
+  return async (_, __, next) => {
+    try {
+      retrieveFlagThrottled(featureFlagService)
+
+      return next()
+    } catch (error) {
+      return next(error)
+    }
+  }
+}
+
+export const retrieveFlags = (featureFlagService: FeatureFlagService) => {
+  featureFlagsToUse.forEach(async flag => {
+    const flagBool = await featureFlagService.getBooleanFlag(flag)
+
+    process.env[flag] = flagBool.toString()
+  })
+}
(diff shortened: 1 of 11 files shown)


server/middleware/setupFeatureFlags.ts
```diff
@@ -0,0 +1,25 @@
...
+  const router = Router()
+  router.use(populateFeatureFlags(featureFlagService))
+  return router
+}
+
+export const populateFeatureFlags = (featureFlagService: FeatureFlagService): RequestHandler => {
+  return async (_, __, next) => {
+    try {
+      featureFlagsToUse.forEach(async flag => {
+        const flagBool = await featureFlagService.getBooleanFlag(flag)
```
richpjames, line 17: @pezholio I think we might want to throttle this. Unsure of a suitable delay, maybe somewhere between 1 and 5 minutes?
https://underscorejs.org/#throttle

richpjames: I have added the throttle on line 10

pezholio: Will this run on every request, or can we ensure it only runs when we want it?

richpjames: Now the throttling is in place I believe it will only run every 1.5 seconds, but will test to confirm

pezholio (approved):
LGTM 👍 - I am a bit worried about making a bunch of calls to the feature flag endpoint, but let's see how we get on before overengineering anything!

PR #1710, opened 2024-04-11
APS-602: Show Completed tasks in Task allocation section

vubalasu:
# Context

# Changes in this PR
Show Completed tasks in Task allocation section
### Before
Screenshot 2024-04-11 at 12 08 19

### After
Screenshot 2024-04-12 at 12 01 51

### DIFF
--- a/integration_tests/mockApis/tasks.ts
+++ b/integration_tests/mockApis/tasks.ts
@@ -17,6 +17,7 @@ export default {
     sortBy = 'createdAt',
     apAreaId = '',
     types = [],
+    isCompleted = 'false',
   }: {
     tasks: Array<Task>
     page: string
@@ -26,6 +27,7 @@ export default {
     sortBy: TaskSortField
     apAreaId: string
     types: Array<TaskType>
+    isCompleted: string
   }): SuperAgentRequest => {
     const queryParameters: Record<string, unknown> = {
       page: {
@@ -40,6 +42,10 @@ export default {
 
       allocatedToUserId: {
         equalTo: allocatedToUserId,
+      },
+
+      isCompleted: {
+        equalTo: isCompleted,
       },
     }
 
--- a/server/utils/tasks/listTable.test.ts
+++ b/server/utils/tasks/listTable.test.ts
@@ -7,6 +7,11 @@ import {
 import {
   allocatedTableRows,
   allocationCell,
+  completedAtDateCell,
+  completedByCell,
+  completedTableRows,
+  decisionCell,
+  getDecisionOutcome,
   getTaskType,
   nameAnchorCell,
   statusBadge,
@@ -271,12 +276,17 @@ describe('table', () => {
         {
           text: 'Allocated',
           active: true,
-          href: '/tasks?allocatedFilter=allocated',
+          href: '/tasks?allocatedFilter=allocated&activeTab=allocated',
         },
         {
           text: 'Unallocated',
           active: false,
-          href: '/tasks?allocatedFilter=unallocated',
+          href: '/tasks?allocatedFilter=unallocated&activeTab=unallocated',
+        },
+        {
+          active: false,
+          href: '/tasks?allocatedFilter=allocated&activeTab=completed',
+          text: 'Completed',
         },
       ])
     })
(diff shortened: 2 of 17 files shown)


server/utils/tasks/listTable.ts
```diff
@@ -28,6 +29,32 @@ const getTaskType = (task: Task): string => {
   return taskType
 }
 
+const getDecisionOutcome = (outcome: string): string => {
```
pezholio, line 32: I think I'd rather have `getDecisionOutcome` functions for both decision type, with either `AssessmentDecision`, `PlacementApplicationDecision` or `PlacementRequestTaskOutcome` passed in. That way you can have a record type like (for example):

```typescript
const assessmentDecisionTranslations: Record<AssessmentDecision, string> = {
    accepted: 'Accepted',
    rejected: 'Rejected'
}
```
 
And then you can have a getDecisionOutcome translator like so:

```typescript
const getDecisionOutcome = (outcome: AssessmentDecision): string => {
    return assessmentDecisionTranslations[outcome]
}
```

That way if any of the types change, we're alerted to this and can fix it.

server/utils/tasks/listTable.ts
```diff
@@ -28,6 +29,32 @@ const getTaskType = (task: Task): string => {
...
+    case 'withdraw':
+      return 'Withdraw'
+    case 'withdrawn_by_pp':
+      return 'Withdrawn by pp'
+    default:
+      return ''
+  }
+}
+
+export const getDecision = (task: Task): string => {
```
pezholio, line 47: Looks like we're missing a case for `PlacementRequestTask`s?

pezholio (changes requested):

server/utils/tasks/listTable.ts
vubalasu: but we are requesting only 'PlacementApplication', 'Assessment'

pezholio: I think it'd be best to have it there for completeness, as this part of the code doesn't have any knowledge of what we've requested at the controller level. If this changes for any reason, or we try and use this code in a place where we request all types, we could end up with unintended consequences.

server/utils/tasks/listTable.ts
```diff
@@ -93,13 +147,44 @@ const unallocatedTableRows = (tasks: Array<Task>): Array<TableRow> => {
...
 
-const tasksTableRows = (tasks: Array<Task>, allocatedFilter: string): Array<TableRow> => {
-  const rows: Array<TableRow> =
-    allocatedFilter === 'allocated' ? allocatedTableRows(tasks) : unallocatedTableRows(tasks)
+export const completedAtDateCell = (item: Task): TableCell => ({
+  text: item.outcomeRecordedAt ? DateFormats.isoDateTimeToUIDateTime(item.outcomeRecordedAt) : '',
+})
+
+export const completedByCell = (task: Task): TableCell => ({
+  text: task.allocatedToStaffMember?.name,
```
richpjames, line 155: If the `task` doesn't have an `allocatedToStaffMember` property the `text` property in the returned object will be `undefined`. It's better to return an empty string

integration_tests/mockApis/tasks.ts
```diff
@@ -26,6 +27,7 @@ export default {
     sortBy: TaskSortField
     apAreaId: string
     types: Array<TaskType>
+    isCompleted: string
```
richpjames, line 30: I *think* this should be a `boolean`

server/utils/tasks/listTable.ts
```diff
@@ -28,6 +39,49 @@ const getTaskType = (task: Task): string => {
...
+
+const getPlacementApplicationDecisionOutcome = (outcome: PlacementApplicationDecision): string => {
+  return placementApplicationDecisionTranslations[outcome]
+}
+
+const getPlacementRequestTaskOutcome = (outcome: PlacementRequestTaskOutcome): string => {
+  return placementRequestTaskOutcomeTranslations[outcome]
+}
+
+export const getDecisionOutcome = (task: Task): string => {
```
richpjames, line 71: In the After screenshot it looks like this function is returning `undefined` 
image

server/utils/tasks/listTable.ts
```diff
@@ -93,13 +147,44 @@ const unallocatedTableRows = (tasks: Array<Task>): Array<TableRow> => {
...
+
+const completedTableRows = (tasks: Array<Task>): Array<TableRow> => {
+  const rows = [] as Array<TableRow>
+
+  tasks.forEach(task => {
+    rows.push([
+      nameAnchorCell(task),
+      completedAtDateCell(task),
+      completedByCell(task),
+      taskTypeCell(task),
```
richpjames, line 170: From the 'After' screenshot it looks like the `Task type` column is empty, it would be good to understand why

vubalasu:
@richpjames pushed please check

server/utils/tasks/listTable.test.ts
```diff
@@ -331,14 +376,88 @@ describe('table', () => {
...
+          completedAtDateCell(task),
+          completedByCell(task),
+          taskTypeCell(task),
+          decisionCell(task),
+        ],
+      ])
+    })
+  })
+
+  describe('completedTableHeaders', () => {
```
richpjames, line 425: This should be `tasksTableHeader`

richpjames (approved):

pezholio (approved):

PR #1711, opened 2024-04-11
(APS 603) Show pending placement request in CRU

pezholio:
This updates the CRU dashboard to add a new tab that shows applications without placement requests. I've also done a bit of shifting around to tidy up the controller and path names.

## Screenshot

Placement Requests -- should list applications that have no placement request (1)


server/controllers/redirectController.ts
```diff
@@ -0,0 +1,10 @@
+import { Request, RequestHandler, Response } from 'express'
+import { Path } from 'static-path'
+
+export default class RedirectController {
+  redirect<T extends string>(path: Path<T>): RequestHandler {
+    return async (_req: Request, res: Response) => {
+      return res.redirect(301, path.pattern)
```
richpjames, line 7: I think it would be clearer if we didn't return the result of `res.redirect` as it is `void`

richpjames (approved):

PR #1714, opened 2024-04-15
APS-612: Show applicant guidance when blocked by LAO restriction

vubalasu:
# Context

# Changes in this PR
Show applicant guidance when blocked by LAO restriction
### Before
image

### After
Screenshot 2024-04-12 at 17 30 37

### DIFF
--- a/server/controllers/apply/applicationsController.test.ts
+++ b/server/controllers/apply/applicationsController.test.ts
@@ -337,7 +337,7 @@ describe('applicationsController', () => {
       beforeEach(() => {
         request = createMock<Request>({
           user: { token },
-          flash: jest.fn().mockReturnValue([person.crn]),
+          flash: jest.fn().mockReturnValueOnce([person.crn]).mockReturnValueOnce(['true']),
         })
         personService.findByCrn.mockResolvedValue(person)
         personService.getOffences.mockResolvedValue([offence])
@@ -424,6 +424,7 @@ describe('applicationsController', () => {
           errors: errorsAndUserInput.errors,
           errorSummary: errorsAndUserInput.errorSummary,
           ...errorsAndUserInput.userInput,
+          restrictedPerson: 'true',
         })
         expect(request.flash).toHaveBeenCalledWith('crn')
       })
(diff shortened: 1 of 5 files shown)


pezholio (commented):
I'm thinking actually the way we handle this error doesn't translate into reality. At the moment, if you try a CRN that is restricted in dev (e.g X698334), the API returns a 403, which bubbles up to the frontend and logs the user out.

pezholio:
Ah, this is because I have the Restricted CRNs permission, which isn't in use for applicants. When I remove that permission, it works as expected 👍

server/controllers/apply/applicationsController.test.ts
```diff
@@ -424,6 +424,7 @@ describe('applicationsController', () => {
           errors: errorsAndUserInput.errors,
           errorSummary: errorsAndUserInput.errorSummary,
           ...errorsAndUserInput.userInput,
+          restrictedPerson: person.crn,
```
richpjames, line 427: This value is set to be 'true' in People Controller but here we are setting it to be a CRN

pezholio (approved):
I think this is OK for now. I've had a look through the design system to see if there's a better way of giving more verbose error guidance, but there doesn't look to be anything better out of the box. Hopefully we can revisit this when we have a designer 👍

server/controllers/apply/applicationsController.ts
```diff
@@ -145,12 +145,13 @@ export default class ApplicationsController {
           })
         }
       }
-
+      const restrictedPerson = req.flash('restrictedPerson')[0]
```
richpjames, line 148: could this error if req.flash doesn't have an entry for 'restrictedPerson'?

richpjames (changes requested):

server/controllers/apply/applicationsController.test.ts
vubalasu: updated

server/controllers/apply/applicationsController.ts
vubalasu: gets undefined

richpjames (approved):

PR #1716, opened 2024-04-15
(APS-622) Mappa missing error message could be more helpful

pezholio:
This adds an error message to the Risk Flags sidebar to inform a user that the Mappa is not showing because of an error. This allows us to differentiate between cases that genuinely do not have a Mappa value and those that have failed.

We need to do some other work to ensure we can backfill when errors do occur, but this should help reduce confusion in the meantime.

### Before

image

### After

image

### DIFF
--- a/server/utils/utils.test.ts
+++ b/server/utils/utils.test.ts
@@ -188,6 +188,7 @@ describe('mapApiPersonRiskForUI', () => {
     expect(actual).toEqual(
       expect.objectContaining({
         mappa: {
+          status: undefined,
           lastUpdated: '',
         },
       }),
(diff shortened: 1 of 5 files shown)


server/utils/utils.test.ts
```diff
@@ -172,100 +171,53 @@ describe('mapApiPersonRiskForUI', () => {
...
-      },
-      tier: {
-        lastUpdated: DateFormats.isoDateToUIDate(risks.tier?.value?.lastUpdated || ''),
-        level: risks.tier.value?.level,
-      },
-    })
+    expect(actual).toEqual(
+      expect.objectContaining({
+        mappa: {
+          lastUpdated: '',
```
richpjames, line 191: We should test the value of `mappa.status` here too

richpjames (approved):

PR #1719, opened 2024-04-16
APS-613: Applicant sees guidance when able to create an LAO application

vubalasu:
# Context

# Changes in this PR
Applicant sees guidance when able to create an LAO application
### Before
Screenshot 2024-04-15 at 15 38 53

### After
Screenshot 2024-04-15 at 15 41 45


richpjames (commented):
Have we been told to use `EQUIP` instead of `EQuiP`?  
I think EQuiP is the correct casing 
image

vubalasu:
@richpjames it mentioned revised copy in the ticket with
image

PR #1722, opened 2024-04-16
APS 589 - View timeline events by crn

richpjames:
# Context
[JIRA](https://dsdmoj.atlassian.net/jira/software/c/projects/APS/boards/1328?assignee=62a050ba9f5d480069c97f49&selectedIssue=APS-589) 

# Changes in this PR
- This adds the timeline screen but it isn't easy to judge how it will look without realistic data that we have in preprod. Rather than feature switch it I have opted to not link to the first page from anywhere in the in the service

## Screenshots of UI changes
Application timeline -- shows a timeline for a CRN

Application timeline -- shows a timeline for a CRN (1)

PR #1731, opened 2024-04-17
APS 589 - View timeline events by crn

richpjames:
# Context
#1722 added the timeline and this PR builds on it. 
We add usability improvements like error handling and changing the initial request from a POST to a GET to allow the user to bookmark, share the link and use the back button in the browser.

[JIRA](https://dsdmoj.atlassian.net/jira/software/c/projects/APS/boards/1328?assignee=62a050ba9f5d480069c97f49&selectedIssue=APS-589)

### DIFF
--- a/server/routes/people.ts
+++ b/server/routes/people.ts
@@ -21,7 +21,7 @@ export default function routes(controllers: Controllers, router: Router, service
   })
 
   get(peoplePaths.timeline.find.pattern, timelineController.find(), { auditEvent: 'SEARCH_FOR_TIMELINE' })
-  post(peoplePaths.timeline.find.pattern, timelineController.show(), {
+  get(peoplePaths.timeline.show.pattern, timelineController.show(), {
     auditEvent: 'SHOW_PERSON_TIMELINE',
     auditBodyParams: ['crn'],
   })
--- a/server/controllers/people/peopleController.test.ts
+++ b/server/controllers/people/peopleController.test.ts
@@ -5,7 +5,7 @@ import PeopleController from './peopleController'
 import PersonService from '../../services/personService'
 import { personFactory } from '../../testutils/factories'
 import { errorMessage, errorSummary } from '../../utils/validation'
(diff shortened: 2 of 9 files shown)


server/routes/people.ts
```diff
@@ -21,7 +21,7 @@ export default function routes(controllers: Controllers, router: Router, service
   })
 
   get(peoplePaths.timeline.find.pattern, timelineController.find(), { auditEvent: 'SEARCH_FOR_TIMELINE' })
-  post(peoplePaths.timeline.find.pattern, timelineController.show(), {
+  get(peoplePaths.timeline.show.pattern, timelineController.show(), {
```
richpjames, line 24: @pezholio are these verbs as you'd expect? `find` for the screen the user enters the CRN, `show` for the timeline screen itself?

pezholio: Yeah, I think that makes sense. I'm getting more and more agnostic about "REST-y" verbs as I go on, so I think whatever seems sensible is fine 👍

pezholio: What WOULD be great though would be if we could have a canonical URL for the timeline, so rather than having:

```
people/timeline/show?crn=ABC123
```

as the sharable / bookmarkable URL, we had:

```
people/CRN123/timeline
```

I guess to do that we'd POST to the `find` endpoint, if we find the user we redirect to the canonical URL, otherwise we return an error.

pezholio: I think it's mainly for "correctness" and ensure's there's a canonical URL that people can refer to. Working with Tim Berners-Lee has rubbed off on me! It's not a major thing

richpjames: I think in the interests of getting this in before the demo I think it might be best to merge and come back to it

pezholio: No, I totally agree 👍

pezholio (approved):
LGTM 👍

PR #1734, opened 2024-04-18
(APS 637) Always show offences on adhoc booking screen

pezholio:
This follows on from #1728 to update the adhoc bookings screen to always show the offences, rather than assuming if one is present then that's the correct one.

## Screenshot

image

### DIFF
--- a/server/utils/offenceUtils.ts
+++ b/server/utils/offenceUtils.ts
@@ -42,4 +42,17 @@ const offenceRadioButton = (offence: ActiveOffence) => {
   `
 }
 
-export { offenceTableRows, offenceRadioButton }
+const offenceRadioItems = (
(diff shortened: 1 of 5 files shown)


server/utils/offenceUtils.ts
```diff
@@ -42,4 +42,14 @@ const offenceRadioButton = (offence: ActiveOffence) => {
   `
 }
 
-export { offenceTableRows, offenceRadioButton }
+const offenceRadioItems = (offences: Array<ActiveOffence>, selectedOffenceNumber: string | null | undefined) => {
```
richpjames, line 45: `ActiveOffence[offenceId]` is a bit more descriptive than `string` here

PR #1742, opened 2024-04-19
(APS 691) Show requested information on information received page

pezholio:
We've had a user request that they've "lost" the information they requested after submitting a request for information, which is needed for them to review what their request was. This adds a details component which surfaces this information for them to refer back to.

I've also updated the timeline to add HTML linebreaks to break up long pieces of text in the timeline (preparing us for https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1783).

## Screenshot

image

image


richpjames (approved):
I think it's a good idea to make this info more visible to the user, good work Stu!

PR #1759, opened 2024-04-24
APS 650 - See application suitable status label on Apply pane

richpjames:
# Context
We need to update the copy on one of the Application Status tags. There will be more work forthcoming on updating other statuses.
We currently have a bunch of different statuses to render [tags](https://design-system.service.gov.uk/components/tag/) with different names in different places.

[JIRA](https://dsdmoj.atlassian.net/jira/software/c/projects/APS/boards/1328?assignee=62a050ba9f5d480069c97f49&selectedIssue=APS-650)

# Changes in this PR
We add a class to render tags and then move the rendering of the majority of the tags in the application to use this class in order to DRY things up. We haven't exhaustively added all rendering of tags to this as there are some more edge-case-y simple tags but the majority of tags are now here.


server/utils/statusTag.ts
```diff
@@ -0,0 +1,81 @@
+import { ApprovedPremisesApplicationStatus as ApplicationStatus } from '../@types/shared'
+
+export type StatusTagOptions = { addLeftMargin?: boolean; showOnOneLine?: boolean }
+type StatusColour = 'green' | 'grey' | 'blue' | 'yellow' | 'red' | 'pink' | ''
+type Status = ApplicationStatus
+export class StatusTag<T extends Status> {
```
pezholio, line 15: Could this be an `interface` instead of a class? I don't think it gets used on it's own

server/utils/statusTag.ts
```diff
@@ -103,3 +102,34 @@ export class TaskStatusTag extends StatusTag<TaskStatus> {
     })
   }
 }
+
+export class BookingStatusTag extends StatusTag<BookingStatus> {
```
pezholio, line 156: I wonder if it might be better to have these in seperate files namespaced into the appropriate folder, i.e `utils/bookings/bookingStatus.ts` etc etc

pezholio (approved):
I think this is really good overall, just got some minor tidying up suggestions. Feel free to push back!

server/utils/statusTag.ts
richpjames: Yeah I did think about this, it should be an abstract class. The issue is we then loose the ability to test the `html` method logic for all subclasses because you can't instantiate an abstract class to test it. 
I think a solution would be to extract the `html` logic to a seperate function that `html` simply calls... I'll give it a go

server/utils/statusTag.ts
richpjames: yeah good shout

PR #1771, opened 2024-04-26
Fix/Replace 'Application suitable' statuses with one status

richpjames:
In the UI the three applicationSuitableStatuses are now represented by one status 'Application Suitable'. For the Application Status Select dropdown on the application dashboard we were showing each of these with the copy 'Application suitable' but what we need to show one 'Application suitable' status to represent the three API statuses. We have allowed the endpoint to receive an array of statuses here: https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1813/commits/638c5249b1e9b5bb7295bdae67b1570a32a0eaa5 and now we make the change in the UI. We simply set the value of the 'Application suitable' status to be an array of the three statuses and, once the types have been corrected, everything works


pezholio (approved):
LGTM 👍

PR #1773, opened 2024-04-29
Show all 'Application suitable' statuses as the same colour

richpjames:
We need to ensure that all application suitable statuses are the same colour for now, as they are semantically the same for users.

### DIFF
--- a/server/utils/applications/statusTag.ts
+++ b/server/utils/applications/statusTag.ts
@@ -18,6 +18,8 @@ export class ApplicationStatusTag extends StatusTag<ApplicationStatus> {
     pendingPlacementRequest: APPLICATION_SUITABLE,
   }
 
+  readonly applicationSuitableStatus: 'blue' = 'blue' as const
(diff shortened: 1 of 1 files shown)


edavey:
should we expand this commit message little to explain in more detail why we think it's wrong to rely to a colour change to distinguish between:

- accepted with placement request
- accepted without placement request

My understanding is that using colour is unreliable for accessibility reasons and we recommend relying on the separate call to action in the "actions" column e.g:

- "View placement requests"
- "Create placement request"

edavey (approved):

richpjames:
@edavey this is actually a defect left from #1759 unrelated to [APS-651](https://dsdmoj.atlassian.net/jira/software/c/projects/APS/boards/1328?assignee=62a050ba9f5d480069c97f49&selectedIssue=APS-651) but all the more reason to leave a more comprehensive commit message. Thanks for the prompt

PR #1790, opened 2024-05-01
Fix/allow feature flags to fail gracefully

richpjames:
# Changes in this PR
Previously if our calls to the Feature Flag service failed then the call would throw an error and the page could crash. Now we just set the value to false.

### DIFF
--- a/server/middleware/setupFeatureFlags.ts
+++ b/server/middleware/setupFeatureFlags.ts
@@ -2,6 +2,7 @@ import { RequestHandler, Router } from 'express'
 import { throttle } from 'underscore'
 import { Services } from '../services'
 import FeatureFlagService, { FeatureFlag } from '../services/featureFlagService'
+import logger from '../../logger'
 
 export const featureFlagsToUse: Array<FeatureFlag> = [
   'allow-sufficient-information-request-without-confirmation',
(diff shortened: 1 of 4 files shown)


server/middleware/setupFeatureFlags.ts
```diff
@@ -30,9 +31,22 @@ export const featureFlagHandler = (featureFlagService: FeatureFlagService): Requ
...
+    featureFlagsToUse.map(async flag => {
+      const flagBool = await featureFlagService.getBooleanFlag(flag)
+      process.env[flag] = flagBool.toString()
+    }),
+  )
+}
 
-    process.env[flag] = flagBool.toString()
-  })
+export const retrieveFlag = async (flag: FeatureFlag, featureFlagService: FeatureFlagService) => {
```
vubalasu, line 42: may be this can be simplified
export const retrieveFlag = async (flag: FeatureFlag, featureFlagService: FeatureFlagService) => {
  let flagValue = false
  try {
    flagValue = await featureFlagService.getBooleanFlag(flag)
  } catch (error) {
    logger.error(`Error retrieving feature flag ${flag}`, error)
  }
  return flagValue
}

edavey (approved):
Have you been able to test this out effectively locally? Or indeed in the cloud `dev` env?

server/middleware/setupFeatureFlags.ts
richpjames: yeah I think that is a good idea

vubalasu (approved):

richpjames:
@edavey I have tested this locally now and it works as expected. I am not sure how to test it in the cloud env as we don't own the infrastructure Flipt runs on

PR #1791, opened 2024-05-01
Fix/Run E2Es from UI repo in pipeline

richpjames:
Some tweaks to our CI pipeline following #1495 to allow the E2E tests to run properly in our CI pipeline. 
- Require the E2Es to pass before merging. Whilst the E2Es won't be running against the code that has changed in the PR (they require the built and deployed dev Docker container of the UI to run - because of HMPPS Auth) this will prevent us from merging code when the E2Es have a pre-existing breakage.
- We remove the superfluous `npm i` step. 
- Remove a test path to Withdraw an application after submitted. This was removed in the [E2E test repo](https://github.com/ministryofjustice/hmpps-approved-premises-e2e/) but not copied across. 
- Update the .gitignore to include some Playwright generated files
- Update the `package.json` scripts so they function properly


pezholio (approved):

pezholio:
Bit of a sidebar, it might be possible to add a manual step to build and deploy to the test environment and run the E2Es there too - probably overkill to have it on every push, but might be good to have as an option

PR #1794, opened 2024-05-01
Remove NEW_WITHDRAWALS_FLOW_DISABLED feature flag

richpjames:
Given the new withdrawals flow has been live for months now without issue we should remove the feature flag to reduce complexity in the application


edavey (approved):
are there any accompanying changes to the E2E test suite needed?

richpjames:
@edavey I don't think so. The tests are written to test the app with the feature switch off (new withdrawals flow enabled) and we don't flip the feature switch as part of testing

PR #1796, opened 2024-05-02
Run E2Es with deployed code

richpjames:
#1791 enabled running E2Es on PRs but also inadvertently stopped them running on the just-merged deployed docker container. With this change I am hoping that E2Es will now run both on PR AND against main after a change is deployed


edavey (approved):

davidatkinsuk (commented):
LGTM

PR #1806, opened 2024-05-08
Aps 736 UI build list of requests for placements from new api endpoints

richpjames:
# Context

# Changes in this PR


edavey:
Great stuff! Do you think it would be easier to scan the list of RfPs if the "Status" row were at the top (rather than the bottom) of each element?

status_at_top_or_bottom

richpjames:
@edavey yes, good idea and easy change

vubalasu (approved):

server/utils/placementRequests/statusTag.ts
```diff
@@ -0,0 +1,33 @@
...
+  static readonly colours: Record<RequestForPlacementStatus, string> = {
+    awaiting_match: 'yellow',
+    person_arrived: 'blue',
+    person_departed: 'blue',
+    person_not_arrived: 'blue',
+    placement_booked: 'grey',
+    request_rejected: 'red',
+    request_submitted: 'yellow',
+    request_withdrawn: 'red',
+  }
```
edavey, line 25: Nice! These configuration lists will be easy to maintain.

I think we have the wrong colour for `person_not_arrived`. This status means "the person failed to arrive", so I think red is probably the right colour? 

Does it still make sense for `placement_booked` to be in grey or is blue more appropriate as in Adam's mockup?

server/utils/placementRequests/requestForPlacementSummaryCards.test.ts
```diff
@@ -0,0 +1,165 @@
...
+import paths from '../../paths/apply'
+import { DateFormats } from '../dateUtils'
+import { RequestForPlacementStatusTag } from './statusTag'
+
+describe('RequestForPlacementSummaryCards', () => {
+  const actingUserId = '123'
+  const applicationId = '456'
+
+  describe('if the RfP cannot be directly withdrawn', () => {
+    it('should generate a summary card for the request section', () => {
```
edavey, line 15: would it be more accurate to say that this test is demonstrating that we see both a summary card and an empty list of available actions? e.g.

`it('should generate a summary card for the request section AND include no available actions', () => {`

which makes me think that it might be clearer to describe the behaviour in 4 tests grouped in 2 scenarios:

```
describe('if the RfP cannot be directly withdrawn', () => {
    it('should generate a summary card for the request section'
    it('should not offer any actions'

 describe('if the RfP can be withdrawn', () => {
    it('should generate a summary card for the request section'
    it('should include a withdraw action item'
```

richpjames:
@vubalasu good idea

edavey (approved):
👍 colour-coded status labels, full RfP detail -- with 20% less code!

edavey:
> @richpjames looks good not sure is it good to sort rfps latest first

good point Bala! thank you

server/utils/placementRequests/statusTag.ts
richpjames: To be honest I was guessing these colours so happy to defer to these suggestions

PR #1807, opened 2024-05-08
E2Es: Assert the expected task due date is >= actual due date

richpjames:
We've been having CI failures whereby the E2Es expect the due date for a task to be completed by was further in the future than the API said it was because the API accounts for working days and bank holidays. In order to counter this we now peform a comparison of the range the due date should be in. The alternative is duplicating the entire logic from the API into the E2Es which seems like the greater evil.


davidatkinsuk (approved):
LGTM 👍

PR #1812, opened 2024-05-09
APS-777: Show newest applications first in Apply -- as default ordering

vubalasu:
…s unsuitable

# Context

 https://dsdmoj.atlassian.net/browse/APS-777 

# Changes in this PR
Show newest applications first in Apply -- as default ordering
### After
Screenshot 2024-05-10 at 15 31 35

### DIFF
--- a/integration_tests/tests/admin/placementRequests.cy.ts
+++ b/integration_tests/tests/admin/placementRequests.cy.ts
@@ -555,7 +555,7 @@ context('Placement Requests', () => {
   it('should list applications that have no placement request', () => {
     const applications = applicationSummaryFactory.buildList(2)
 
-    cy.task('stubAllApplications', { applications, page: '1' })
+    cy.task('stubAllApplications', { applications, page: '1', sortDirection: 'asc' })
 
     // Given I am on the placement request dashboard filtering by the pendingPlacement status
     const listPage = ListPage.visit('status=pendingPlacement')
@@ -563,7 +563,7 @@ context('Placement Requests', () => {
     // Then I should see a list of applications with no placement requests
     listPage.shouldShowApplications(applications)
 
-    cy.task('verifyDashboardRequest', { status: 'pendingPlacementRequest' }).then(requests => {
+    cy.task('verifyDashboardRequest', { status: 'pendingPlacementRequest', sortDirection: 'asc' }).then(requests => {
       expect(requests).to.have.length(1)
     })
 
@@ -575,6 +575,7 @@ context('Placement Requests', () => {
     cy.task('stubAllApplications', {
       applications: apAreaApplications,
       page: '1',
+      sortDirection: 'asc',
       searchOptions: { apAreaId: apArea.id, releaseType: 'rotl' },
     })
     listPage.getSelectInputByIdAndSelectAnEntry('apArea', apArea.name)
@@ -586,6 +587,7 @@ context('Placement Requests', () => {
 
     cy.task('verifyDashboardRequest', {
       status: 'pendingPlacementRequest',
+      sortDirection: 'asc',
       searchOptions: { apAreaId: apArea.id, releaseType: 'rotl' },
     }).then(requests => {
       expect(requests).to.have.length(1)
@@ -594,7 +596,7 @@ context('Placement Requests', () => {
   ;(['tier', 'releaseType'] as const).forEach(field => {
     it(`supports pending placement requests sorting by ${field}`, () => {
       const applications = applicationSummaryFactory.buildList(2)
-      cy.task('stubAllApplications', { applications, page: '1' })
+      cy.task('stubAllApplications', { applications, page: '1', sortDirection: 'asc' })
       cy.task('stubAllApplications', {
         applications,
         sortBy: field,
--- a/integration_tests/tests/apply/dashboard.cy.ts
+++ b/integration_tests/tests/apply/dashboard.cy.ts
@@ -65,7 +65,52 @@ context('All applications', () => {
   })
 
   it('supports sorting by createdAt', () => {
-    shouldSortByField('createdAt')
+    // Given I am logged in
+    cy.signIn()
+
+    // And there is a page of applications
+    const applications = applicationSummaryFactory.buildList(10)
+
+    cy.task('stubAllApplications', { applications, page: '1' })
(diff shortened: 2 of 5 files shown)


server/services/applicationService.test.ts
```diff
@@ -376,7 +376,7 @@ describe('ApplicationService', () => {
       expect(result).toEqual(paginatedResponse)
 
       expect(applicationClientFactory).toHaveBeenCalledWith(token)
-      expect(applicationClient.dashboard).toHaveBeenCalledWith(1, 'createdAt', 'asc', {})
+      expect(applicationClient.dashboard).toHaveBeenCalledWith(1, 'createdAt', 'desc', {})
```
edavey, line 379: do you think we should go a little further and use this opportunity to document our intention here?

e.g. by describing 'what' and 'why' in our scenario e.g.

```ts
it('calls the dashboard client method with sorting options to display newest first', async () => {
  ...
}
```

edavey (commented):
It looks like the commit message from 729 has ended up on this commit?

server/services/applicationService.test.ts
```diff
@@ -393,6 +393,15 @@ describe('ApplicationService', () => {
         status: 'assesmentInProgress',
       })
     })
+
+    it('parameters not passed then calls the client with default values for page no and search and sort options', async () => {
```
edavey, line 397: I think perhaps the key point we need to include here is that that the default params give us our applications **ordered newest first**?

server/controllers/apply/applicationsController.test.ts
```diff
@@ -128,6 +128,87 @@ describe('applicationsController', () => {
...
+      const paginatedResponse = paginatedResponseFactory.build({
+        data: applicationFactory.buildList(2),
+      }) as PaginatedResponse<ApprovedPremisesApplicationSummary>
+
+      const paginationDetails = {
+        hrefPrefix: paths.applications.dashboard({}),
+        pageNumber: 1,
+      }
+
+      applicationService.dashboard.mockResolvedValue(paginatedResponse)
```
richpjames, line 142: it would improve the clarity here if we used `jest-when` like we do here: https://github.com/ministryofjustice/hmpps-approved-premises-ui/blob/1c246008d83b48c32398507b194fd92876486858/server/controllers/tasksController.test.ts#L59 
It helps to show what we're expecting the function to be called with and preserves type safety

server/controllers/apply/applicationsController.ts
```diff
@@ -34,11 +34,19 @@ export default class ApplicationsController {
...
+      const paginationDetails = getPaginationDetails<ApplicationSortField>(
         req,
         paths.applications.dashboard({}),
         searchOptions,
       )
+
+      const { pageNumber, hrefPrefix, sortBy } = paginationDetails
+      let { sortDirection } = paginationDetails
+
+      if (!sortBy || (sortBy === 'createdAt' && !sortDirection)) {
```
richpjames, line 46: Should we be setting the `sortBy` to `createdAt` here too?

integration_tests/tests/admin/placementRequests.cy.ts
```diff
@@ -555,15 +555,15 @@ context('Placement Requests', () => {
   it('should list applications that have no placement request', () => {
     const applications = applicationSummaryFactory.buildList(2)
 
-    cy.task('stubAllApplications', { applications, page: '1' })
+    cy.task('stubAllApplications', { applications, page: '1', sortDirection: 'asc' })
```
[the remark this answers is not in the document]
vubalasu: by default it is desc and here we are overriding with asc

server/controllers/apply/applicationsController.ts
vubalasu: we need sortDirection  to set desc when sortBy not provided and sortBy is createdAt here

server/controllers/apply/applicationsController.test.ts
```diff
@@ -128,6 +128,87 @@ describe('applicationsController', () => {
...
+        token,
+        paginationDetails.pageNumber,
+        undefined,
+        'desc',
+        searchOptions,
+      )
+      expect(getSearchOptions).toHaveBeenCalledWith(request, ['crnOrName', 'status'])
+      expect(getPaginationDetails).toHaveBeenCalledWith(request, paths.applications.dashboard({}), searchOptions)
+    })
+    it('if sort by created at and sort direction not provided, calls the dashboard service with the sort direction desc', async () => {
```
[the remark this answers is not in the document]
vubalasu: added test this is when the parameters not provided to check it sorts by desc

server/controllers/apply/applicationsController.ts
richpjames: Maybe I'm missing something here but if `sortBy` is falsy and we set `sortDirection` to `desc` what will we be sorting by?

integration_tests/tests/apply/dashboard.cy.ts
```diff
@@ -65,7 +65,52 @@ context('All applications', () => {
...
+    // Then the API should have received a request for the sort
+    cy.task('verifyDashboardRequest', { page: '1', sortBy: 'createdAt', sortDirection: 'desc' }).then(requests => {
+      expect(requests).to.have.length.greaterThan(0)
+    })
+
+    // When I sort by created at
+    page.clickSortBy('createdAt')
+
+    // Then the API should have received a request for the sort
+    cy.task('verifyDashboardRequest', { page: '1', sortBy: 'createdAt', sortDirection: 'desc' }).then(requests => {
```
richpjames, line 97: This assertion is the same as the one on line 89. It would be good to assert the number of requests that we expect

vubalasu: sortby is falsy then we have in applicationService we have sortBy cretaedAt by default

vubalasu: yes we can do I assert the number of requests but it will be cleaner to understand on first request and after clicking sort to check so easily i can undestand

richpjames: Under what circumstances could the assertion on line 89 pass but the one on 97 fail?

richpjames (approved):

PR #1813, opened 2024-05-09
E2Es: Correct deadline assertion

richpjames:
Trying to manually calculate if the emergency application was due today or tomorrow was prone to error. Now we use a more fuzzy date range it can be avoided


davidatkinsuk (approved):
LGTM 👍

PR #1814, opened 2024-05-09
APS 736 - Allow document on RfPs to be nullable

richpjames:
We are seeing [this error](https://ministryofjustice.sentry.io/issues/5325707450/?environment=preprod&project=4503931667742720&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=90d&stream_index=0) when rendering the RfPs pages as the document can be nullable but the UI code doesn't allow for this. 

We also add a 'Withdrawal Reason' field for when a RfP is withdrawn

## Before 
image

## After
image


davidatkinsuk (approved):
LGTM 👍

PR #1815, opened 2024-05-09
APS 736 - Remove flaky assertion

richpjames:
The rows array can include other properties dependent on status but this is the only one we're interested in


davidatkinsuk (approved):
LGTM 👍

PR #1819, opened 2024-05-13
Make A1 tiers eligible for AP without explanation

richpjames:
Previously applicants for people with A1 tiers were told that the tier was ineligible and would have to explain why their's was an exceptional case. However A1s is an eligible case and shouldn't require explanation.

### DIFF
--- a/server/utils/personUtils.ts
+++ b/server/utils/personUtils.ts
@@ -9,7 +9,7 @@ const tierBadge = (tier: string): string => {
 }
 
 const isApplicableTier = (sex: string, tier: string): boolean => {
-  const applicableTiersAll = ['A3', 'A2', 'B1', 'B3', 'B2', 'B1']
+  const applicableTiersAll = ['A1', 'A3', 'A2', 'B1', 'B3', 'B2']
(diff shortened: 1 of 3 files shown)


davidatkinsuk (approved):
LGTM 👍

server/utils/personUtils.ts
```diff
@@ -9,7 +9,7 @@ const tierBadge = (tier: string): string => {
 }
 
 const isApplicableTier = (sex: string, tier: string): boolean => {
-  const applicableTiersAll = ['A3', 'A2', 'B1', 'B3', 'B2', 'B1']
+  const applicableTiersAll = ['A1', 'A3', 'A2', 'B1', 'B3', 'B2']
```
edavey, line 12: amazing that we hadn't noticed that B1 was repeated! 🤦

richpjames: Good idea Ed

richpjames: The jumbled order of the tiers in the array made the duplicate hard to spot so I have put them in order now

PR #1836, opened 2024-05-16
Add ADR for using roles to feature switch manage functionality

richpjames:


doc/architecture/decisions/0014-use-a-new-role-to-feature-flag-out-of-service-bed-functionality.md
```diff
@@ -0,0 +1,25 @@
...
+
+We will add a new role to enable the Out of Service Beds functionality for specified users
+
+## Consequences
+
+- We could enable the role only in the API and not show it in the User Management UI
+- We could show it to all users in the User Management UI
+- We could add a feature switch to only allow the option to give users the role in the User Management UI in dev and preprod only
+- We will be able to have more granular control over who sees the new functionality than an on/off flag would allow
+- We could test in prod for users who have the role set
```
edavey, line 25: Do we need a little more detail here? Am I correct to think that you are highlighting 3 options here:

1. We could enable the role only in the API and not show it in the User Management UI
2. We could show it to all users in the User Management UI
3. We could add a feature switch to only allow the option to give users the role in the User Management UI in dev and preprod only

and that the "consequences" of this will be that:

> - We will be able to have more granular control over who sees the new functionality than an on/off flag would allow
> - We could test in prod for users who have the role set

Is that right? Perhaps we should talk through and add a little more detail?

richpjames: Yes that's correct. The consequences of the decision are: 
```
- We will be able to have more granular control over who sees the new functionality than an on/off flag would allow
- We could test in prod for users who have the role set 
```
And the decision enables the three options you've highlighted above. 
Happy to talk it through and add some more detail

richpjames:
@edavey I feel like a discussion of Flipt is redundant now that there is talk of deprecating the Flipt/feature flag service

PR #1843, opened 2024-05-20
APS-819 Roll back 'applicationinsights' package to v2

edavey:
See Slack thread: https://mojdt.slack.com/archives/C03K0HB0LBE/p1716222154095039 and
https://mojdt.slack.com/archives/C01AHT2AMCJ/p1714749514824029

On Apr 18 the renovate bot upgraded applicationinsights from `2.9.5` to `3.0.0`
https://github.com/ministryofjustice/hmpps-approved-premises-ui/commit/13256536196b136b187b9f1e9c55398357e5a790

It seems that v3 of applicationinsights does not play nicely
with the HMPPS "Typescript template" and we can't easily view
requests by `cloud_RoleName` e.g.

```
requests
| where cloud_RoleName == "approved-premises-ui"
```

See this revert of May 3 on the HMPPS "typescript template":
https://github.com/ministryofjustice/hmpps-template-typescript/pull/353

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,7 @@
         "@sentry/tracing": "^7.14.1",
         "@total-typescript/shoehorn": "^0.1.0",
         "agentkeepalive": "^4.2.1",
-        "applicationinsights": "^3.0.0",
+        "applicationinsights": "^2.9.5",
         "body-parser": "^1.20.0",
         "bunyan": "^1.8.15",
         "bunyan-format": "^0.2.1",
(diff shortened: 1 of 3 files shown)


richpjames (approved):
Makes sense. It would be good to have some alerting if this breaks in future. Is this possible in App Insights?

edavey:
> Makes sense. It would be good to have some alerting if this breaks in future. Is this possible in App Insights?

I don't there there's an actual error or exception here, rather some incompatibility with the Typescript Template. I don't think it's going to be possible to use AppInsights to tell use when there's a problem using AppInsights! Too meta 🤯

richpjames:
@edavey I meant an alert on a lack of data as opposed to an exception EG 'if we don't receive any requests with the `cloud_RoleName` for 48 hours raise an alert'

edavey:
> @edavey I meant an alert on a lack of data as opposed to an exception EG 'if we don't receive any requests with the `cloud_RoleName` for 48 hours raise an alert'

yes that might be possible in AppInsights...

PR #1844, opened 2024-05-20
APS-690: Don't show 'requests for placement' if user doesn't have WOR…

vubalasu:
…KFLOW_MANAGER or MATCHER permissions

# Context

# Changes in this PR
Don't show 'requests for placement' if user doesn't have WORKFLOW_MANAGER or MATCHER permissions on assess
### Before
image

### After
Screenshot 2024-05-20 at 17 42 06

### DIFF
--- a/integration_tests/tests/assess/list.cy.ts
+++ b/integration_tests/tests/assess/list.cy.ts
@@ -7,7 +7,7 @@ context('List assessments', () => {
   beforeEach(() => {
     cy.task('reset')
     cy.task('stubSignIn')
-    cy.task('stubAuthUser')
+    cy.task('stubAuthUser', { roles: ['workflow_manager'] })
   })
 
   it('should list assessments', () => {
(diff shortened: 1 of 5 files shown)


server/utils/placementApplications/utils.ts
```diff
@@ -0,0 +1,36 @@
+import { UserDetails } from '../../@types/ui'
+import { createQueryString } from '../utils'
+import { TabItem } from '../tasks/listTable'
+import assessPaths from '../../paths/assess'
+import { hasRole } from '../users'
+
+export const assessmentsTabItems = (user: UserDetails, activeTab?: string): Array<TabItem> => {
```
richpjames, line 7: Other than in the tests where is this function used?

vubalasu: some issue didnt pushed some files rebased and pushed again

richpjames, line 7: Given this util is called `assessmentTabItems` and it is used solely on the `/assessments` page I think it would be clearer if it was in the `/server/utils/assessments` folder

richpjames (changes requested):

PR #1847, opened 2024-05-22
APS-812: If users navigates away from application when told tier not …

vubalasu:
…eligible, if they return they're always prompted to select risk level

# Context

# Changes in this PR
If users navigates away from application when told tier not eligible, if they return they're always prompted to select risk level
## Screenshots of UI changes
No changes in UI

### DIFF
--- a/server/utils/applications/utils.ts
+++ b/server/utils/applications/utils.ts
@@ -206,7 +206,7 @@ const isInapplicable = (application: Application): boolean => {
   return false
 }
 
-const firstPageOfApplicationJourney = (application: Application) => {
+const tierQualificationPage = (application: Application) => {
   if (!isFullPerson(application.person)) throw new RestrictedPersonError(application.person.crn)
 
   if (!application?.risks?.tier?.value) {
@@ -215,6 +215,15 @@ const firstPageOfApplicationJourney = (application: Application) => {
 
   if (!isApplicableTier(application.person.sex, application.risks?.tier?.value?.level)) {
     return paths.applications.pages.show({ id: application.id, task: 'basic-information', page: 'is-exceptional-case' })
+  }
+  return undefined
+}
+
+const firstPageOfApplicationJourney = (application: Application) => {
+  const firstPage = tierQualificationPage(application)
+
+  if (firstPage) {
+    return firstPage
   }
 
   return paths.applications.pages.show({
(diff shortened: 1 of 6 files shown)


server/utils/taskListUtils.test.ts
```diff
@@ -15,6 +17,8 @@ describe('taskListUtils', () => {
   describe('taskLink', () => {
     describe('with an application', () => {
       const application = applicationFactory.build({ id: 'some-uuid' })
+      ;(isFullPerson as jest.MockedFunction<typeof isFullPerson>).mockReturnValue(true)
```
richpjames, line 20: I think it's best we put lines 20 and 21 in a `beforeEach` and make sure we do `jest.resetAllMocks()` before.

server/utils/taskListUtils.test.ts
```diff
@@ -68,6 +72,19 @@ describe('taskListUtils', () => {
           })}" aria-describedby="eligibility-second-task" data-cy-task-name="second-task">Second Task</a>`,
         )
       })
+
+      it('should return a link of is-exceptional-case if risk tier not applicable', () => {
+        task.status = 'in_progress'
+        ;(isApplicableTier as jest.MockedFunction<typeof isApplicableTier>).mockReturnValue(false)
```
richpjames, line 78: it's clearer to use `when` like here: https://github.com/ministryofjustice/hmpps-approved-premises-ui/blob/78c23861471b7b60383d311daf6777907bc170f1/server/controllers/tasksController.test.ts#L67

server/utils/applications/utils.ts
```diff
@@ -206,7 +206,7 @@ const isInapplicable = (application: Application): boolean => {
   return false
 }
 
-const firstPageOfApplicationJourney = (application: Application) => {
+const firstPageOfApplicationJourneyBasedOnTier = (application: Application) => {
```
richpjames, line 209: I think `tierQualificationPage` would be clearer. 
Also this function should be tested

richpjames (changes requested):
I think these changes would now allow the user to skip the `is-exceptional-case` page if they navigate away from the application page when told the tier is not eligible.
It would be useful if we added an integration test to prove the bug is fixed.

richpjames (approved):
Thanks for adding the tests, this looks good to merge now once the unit tests are passing.

PR #1855, opened 2024-05-24
APS-832 -  Version 'Manage' functionality

richpjames:
This PR builds on #1850 
# Context

We need to introduce changes to the existing 'Manage' functionality.
These new changes will be introduced under the `/manage/` URL namespace.
This PR makes the necessary functionality available under the namespace.
For now we just add new routes to the existing functionality (apart from 'Out of Service Beds' which was duplicated in #1850). 
We add new integration tests to test the new routes and ensure they work as expected.

[JIRA](https://dsdmoj.atlassian.net/jira/software/c/projects/APS/boards/1328?assignee=62a050ba9f5d480069c97f49&selectedIssue=APS-832&sprints=6188)


edavey (approved):
👍 , I think the "v2 manage" works well to clearly partition the new out-of-service-bed code

PR #1857, opened 2024-05-24
APS-785: Improve UX when advancing through the 5 OASys steps for risk

vubalasu:
# Context

# Changes in this PR
 Improve UX when advancing through the 5 OASys steps for risk
### Before
Screenshot 2024-05-24 at 13 35 46

### After
Screenshot 2024-05-23 at 15 12 42
Screenshot 2024-05-23 at 15 12 51
Screenshot 2024-05-23 at 15 12 58
Screenshot 2024-05-23 at 15 13 05
Screenshot 2024-05-23 at 15 14 10

### DIFF
--- a/server/views/applications/pages/oasys-import/_oasys-screen.njk
+++ b/server/views/applications/pages/oasys-import/_oasys-screen.njk
@@ -17,7 +17,8 @@
       {% if oasysDisabled or page.oasysSuccess == false %}
         <h1 class="govuk-heading-l">Provide risk information</h1>
       {% else %}
-        <h1 class="govuk-heading-l">{{page.title}}</h1>
+        <h1 class="govuk-heading-l govuk-!-margin-bottom-2">{{page.title}}</h1>
+        <p>Risk Information (part {{page.riskTaskStep}} of 5)</p>
         <p>OASys last updated: {{formatDate(page.oasysCompleted)}}</p>
       {% endif %}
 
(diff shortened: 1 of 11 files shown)


server/views/applications/pages/oasys-import/_oasys-screen.njk
```diff
@@ -17,7 +17,8 @@
       {% if oasysDisabled or page.oasysSuccess == false %}
         <h1 class="govuk-heading-l">Provide risk information</h1>
       {% else %}
-        <h1 class="govuk-heading-l">{{page.title}}</h1>
+        <h1 class="govuk-heading-l govuk-!-margin-bottom-2">{{page.title}}</h1>
+        <p>{{page.riskTaskStep}}</p>
```
richpjames, line 21: If `riskTaskStep` was an integer then in the template we could have:
`<p>Risk Information (part {{page.riskTaskStep}} of 5)</p>`
so there would be less places to update if we ever need to change the copy (EG if a page was added/removed).

richpjames: Thanks for making the edit I suggested in the OasysPage classes but now `page.riskTaskStep` is an integer we have to update this line to be `<p>Risk Information (part {{page.riskTaskStep}} of 5)</p>`

vubalasu: yah sorry the commit push failed because of token expired and pushed again please check @richpjames

richpjames (approved):

PR #1859, opened 2024-05-24
Add call to init Sentry

richpjames:
When upgrading Sentry (#1832) I missed this.
This copies the approach used:
https://github.com/ministryofjustice/probation-search-ui


davidatkinsuk (approved):
LGTM 👍

PR #1860, opened 2024-05-24
Revert Sentry upgrade

richpjames:
#1823 and #1859 upgraded @sentry/node to v8 but I wasn't able to see exceptions coming through to Sentry in dev since the upgrade. I have spent some time looking into it but haven't been able to figure it out. Given current time pressures I can't dedicate any more time to it so I think it's best to revert the changes for now and upgrade at a later date.


davidatkinsuk (commented):
LGTM 👍

davidatkinsuk (approved):

PR #1864, opened 2024-05-28
APS 836 Introduce new 'Manage' permissions

richpjames:
# Context

In order to show new Manager content to selected users and preserve the existing Manage content for a different group of selected users we need to introduce two new roles legacy_manager and future_manager. 
Where we currently show the Premises ‘actions’ and Booking ‘actions’ we will use these new roles to determine which actions to make available to the users. 

[JIRA](https://dsdmoj.atlassian.net/jira/software/c/projects/APS/boards/1328?assignee=62a050ba9f5d480069c97f49&selectedIssue=APS-836&sprints=6188%2C6585)

# Changes in this PR

### Before
image

image


server/utils/bookings/index.ts
```diff
@@ -175,6 +175,23 @@ export const bookingActions = (user: UserDetails, booking: Booking): Array<Ident
...
+      {
+        items: [
+          {
+            text: 'Withdraw placement',
+            classes: 'govuk-button--secondary',
+            href: applyPaths.applications.withdraw.new({ id: booking.applicationId }),
+          },
+        ],
+      },
+    ]
```
edavey, line 190: Quite likely I'm getting confused, but as the v2BookingActions will only be available to users with the `FUTURE_MANAGER` role what's the reason to limit the actions? 

We will ultimately be offering (almost) all the current features:

- update departure dates
- log a departure
- ~~withdraw an arrived placement~~
- move a person to a new bed
- log an arrival
- log a non-arrival

Perhaps you anticipate that these features will be significantly different in v2 manage and so prefer to start from a blank slate and build up incrementally as we port enhanced features over?

edavey (approved):
Nice work, v clear. The only thing that I don't follow properly is why we're removing all the actions from the v2BookingActions.

server/utils/premises/index.ts
```diff
@@ -3,14 +3,15 @@ import type {
...
+  Premises,
   ApprovedPremisesSummary as PremisesSummary,
 } from '@approved-premises/api'
-import { BedOccupancyRangeUi, SelectGroup, SummaryList } from '@approved-premises/ui'
-import { DateFormats } from './dateUtils'
-import { addOverbookingsToSchedule } from './addOverbookingsToSchedule'
-import { htmlValue, textValue } from './applications/helpers'
-import paths from '../paths/manage'
-import { linkTo } from './utils'
+import { BedOccupancyRangeUi, SelectGroup, SummaryList, UserDetails } from '@approved-premises/ui'
```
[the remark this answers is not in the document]
richpjames: yep, good spot

server/utils/bookings/index.ts
richpjames: Yes, your suspicion is correct. I think we are best copying over these features piece by piece as and when they are needed. 
If we copy them over now there will be a lot of duplicated code which we will have to alter slightly to fit the new approach which we will then have to test and support.
It should be easy enough to add these features when the time comes but for now I don't think it's worth us spending time on them when they *could* change drastically in the future.

edavey (approved):
🚀 nice work, thanks for working through this so carefully

PR #1875, opened 2024-05-31
APS-760:Add validation to index offence selection screen

vubalasu:
# Context

# Changes in this PR
Add validation to index offence selection screen
### Before
image

### After
Screenshot 2024-05-30 at 14 50 12

### DIFF
--- a/server/controllers/apply/applicationsController.ts
+++ b/server/controllers/apply/applicationsController.ts
@@ -13,6 +13,7 @@ import { getPaginationDetails } from '../../utils/getPaginationDetails'
 import { ApplicationSortField } from '../../@types/shared'
 import { ApplicationDashboardSearchOptions } from '../../@types/ui'
 import { getSearchOptions } from '../../utils/getSearchOptions'
+import { RestrictedPersonError } from '../../utils/errors'
 
 export const tasklistPageHeading = 'Apply for an Approved Premises (AP) placement'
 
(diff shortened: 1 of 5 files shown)


server/controllers/apply/applicationsController.ts
```diff
@@ -165,7 +166,14 @@ export default class ApplicationsController {
     return async (req: Request, res: Response) => {
       const { crn, offenceId } = req.body
 
+      const person = await this.personService.findByCrn(req.user.token, crn)
+      if (!isFullPerson(person)) throw new RestrictedPersonError(crn)
       const offences = await this.personService.getOffences(req.user.token, crn)
+      if (!offenceId) {
+        addErrorMessageToFlash(req, 'You must select the offence', 'offenceId')
```
edavey, line 173: perhaps 

> You must select the index offence

would be more accurate?

server/controllers/apply/people/offencesController.test.ts
```diff
@@ -45,6 +47,8 @@ describe('OffencesController', () => {
 
       expect(response.render).toHaveBeenCalledWith('applications/people/selectOffence', {
         pageHeading: `Select index offence for ${person.name}`,
+        errors: {},
+        errorSummary: [],
```
edavey, line 51: are we actually describing the unhappy path as well as the happy path?

I was expecting to see a scenario along the lines of 

```ts
expect(response.render).toHaveBeenCalledWith('applications/people/selectOffence', {
        ...
        errors: errorsAndUserInput.errors,
        errorSummary: errorsAndUserInput.errorSummary,
        ...errorsAndUserInput.userInput,
      })
```

edavey (commented):
Thanks for this. A couple of comments:

- a minor suggestion on the error message and 
- not sure that the error case is clearly described in the controller test?

server/controllers/apply/applicationsController.ts
vubalasu: updated

server/controllers/apply/people/offencesController.test.ts
vubalasu: @edavey added

edavey (approved):
thanks for the tweaks

PR #1878, opened 2024-06-03
APS-810:502 error gateway if submitting person timeline multiple time…

vubalasu:
…s with no CRN

# Context

# Changes in this PR
Fix 502 error gateway if submitting person timeline multiple times with no CRN
Person timeline crn search gives error on second time submit 
the issue is on TimelineController redirect with out return and so it is updating the response headers after repose send to client
## Screenshots of UI changes
No changes in UI

### DIFF
--- a/server/controllers/people/timelineController.ts
+++ b/server/controllers/people/timelineController.ts
@@ -28,16 +28,15 @@ export default class TimelineController {
 
       if (!crn?.trim()) {
         addErrorMessageToFlash(req, 'You must enter a CRN', 'crn')
-        res.redirect(paths.timeline.find({}))
+        return res.redirect(paths.timeline.find({}))
       }
 
       try {
         const timeline = await this.personService.getTimeline(req.user.token, crn.trim())
-
-        res.render('people/timeline/show', { timeline, crn, pageHeading: `Timeline for ${crn}` })
+        return res.render('people/timeline/show', { timeline, crn, pageHeading: `Timeline for ${crn}` })
       } catch (error) {
         crnErrorHandling(req, error, crn)
-        res.redirect(paths.timeline.find({}))
+        return res.redirect(paths.timeline.find({}))
       }
     }
   }


server/controllers/people/timelineController.ts
```diff
@@ -28,16 +28,16 @@ export default class TimelineController {
...
 
       try {
         const timeline = await this.personService.getTimeline(req.user.token, crn.trim())
 
-        res.render('people/timeline/show', { timeline, crn, pageHeading: `Timeline for ${crn}` })
+        return res.render('people/timeline/show', { timeline, crn, pageHeading: `Timeline for ${crn}` })
       } catch (error) {
         crnErrorHandling(req, error, crn)
-        res.redirect(paths.timeline.find({}))
+        return res.redirect(paths.timeline.find({}))
```
edavey, line 40: thanks Bala, do you think you could:

- add into the commit message an explanation of the problem and how this fixes it?
- include a unit test to document this in the code and prevent regression?

vubalasu: @edavey 
add into the commit message an explanation of the problem and how this fixes it?
updated the commit message
include a unit test to document this in the code and prevent regression?
502 error gateway if submitting person timeline multiple times with no CRN
Person timeline crn search gives error on second time submit 
the issue is on TimelineController redirect with out return and so it is updating the response headers after repose send to client
Normal unit test is already there and
is not possible to write a unit test especially for this scenario

edavey (approved):
I'm afraid I'm still unclear on the issue and the fix and I think it's essential that you add a clear description of both to the *commit message* as distinct from the PR conversation.

The commit history is our primary resource for understanding *why* the code has been changed.

See:

- https://github.com/alphagov/styleguides/blob/master/git.md
- https://tekin.co.uk/2018/07/resources-for-creating-useful-revision-histories
- https://blog.mocoso.co.uk/posts/talks/telling-stories-through-your-commits/

PR #1880, opened 2024-06-05
APS-822: Populate 'reasonForShortNotice' and 'reasonForShortNoticeOth…

vubalasu:
…er' on app submission

# Context

# Changes in this PR
 Populate 'reasonForShortNotice' and 'reasonForShortNoticeOther' on app submission
## Screenshots of UI changes
No ui changes

### DIFF
--- a/server/utils/applications/getApplicationData.ts
+++ b/server/utils/applications/getApplicationData.ts
@@ -22,6 +22,7 @@ import { noticeTypeFromApplication } from './noticeTypeFromApplication'
 import Situation from '../../form-pages/apply/reasons-for-placement/basic-information/situation'
 import ConfirmYourDetails from '../../form-pages/apply/reasons-for-placement/basic-information/confirmYourDetails'
 import { applicantAndCaseManagerDetails } from './applicantAndCaseManagerDetails'
+import { reasonForShortNoticeDetails } from './reasonForShortNoticeDetails'
 
 type FirstClassFields<T> = T extends UpdateApprovedPremisesApplication
   ? Omit<UpdateApprovedPremisesApplication, 'data'>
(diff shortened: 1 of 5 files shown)


server/utils/applications/getApplicationData.ts
```diff
@@ -79,6 +81,14 @@ const firstClassFields = <T>(
     caseManagerIsNotApplicant,
     noticeType,
   } as FirstClassFields<T>
+
+  const startDateOutsideOfTimescales = startDateOutsideOfNationalStandardsTimescales(application)
+  if (startDateOutsideOfTimescales) {
+    const reasonForShortNotice = retrieveQuestionResponse(application, ReasonForShortNotice, 'reason')
```
richpjames, line 87: I think it would be better if we followed the same pattern as all the other properties in the object that is returned from this function.  
We should create a util function that returns `reasonForShortNotice` and `reasonForShortNoticeOther` and add this to the return value. `applicationAndCaseManagerDetails` is a good example of this.

richpjames (changes requested):

server/utils/applications/reasonForShortNoticeDetails.ts
```diff
@@ -0,0 +1,24 @@
...
+export const reasonForShortNoticeDetails = (
+  application: Application,
+): {
+  reasonForShortNotice: string
+  reasonForShortNoticeOther: string
+} => {
+  const startDateOutsideOfTimescales = startDateOutsideOfNationalStandardsTimescales(application)
+  let reasonForShortNotice
+  let reasonForShortNoticeOther
+  if (startDateOutsideOfTimescales) {
```
richpjames, line 15: I don't think we need to check if the start date is outside of national timescales
`retrieveOptionalQuestionResponseFromFormArtifact` will return undefined if the start date is inside of national timescales.

vubalasu: there may be scenario where first the releasedate is OutsideOfNationalStandardsTimescales and entered reasonForShortNotice and later chnaged releasedate and it is not OutsideOfNationalStandardsTimescales but still it will send reasonForShortNotice if we dont have this check startDateOutsideOfNationalStandardsTimescales

richpjames: Yes, very good point.

richpjames (approved):

PR #1887, opened 2024-06-07
APS 852  - Use new OOS beds type definitions

richpjames:
# Context

[jira](https://dsdmoj.atlassian.net/jira/software/c/projects/APS/boards/1328?assignee=62a050ba9f5d480069c97f49&assignee=712020%3A23fbf793-ee9a-4ecd-ad4d-37beaa0edeaa&selectedIssue=APS-858&sprints=6188%2C6585) 
#1850 introduced some stop-gap changes that allowed us to make a start on the v2 Out Of Service bed functionality without having to wait for the API to have the requisite changes. Now the API is ready we can remove these temporary changes and implement a more permanent solution.

# Changes in this PR
- We introduce the new OOS bed entities and adapt the code to use them
- Use the new OOS bed endpoints in the OOS bed client
- Remove the temporary OOS bed types

### DIFF
--- a/integration_tests/mockApis/outOfServiceBed.ts
+++ b/integration_tests/mockApis/outOfServiceBed.ts
@@ -33,7 +33,7 @@ export default {
       },
     }),
 
-  stubUpdateOutOfServiceBedErrors: ({ outOfServiceBed, premisesId }): SuperAgentRequest =>
+  stubUpdateOutOfServiceBedErrors: ({ outOfServiceBed, premisesId, params }): SuperAgentRequest =>
     stubFor({
       request: {
         method: 'PUT',
@@ -46,12 +46,7 @@ export default {
           type: 'https://example.net/validation-error',
           title: 'Invalid request parameters',
           code: 400,
-          'invalid-params': [
-            {
-              propertyName: `$.endDate`,
-              errorType: 'empty',
-            },
-          ],
+          'invalid-params': params,
         },
       },
     }),
--- a/server/testutils/factories/outOfServiceBed.ts
+++ b/server/testutils/factories/outOfServiceBed.ts
@@ -1,22 +1,36 @@
 import { Factory } from 'fishery'
 import { faker } from '@faker-js/faker/locale/en_GB'
+import {
+  NamedId,
+  NewCas1OutOfServiceBed as NewOutOfServiceBed,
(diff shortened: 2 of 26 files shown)


server/testutils/factories/outOfServiceBed.ts
```diff
@@ -1,22 +1,36 @@
...
+  reason: referenceDataFactory.outOfServiceBedReason().build(),
+  referenceNumber: faker.helpers.arrayElement([faker.string.uuid(), undefined]),
+  notes: faker.helpers.arrayElement([faker.lorem.sentence(), undefined]),
+  daysLostCount: faker.number.int(),
+  temporality: faker.helpers.arrayElement(['past', 'current', 'future'] as const),
+  status: faker.helpers.arrayElement(['active', 'cancelled'] as const),
+  cancellation: undefined,
+}))
+
+const namedIdFactory = Factory.define<NamedId>(() => ({
```
patrickjfl, line 31: minor but is this supposed to be `NameId`?

integration_tests/mockApis/outOfServiceBed.ts
```diff
@@ -46,12 +46,7 @@ export default {
           type: 'https://example.net/validation-error',
           title: 'Invalid request parameters',
           code: 400,
-          'invalid-params': [
-            {
-              propertyName: `$.endDate`,
-              errorType: 'empty',
-            },
-          ],
+          'invalid-params': params,
```
patrickjfl, line 49: nice

server/testutils/factories/outOfServiceBed.ts
richpjames: Nah I think `NamedId` is correct IE an ID with a name. Is that right @danlivings-dxw ?

patrickjfl: oh, I didn't realise this was an API type. still a bit puzzling IMO, but not a big deal

danlivings-dxw: `NamedId` is correct - the thinking behind why it's called that is that it's an ID of something that was named, rather than it being explicitly a name-ID pair, even though it's represented by one.

richpjames: I like it !

PR #1894, opened 2024-06-10
Allow CRU admin to view out of service beds

patrickjfl:
# Context

Jira ticket: https://dsdmoj.atlassian.net/jira/software/c/projects/APS/boards/1328?assignee=712020%3A23fbf793-ee9a-4ecd-ad4d-37beaa0edeaa&selectedIssue=APS-858&sprints=6188

# Changes in this PR

This PR adds an index view to allow users with the `future_manager` to view all out of service beds across all Approved Premises. It also adds a tile to the dashboard page for users with this role, to allow them to navigate to the index page.

## Screenshots of UI changes

Screenshot 2024-06-10 at 09 40 18

Screenshot 2024-06-10 at 09 40 54


integration_tests/tests/v2Manage/beds.cy.ts
```diff
@@ -47,4 +47,24 @@ context('Beds', () => {
     // Then I am taken to the mark bed out of service page
     Page.verifyOnPage(OutOfServiceBedCreatePage)
   })
+
+  it('should allow Future Manager to manage out of service beds from the bed list page', () => {
```
richpjames, line 51: I think this has snuck into the wrong commit

cypress.config.ts
```diff
@@ -67,5 +67,7 @@ export default defineConfig({
     excludeSpecPattern: '**/!(*.cy).ts',
     specPattern: 'integration_tests/tests/**/*.cy.{js,jsx,ts,tsx}',
     supportFile: 'integration_tests/support/index.ts',
+    viewportHeight: 800,
+    viewportWidth: 1020,
```
richpjames, line 71: I think it's worth noting in the commit message that this is the default maximum page width as specified here: https://design-system.service.gov.uk/styles/layout/

server/paths/manage.ts
```diff
@@ -98,6 +98,7 @@ const v2BedsPath = v2SinglePremisesPath.path('beds')
 const v2DateChangesPath = v2BookingPath.path('date-changes')
 const v2ExtensionsPath = v2BookingPath.path('extensions')
 const outOfServiceBedsPath = v2SinglePremisesPath.path('beds/:bedId/out-of-service-beds')
+const outOfServiceBedsIndexPath = path('/manage/out-of-service-beds')
```
richpjames, line 101: I think it would be worth creating a new path `const v2ManagePath = path('/manage')` and using it for this new path and `v2PremisesPath` - it will help if we ever need to change the path.

integration_tests/tests/dashboard.cy.ts
```diff
@@ -28,17 +28,6 @@ context('Dashboard', () => {
     dashboardPage.shouldNotShowCard('manage')
   })
```
richpjames, line 30: @edavey it would be good to hear your feedback on this. We didn't think that a user without roles should be able to access 'Apply'.

server/testutils/jest.setup.ts
```diff
@@ -26,6 +26,7 @@ const apiSpecs = {
     command: (openAPIUrl: string) => `if [ ! -f ${apiSpecPaths.cas1Spec} ]; then
     curl -s "${openAPIUrl}" |
     sed -E 's@/premises@/cas1/premises@g' |
+    sed -E 's@ /out-of-service-beds@ /cas1/out-of-service-beds@g' |
```
richpjames, line 29: Not a big deal but I would add this change to the commit where we add the new client method

richpjames (commented):
Looks good on the whole, have left some minor comments

server/paths/manage.ts
patrickjfl: good shout

server/testutils/jest.setup.ts
patrickjfl: ok yeah, have done

integration_tests/tests/dashboard.cy.ts
richpjames: We agreed on a call with @edavey  that all users including those without roles should be able to access 'Apply' as having a Delius account (and therefore access to the service) is sufficient.

richpjames (approved):

PR #1895, opened 2024-06-10
Handle bookings without applicationIds

richpjames:
This was breaking in dev when a booking didn't have an applicationId. The property access was valid but static-path was erroring as it can't create a path where id === undefined.


patrickjfl (approved):
nice fix 👍

PR #1896, opened 2024-06-10
Add roles for manage.spec.ts

richpjames:
These tests were failing because the test users were given all the roles in dev. However 'View all out of service beds' requires the 'future_manager' role to pass and 'Mark a booking as cancelled' requires the user *does not* have the 'future_manager' role to pass. This is because in the `bookingActions` function we will return the v2BookingActions if the user has the 'future_manager' role and the cancellation test neccessitates the 'legacyBookingActions' function. Now we set the roles before each test as we do for the apply and assess tests.


patrickjfl (approved):
nice job 🚀

PR #1897, opened 2024-06-11
APS-840: Use /applications/{applicationId}/withdrawablesWithNotes and…

vubalasu:
… show notes in UI

# Context

# Changes in this PR
(UI) Use /applications/{applicationId}/withdrawablesWithNotes and show notes
### Before
image
image

### After
Screenshot 2024-06-11 at 09 17 49
Screenshot 2024-06-11 at 09 21 34

### DIFF
--- a/server/controllers/apply/withdrawablesController.test.ts
+++ b/server/controllers/apply/withdrawablesController.test.ts
@@ -10,6 +10,7 @@ import placementAppPaths from '../../paths/placementApplications'
 import { Withdrawable } from '../../@types/shared'
 import applyPaths from '../../paths/apply'
 import { sortAndFilterWithdrawables } from '../../utils/applications/withdrawables'
+import withdrawablesFactory from '../../testutils/factories/withdrawablesFactory'
 
 jest.mock('../../utils/applications/withdrawables')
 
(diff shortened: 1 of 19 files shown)


server/testutils/factories/withdrawablesFactory.ts
```diff
@@ -0,0 +1,10 @@
+import { Factory } from 'fishery'
+import { Withdrawables } from '../../@types/shared'
+
+export default Factory.define<Withdrawables>(() => ({
+  notes: [
+    '1 or more placements cannot be withdrawn as they have an arrival',
+    '1 or more placements cannot be withdrawn as they have an arrival recorded in Delius',
+  ],
+  withdrawables: [],
```
richpjames, line 9: We should add some withdrawables here rather than having an empty array then we don't have to separately create the withdrawables every time we use this factory.

server/views/applications/withdrawables/new.njk
```diff
@@ -19,6 +19,11 @@
     <div class="govuk-grid-row">
         <div class="govuk-grid-column-two-thirds">
             {% if withdrawables.length %}
+                <ul class="govuk-list govuk-list--bullet">
```
richpjames, line 22: We should check that the notes exist before creating the `ul` otherwise we will be adding an empty list to the markup

server/views/applications/withdrawals/new.njk
```diff
@@ -25,6 +25,12 @@
 
         {{ showErrorSummary(errorSummary) }}
 
+          <ul class="govuk-list govuk-list--bullet">
```
richpjames, line 28: We should check that the notes exist before creating the `ul` otherwise we will be adding an empty list to the markup

server/controllers/apply/withdrawablesController.test.ts
```diff
@@ -41,13 +42,14 @@ describe('withdrawablesController', () => {
         const placementRequestWithdrawable = withdrawableFactory.build({ type: 'placement_request' })
         const placementApplicationWithdrawable = withdrawableFactory.build({ type: 'placement_application' })
         const applicationWithdrawable = withdrawableFactory.build({ type: 'application' })
-        const withdrawables = [placementRequestWithdrawable, placementApplicationWithdrawable, applicationWithdrawable]
+        const withdrawable = [placementRequestWithdrawable, placementApplicationWithdrawable, applicationWithdrawable]
```
richpjames, line 45: This should to remain `withdrawbles` given there are multiple withdrawables here

richpjames (commented):
Do these changes mean that the `applicationService.getWithdrawables` and `applicationClient.withdrawables` methods are no longer used? if so, could we delete them?

richpjames (approved):
LGTM

PR #1900, opened 2024-06-11
Remove roles for manage in E2E tests

richpjames:
#1896 recognized the need to alter the roles for the manage E2E tests but did so incorrectly. 
To fix this we need to skip the OoS bed test. It requires the role 'Future manager' to pass but giving the test user this role breaks the other manage tests. We have not yet been able to work out a way to set the roles programatically. As it is not currently user facing we can skip for now to unblock the pipelines
We then remove the roles from the manage tests which will be set manually in the UI in dev.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -179,7 +179,42 @@ jobs:
           channel: << pipeline.parameters.alerts-slack-channel >>
           template: basic_fail_1
 
-  e2e_environment_test:
+  e2e_environment_test_on_pr:
+    executor:
+      name: hmpps/node
+      tag: << pipeline.parameters.node-version >>
+    parallelism: 4
+    circleci_ip_ranges: true # opt-in to jobs running on a restricted set of IPs
+    steps:
+      - checkout
+      - attach_workspace:
+          at: ~/app
+      - restore_cache:
+          key: dependency-cache-{{ checksum "package-lock.json" }}
+      - run:
+          name: Install Playwright
+          command: npx playwright install
+      - run:
+          name: E2E Check
+          command: |
+            SHARD="$((${CIRCLE_NODE_INDEX}+1))"
+            username="HMPPS_AUTH_USERNAME_$SHARD"
+            password="HMPPS_AUTH_PASSWORD_$SHARD"
+            email="HMPPS_AUTH_EMAIL_$SHARD"
+            name="HMPPS_AUTH_NAME_$SHARD"
+            HMPPS_AUTH_USERNAME="${!username}"
+            HMPPS_AUTH_PASSWORD="${!password}"
+            HMPPS_AUTH_EMAIL="${!email}"
+            HMPPS_AUTH_NAME="${!name}"
+            npm run test:e2e:ci -- --shard=${SHARD}/${CIRCLE_NODE_TOTAL}
+      - store_artifacts:
+          path: playwright-report
+          destination: playwright-report
+      - store_artifacts:
+          path: test-results
+          destination: test-results
+
+  e2e_environment_test_on_merge:
     executor:
       name: hmpps/node
       tag: << pipeline.parameters.node-version >>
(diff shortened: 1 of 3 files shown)


e2e/tests/manage.spec.ts
```diff
@@ -261,15 +261,15 @@ test('Move a booking', async ({ page, person, user }) => {
...
-  // And I am on the dashboard page
-  const dashboard = await visitDashboard(page)
-
-  // And I click the 'View out of service beds' tile
-  dashboard.clickOutOfServiceBeds()
-
-  // Then I am taken to the out of service beds page
-  await OutOfServiceBedsPage.initialize(page, 'View out of service beds')
-})
+// test('View all out of service beds', async ({ page, user }) => {
```
edavey, line 258: could we use the `.skip` annotation here rather than commenting out? 

https://playwright.dev/docs/test-annotations#skip-a-test

That might be clearer in terms of what the runner reports? i.e. that we see explicitly what we we didn't run...

patrickjfl (commented):
The changes look good, but I'm getting a failure on 'download reports' when I run the tests using `npm run test:e2e:ci`. Is that the same for you?

patrickjfl:
> The changes look good, but I'm getting a failure on 'download reports' when I run the tests using `npm run test:e2e:ci`. Is that the same for you?

but then I can see they're passing here 🤷

patrickjfl (approved):

richpjames:
@patrickjfl thanks for the review. That test can be quite flaky historically. I reintroduced the E2Es to run on PRs again in this PR in order to prove that they will pass when running in CI. I think that environment is closer to the post-merge E2E check than what will be seen on local machines but I agree it would be nice to see them pass locally.

.circleci/config.yml
```diff
@@ -179,7 +179,42 @@ jobs:
           channel: << pipeline.parameters.alerts-slack-channel >>
           template: basic_fail_1
 
-  e2e_environment_test:
+  e2e_environment_test_on_pr:
```
edavey, line 182: just to be clear on what this means:  does this mean that the playwright specs in the PR are run against:

1.  the currently deployed version of the UI (the tip of the`main` branch) or
2. against a deployment of the UI code in the PR (feature branch)

I think it's 1? And that what we want is 2?

edavey (commented):
Probably (hopefully) I'm confused but when I use this branch to run the tests against the `dev` environment on my local machine with:

```
npx playwright test --config ./e2e/playwright.config.ts --project=dev
```

I get 13 failures.  What am I doing wrong here?

```
13 failed
    [dev] › tests/apply/appeal.spec.ts:12:5 › Record a successful appeal against a rejected application
    [dev] › tests/apply/appeal.spec.ts:21:5 › Record an unsuccessful appeal against a rejected application
    [dev] › tests/apply/emergency-application.spec.ts:7:5 › Apply, assess, match and book an emergency application for an Approved Premises
    [dev] › tests/apply/no-release-date.spec.ts:8:5 › Apply, assess, match and book an application for an Approved Premises without a release date
    [dev] › tests/apply/request-further-information.spec.ts:7:5 › Request further information on an Application, adds it and proceeds with the assessment
    [dev] › tests/apply/short-notice-application.spec.ts:7:5 › Apply, assess, match and book an short notice application for an Approved Premises with a release date
    [dev] › tests/apply/standard-application.spec.ts:7:5 › Apply, assess, match and book an application for an Approved Premises with a release date
    [dev] › tests/manage.spec.ts:97:5 › Mark a booking as cancelled ────────────────────────────────
    [dev] › tests/manage.spec.ts:120:5 › Change placement dates ────────────────────────────────────
    [dev] › tests/manage.spec.ts:175:5 › Mark a booking as arrived and extend it ───────────────────
    [dev] › tests/manage.spec.ts:208:5 › Mark a booking as not arrived ─────────────────────────────
    [dev] › tests/manage.spec.ts:233:5 › Move a booking ────────────────────────────────────────────
    [dev] › utils/createApplication.spec.ts:7:5 › create undefined application ─────────────────────
```

PR #1910, opened 2024-06-13
Don't set roles in E2E tests

richpjames:
We believe that the setting of these roles causes flakiness due to concurrent tests run interfering with each.
Most of these calls weren't doing anything. With the exception of the setting of 'Emergency APs' which can be set for the user via the UI. In the future we would like to move to a state where we have user profiles with static roles that can be used for each test but this is an interim step.


patrickjfl (approved):
OK, so if we give the user all the necessary roles in the UI, we don't actually need to set roles within the tests at all?

richpjames:
@patrickjfl correct

PR #1914, opened 2024-06-14
E2Es - Enable Apply and Assess tests locally

richpjames:
This PR includes various fixes to enable the Apply and Assess E2E tests to pass locally. 
The manage.spec.ts will still fail until changes are made to how key workers are seeded for APs to enable marking an arrival locally.
[This API PR](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1963) is required for the tests to pass too, but will be merged to main soon.
The changes to fix the tests are a number of small changes that are detailed in the commit messages


e2e/tests/auth.setup.ts
```diff
@@ -8,5 +9,6 @@ setup('authenticate', async ({ page, user }) => {
   await page.getByLabel('Password').fill(user.password)
   await page.getByRole('button', { name: 'Sign in' }).click()
 
+  expect(await page.title()).toBe('Approved Premises')
```
edavey, line 12: Would it be clearer if we wrap this up in an function which is named to express its purpose e.g.

`assertThatLoginSucceeded()`

e2e/tests/admin.spec.ts
```diff
@@ -100,19 +100,24 @@ test('manage users', async ({ page, userToAdd }) => {
...
-
-  // When I click on their name
-  await userListPage.clickEditUser(userToSearchFor)
-
-  // Then I should be taken to the Edit User page
-  await editUserPage.shouldShowUserName(userToSearchFor)
+  // The following behaviour only exists in CI
+  // Locally users are hard deleted. Once they are deleted they cannot be found via search as they are removed from the DB and only reseeded when the local API is restarted
+  // In deployed envs the user is soft deleted and can be found via search
+  if (process.env.CI) {
```
edavey, line 106: I agree that we don't need to fix this now -- but I wonder if this is dangerous pattern:

```
 if (process.env.CI) {
  some test which may well fail in CI and not be possible to reproduce locally
}
```

Would it be better to extract this to a separate test which we [skip](https://playwright.dev/docs/api/class-test#test-skip) for now? That way it would perhaps be more obvious that it's a piece of work which we need to do, rather than an exception to our rule that the tests should run locally?

.env.example
```diff
@@ -6,3 +6,4 @@ APPROVED_PREMISES_ROOT_PATH=approved-premises
 HMPPS_AUTH_USERNAME=
 HMPPS_AUTH_PASSWORD=
 E2E_USER_TO_ADD=BERNARD.BEAKS
+E2E_EMERGENCY_USER_TO_ALLOCATE_TO="JIM SNOW"
```
edavey: There's a bit of a gotcha here in the way we are specifying:

- Delius **username** for `E2E_USER_TO_ADD` and 
- Delius **name** for `E2E_EMERGENCY_USER_TO_ALLOCATE_TO`

Should we try to stress difference, e.g. with:

- `E2E_DELIUS_USERNAME_TO_ADD`
- `E2E_EMERGENCY_ASSESSOR_NAME_TO_ALLOCATE_TO`

e2e/playwright.config.ts
```diff
@@ -4,11 +4,11 @@ import { TestOptions } from '@approved-premises/e2e'
 
 export default defineConfig<TestOptions>({
   testDir: './',
-  fullyParallel: false,
+  fullyParallel: true,
   forbidOnly: !!process.env.CI,
   retries: process.env.CI ? 2 : 0,
   maxFailures: process.env.CI ? 3 : 1,
-  workers: 1,
+  workers: 2,
```
edavey, line 11: nice, thks for the speedup

edavey (approved):
👍 I've made a few suggestions where perhaps we could be clearer with the names of things?

e2e/tests/admin.spec.ts
richpjames: Yeah that makes sense to me.  I agree this is a dangerous pattern.
When originally adding this test I wanted to highlight this behaviour as it seems a bit strange but I'm not sure it's desirable to change the behaviour of the deployed code (as users may want to find previously deleted users) OR the code running locally (I think it may be difficult to recreate the live-like soft-delete).

.env.example
richpjames: yep, good spot

e2e/test.ts
```diff
@@ -18,6 +18,7 @@ export const test = base.extend<TestOptions>({
     },
     { option: true },
   ],
+  userToAdd: [{ name: process.env.E2E_USER_TO_ADD as string }, { option: true }],
```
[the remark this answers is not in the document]
richpjames: I agree this name could be improved, `e2eUserToDeleteAndAdd` makes sense to me. 
But just to be clear this is *not* the user that is driving the test. It is a seperate account that the test uses to add to the service and then delete again.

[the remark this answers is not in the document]
edavey: ah OK, thank for clarifying!

PR #1916, opened 2024-06-14
APS-916: Only users with the role 'Admin' should be able to access 'M…

vubalasu:
…anage user roles'

# Context

# Changes in this PR
Only users with the role 'Admin' should be able to access 'Manage user roles'/'Users'
## Screenshots of UI changes
No UI changes


patrickjfl (changes requested):
Thanks Bala. I think there are a few issues still to be addressed with this work though:

1. The first commit only contains work that you then remove in your second commit. Also the second commit references a completely different ticket (905). Could you clean up the git history? 
2. The E2E tests are failing, I assume they will need updating as part of this work.
3. As well as removing the tile for users without the admin role, can you also restrict the user management route, so that it would be impossible for someone without the role to access the page manually?

Let me know if any of this is unclear or you want to discuss it.

Rosie-Brigham (approved):

PR #1920, opened 2024-06-17
APS-878 - Add IP Whitelist for Palo Alto Egress

davidatkinsuk:
This IP range has been added to the whitelist to support project evolve, migrating/upgrading users to MOJO devices and network. This specific palo-alto-prisma-access-egress range doesn’t exist in the common whitelisting.

I've taken the same approach as CAS3 for this (see https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/983/files)

### DIFF
--- a/helm_deploy/hmpps-approved-premises-ui/values.yaml
+++ b/helm_deploy/hmpps-approved-premises-ui/values.yaml
@@ -67,6 +67,7 @@ generic-service:
     groups:
       - internal
       - probation
+    palo-alto-prisma-access-egress: '128.77.75.64/26'
 
 generic-prometheus-alerts:
   targetApplication: hmpps-approved-premises-ui


helm_deploy/hmpps-approved-premises-ui/values.yaml
```diff
@@ -67,6 +67,7 @@ generic-service:
     groups:
       - internal
       - probation
+    palo-alto-prisma-access-egress: "128.77.75.64/26"
```
richpjames, line 70: Sorry to be a pedant but please could we use single quotes here as that's what's used throughout the rest of the file

davidatkinsuk: i'm all for pedantry :D

richpjames (approved):

PR #1930, opened 2024-06-18
APS-855 - Add new raw app and request for placement reports

davidatkinsuk:
These new reports are currently controlled by a feature flag as we don’t have sufficient data backfilled in production for them to be used.

# Context

# Changes in this PR

### Before

Screenshot 2024-06-18 at 14 27 33

### After

Screenshot 2024-06-18 at 14 29 03


vubalasu:
@davidatkinsuk looks fine but we need to write unit test when config.flags.providePerformanceHubReports is true and update the tests accordingly

vubalasu (approved):

PR #1934, opened 2024-06-20
APS 911 - allow users to access OOS beds list from premises page

patrickjfl:
# Context

Jira ticket: https://dsdmoj.atlassian.net/browse/APS-911

# Changes in this PR

- adds new v2PremisesController to to render the new premises page
- updates actions menu on premises page to include 'Manage out of service bed records'
- adds AP area name above premises name

### Before

Screenshot 2024-06-20 at 14 36 58

### After

Screenshot 2024-06-20 at 14 36 34

### DIFF
--- a/integration_tests/mockApis/premises.ts
+++ b/integration_tests/mockApis/premises.ts
@@ -44,9 +44,25 @@ const stubPremisesSummary = (premises: ExtendedPremisesSummary) =>
     },
   })
 
+const stubSinglePremises = (premises: Premises) =>
+  stubFor({
+    request: {
+      method: 'GET',
+      urlPattern: `/premises/${premises.id}`,
+    },
+    response: {
+      status: 200,
+      headers: {
+        'Content-Type': 'application/json;charset=UTF-8',
+      },
+      jsonBody: premises,
+    },
+  })
+
 export default {
   stubAllPremises,
   stubPremisesSummary,
+  stubSinglePremises,
   stubPremisesWithBookings: (args: { premises: Premises; bookings: Array<Booking> }): Promise<[Response, Response]> =>
     Promise.all([
       stubPremisesSummary(args.premises),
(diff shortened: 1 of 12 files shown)


server/utils/premises/premisesActions.test.ts
```diff
@@ -119,4 +119,24 @@ describe('premisesActions', () => {
       })
     })
   })
+
+  describe('for users with no role', () => {
+    const user = userDetails.build({ roles: [] })
+    const premises = premisesFactory.build()
+
+    it('includes only the manage beds and manage out of service bed actions', () => {
```
edavey, line 127: I think the problem with asserting the equality of the whole list of links is that when something changes we don't get a useful error. That's why we introduced the `toContainManageAction()` and `toContainMenuItem()` matchers.

I think what we actually want to assert here is that `premisesActions()` includes the new link to "Manage out of service beds"?

It's great to describe the fact that we're also showing the "Manage beds" to users with roles but I think if we describe this in 2 scenarios it will be clearer and much easier to maintain e.g.

```ts
describe('for users with no role', () => {
   it('includes the "manage beds" action', () => {})
   it('includes the "out of service beds" action', () => {})
 })
```

integration_tests/mockApis/premises.ts
```diff
@@ -44,9 +44,25 @@ const stubPremisesSummary = (premises: ExtendedPremisesSummary) =>
     },
   })
 
+const stubSinglePremises = (premises: Premises) =>
```
edavey, line 47: is it worth including a mention in this commit message as to why we need to fetch the premises information?

edavey (approved):
Nice work, v clear, thank you! 

I've suggested breaking down the unit testing of `premisesActions()` further...

server/utils/premises/premisesActions.test.ts
patrickjfl: right, yes! I see now, thanks. have implemented the change 👍

integration_tests/mockApis/premises.ts
patrickjfl: have done

PR #1943, opened 2024-06-25
APS-597:  New user sees guidance if Delius acct missing staff details

vubalasu:
# Context

- https://dsdmoj.atlassian.net/browse/APS-597 

# Changes in this PR
New user sees guidance if Delius acct missing staff details
### After
Screenshot 2024-06-25 at 11 29 14


Boris-hmpps (approved):
Looks good to me.

PR #1949, opened 2024-06-27
APS-853 Enable Updated App and Placement Reports in Prod

davidatkinsuk:
These reports were disabled in production via commit d2bcf00b89890fcefa49c892ccf5b85463195446 because the required data had not been backfilled.

This data has now been backfilled, so these reports can be enabled


RobBoothMOJ (approved):
Approved

PR #1950, opened 2024-06-27
APS-946: Prevent users from changing the start and end date of a book…

vubalasu:
# Context

https://dsdmoj.atlassian.net/browse/APS-946 

# Changes in this PR
Prevent users from changing the arrival and departure date of a booking, when both dates remain the same as the current dates
### Before
The form will be submitted even though the arrival and departure dates are selected for change but didn't changed of a booking
### After
Screenshot 2024-06-27 at 14 07 25


Boris-hmpps (approved):
Looks good to me.

PR #1961, opened 2024-07-03
Chore/update v2 manage integration test pages

patrickjfl:
# Changes in this PR

To tidy up the v2 code, I have removed the v2 integration test page files that don't sit within the v2 directory, and updated those that do.

This PR is currently blocked until [this API PR ](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/2021)is merged and the associated types have been updated in the UI.


edavey (approved):
thanks for tidying up

server/utils/outOfServiceBedUtils.test.ts
```diff
@@ -85,7 +85,7 @@ describe('outOfServiceBedUtils', () => {
           text: 'Room',
         },
         {
-          text: 'Out of service from',
+          text: 'Start date',
```
richpjames, line 88: Do we want to change the user facing description? 
To me 'Out of service from' is clearer than 'Start date' but I'm not sure what discussions have happened around this whilst I've been away

richpjames (approved):

server/utils/outOfServiceBedUtils.test.ts
patrickjfl: as discussed on slack, we will leave as is pending feedback

PR #1964, opened 2024-07-03
APS-884: Future Manager creates out-of-service bed record from "v2 manage"

edavey:
This PR implements the "Future manager creates an out-of-service bed record" and incorporates:

- [APS-953](https://dsdmoj.atlassian.net/browse/APS-953): validate presence of OOSB "notes"
- [APS-956](https://dsdmoj.atlassian.net/browse/APS-956): obtain "reasons" from new CAS1 reference data endpoint
- [APS-910](https://dsdmoj.atlassian.net/browse/APS-910): redirect on success to v2 bed page with improved flash notification

[APS-953]: https://dsdmoj.atlassian.net/browse/APS-953?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[APS-956]: https://dsdmoj.atlassian.net/browse/APS-956?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[APS-910]: https://dsdmoj.atlassian.net/browse/APS-910?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/i18n/en/errors.json
```diff
@@ -93,23 +93,13 @@
...
     "empty": "You must enter a end date",
     "invalid": "The end date is an invalid date",
     "beforeStartDate": "The end date must be before the start date",
-    "conflict": "This bedspace is not available for these dates"
-  },
-  "outOfServiceTo": {
-    "empty": "You must enter an end date",
-    "invalid": "The end date is an invalid date",
-    "conflict": "This bedspace is not available for these dates"
+    "conflict": "This booking is not available for these dates"
```
richpjames, line 105: I made these changes on a seperate branch and ran into the same realisation as detailed in your commit message. 
The simplest suggestion I can think of is namespacing the error messages so rather than  
``` { ... 
 "endDate": {
   "conflict": "... error message..."
}
```   
we have: 
``` { ... 
"outOfServiceBeds": {
 "endDate": {
   "conflict": "... error message..."
  }
"bookings":{
 "endDate": {
   "conflict": "... error message..."
  } 
 }
```
But I imagine this will require quite a lot of work in the API to implement.  
Specifically for the `conflict` error we could lean on [this component](https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/780/commits/718e5bd292598c6dae76f4da9ba5a3668324b860) but the approach doesn't scale.

server/utils/bedUtils.test.ts
```diff
@@ -65,12 +67,32 @@ describe('bedUtils', () => {
...
       it('returns a "V1" link to manage the room', () => {
         expect(actionCell(bed, premisesId, user)).toEqual({
-          html: bedLink(bed, premisesId),
+          html: v1BedLink(bed, premisesId),
+        })
+      })
+    })
+
+    describe('when no user is given', () => {
+      const nullUser: UserDetails = null
```
richpjames, line 92: If `actionCell` is only called with 2 arguments then the `user` param will be `undefined` (not `null`). There are some subtle differences in the behaviour between  the`null` and `undefined` primitives so I think it would be better to use `undefined` here just to be safe.

e2e/pages/manage/v2MarkBedAsOutOfServicePage.ts
```diff
@@ -0,0 +1,67 @@
...
+      .getByLabel('Day')
+      .fill(getDate(endDate).toString())
+    await this.page
+      .getByRole('group', { name: 'Out of service to' })
+      .getByLabel('Month')
+      .fill(getMonth(endDate).toString())
+    await this.page
+      .getByRole('group', { name: 'Out of service to' })
+      .getByLabel('Year')
+      .fill(getYear(endDate).toString())
```
richpjames, line 53: Not for this PR but I think it would be useful if we have a method on the `BasePage` class that fills a Date field input group given its name

richpjames (approved):

richpjames (approved):
LGTM once the build is passing

server/utils/bedUtils.test.ts
edavey: thanks Rich, I've updated this to describe the behaviour with `undefined` rather than `null`

e2e/pages/manage/v2MarkBedAsOutOfServicePage.ts
edavey: yes that will useful

server/i18n/en/errors.json
edavey: yes I was thinking that the API could use the model name when generating this error messages. But I'm not sure how we'd introduce that change smoothly across all 3 CAS services...

I'll bring this up in our next x-CAS API meeting...

PR #1969, opened 2024-07-05
APS 888 - update view OOS bed page

patrickjfl:
# Context

Jira ticket: https://dsdmoj.atlassian.net/browse/APS-888

# Changes in this PR

Updates v2 show OOS bed page to show details of the 'latest revision', as well as the bed characteristics. Also removes the now redundant form from that page.

- [] I have run the E2E tests locally and they passed

### Before

OutOfServiceBeds -- should show a out of service bed

### After

OutOfServiceBeds -- should show an out of service bed

### DIFF
--- a/server/views/v2Manage/outOfServiceBeds/show.njk
+++ b/server/views/v2Manage/outOfServiceBeds/show.njk
@@ -1,18 +1,20 @@
 {% from "govuk/components/summary-list/macro.njk" import govukSummaryList %}
 {% from "govuk/components/back-link/macro.njk" import govukBackLink %}
-{% from "govuk/components/button/macro.njk" import govukButton %}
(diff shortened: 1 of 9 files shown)


richpjames (approved):

server/views/v2Manage/outOfServiceBeds/show.njk
```diff
@@ -6,6 +6,8 @@
 {% from "govuk/components/radios/macro.njk" import govukRadios %}
 {% from "govuk/components/back-link/macro.njk" import govukBackLink %}
```
richpjames, line 7: I think a lot of these imports will be redundant now, could we remove the unused ones?

patrickjfl: good point!

PR #1980, opened 2024-07-10
APS 889 - Add OOS bed update page

richpjames:
# Context

Users need to be able to update an Out of service bed record with revisions. 

[Jira](https://dsdmoj.atlassian.net/jira/software/c/projects/APS/boards/1328?search=889&selectedIssue=APS-889&sprints=6188) 

# Changes in this PR
- We add the form to enable revisions 
- Allow the revision to be posted to the API
- We add error handling to ensure the users see errors the API returns

### Mock
AP Design Reviews ✏️ (7)

### Development
after


server/paths/manage.ts
```diff
@@ -132,8 +132,8 @@ const v2Manage = {
     create: outOfServiceBedsPath,
     premisesIndex: v2SinglePremisesPath.path('out-of-service-beds').path(':temporality'),
     index: outOfServiceBedsIndexPath.path(':temporality'),
-    update: singlePremisesPath.path('out-of-service-beds').path(':id'),
     show: outOfServiceBedPath.path(':tab'),
+    update: outOfServiceBedPath.path('/update'),
     cancel: singlePremisesPath.path('out-of-service-beds').path(':id').path('cancellations'),
   },
```
edavey: Just wondering whether more smaller things might be easier to manage? See my later question about whether an entirely new "resource" might be clearer? e.g.  something like:

```ts
const outOfServiceBedsUpdatesPath = outOfServiceBedsIndexPath.path(':outOfServiceBedId')
// Path<"/manage/out-of-service-beds/:outOfServiceBedId">

outOfServiceBedUpdates: {
    new: outOfServiceBedsUpdatesPath.path('new'),
    create: outOfServiceBedsUpdatesPath,
  },
```

server/views/v2Manage/outOfServiceBeds/show.njk
```diff
@@ -30,7 +30,7 @@
               items: [
                 {
                   text: "Update record",
-                  href: paths.v2Manage.outOfServiceBeds.update({premisesId: premisesId, id: id})
+                  href: paths.v2Manage.outOfServiceBeds.update({premisesId: premisesId, id: id, bedId: bedId})
```
edavey, line 34: Probably I'm missing something but I think we want to be using the OutOfServiceBed ID here rather than the Bed ID? When we're making a new OOSB then clearly the Bed is a key component, but when we are creating an OOSBUpdate, we are concerned with the OOSB ID and the bed id is no longer important?

integration_tests/pages/page.ts
```diff
@@ -511,4 +513,37 @@ export default abstract class Page {
       this.assertDefinition('Prison', person.prisonName)
     })
   }
+
+  shouldShowOutOfServiceBedDetails(bedDetails: OutOfServiceBed | OutOfServiceBedRevision) {
+    if (bedDetails.startDate) {
```
edavey, line 526: Again I wonder if we should be a bit stricter with our terminology to avoid future confusion when reading this: i.e. this variable is more correctly an `outOfServiceBedDetails` isn't it?

integration_tests/tests/v2Manage/future_manager/updatesOutOfServiceBed.cy.ts
```diff
@@ -34,5 +34,9 @@ describe('Updating an out of service bed', () => {
 
     // Then I should be taken to the OoS bed update page
     const updatePage = Page.verifyOnPage(OutOfServiceBedUpdatePage, outOfServiceBed)
+
+    // And it should show the details of the bed from the OoS bed record
+    updatePage.shouldShowOutOfServiceBedDetails(outOfServiceBed)
+    updatePage.formShouldBePrepopulated()
```
edavey, line 40: very nice!

integration_tests/pages/v2Manage/outOfServiceBeds/outOfServiceBedShow.ts
```diff
@@ -71,6 +71,10 @@ export class OutOfServiceBedShowPage extends Page {
     })
   }
 
+  shouldShowUpdateConfirmationMessage() {
+    this.shouldShowBanner('The out of service bed has been updated')
```
edavey, line 75: 👍 I see this matches what we have in the 'create' action: https://github.com/ministryofjustice/hmpps-approved-premises-ui/blob/main/server/controllers/v2Manage/outOfServiceBedsController.test.ts#L95-L125

To my mind this does read better than "The out of service bed _record_ has been updated" but we can see what others think...

server/controllers/v2Manage/updateOutOfServiceBedsController.test.ts
```diff
@@ -90,4 +96,123 @@ describe('updateOutOfServiceBedController', () => {
...
+      const requestHandler = updateOutOfServiceBedController.create()
+
+      await requestHandler(request, response, next)
+
+      expect(response.redirect).toHaveBeenCalledWith(
+        paths.v2Manage.outOfServiceBeds.show({
+          premisesId,
+          bedId: outOfServiceBed.bed.id,
+          id: outOfServiceBed.id,
+          tab: 'timeline',
```
edavey, line 230: 👍

edavey (approved):
Really clear, thks! 

A couple of suggestions:

1. I think we ought to be explicit (even though verbose) when referring to an `OutOfServiceBed` as distinct from a `Bed`
2. it's worth considering modelling these updates as separate REST resources in your URL scheme. You've already recognised their difference with a separate controller, but with a separate url scheme we could drop the `bedId` (and even the `premisesId`) from the URL scheme. Not sure if that would be a "now" refactoring or not though!

PR #1983, opened 2024-07-11
Use OoS bed entity for bed details, not the revision

richpjames:
Previously we were using the revision which only contains the updated details not the full details. The top level entity is automatically updated when revisions are made so it is safe to use here


edavey (approved):
Apologies for not properly communicating earlier that the OOSBEntity gives us the values of the latest revision. And so we need only handle revisions directly when we produce the timeline

PR #1984, opened 2024-07-11
Translate characteristic.propertyName to a positive statement

edavey:
[Jira APS-1059](https://dsdmoj.atlassian.net/browse/APS-1059)

### Before

before_question

### After

after_statement

Previously we were using `Characteristic.name` which is a string posing a question such as:

> Is this a single room?

or

> Is there a smoke/heat detector in the room?

With this commit we introduce a

```
characteristicUtils.translateCharacteristic(CharacteristicPair)
```

function which attempts to lookup `Characteristic.propertyName` in a JSON translation file (`i18n/en/characteristics.json`) to find a label for the characteristic which is cast in the form of a positive statement form, e.g.

> Single room

or

> Smoke/heat detector

If a translation for the given `propertyName` isn't found we return the `propertyName`. These are (just about) human readable e.g.

- `acceptsSexOffenders`
- `acceptsChildSexOffenders`

### DIFF
--- a/server/i18n/en/characteristics.json
+++ b/server/i18n/en/characteristics.json
@@ -1,0 +1,23 @@
+{
+  "isSingle": "Single room",
+  "isFullyFm": "All furnishings and bedding supplied by FM",
+  "hasCrib7Bedding": "Crib 7 rated bedding",
+  "hasSmokeDetector": "Smoke/heat detector",
+  "isTopFloorVulnerable": "On top floor and vulnerable",
+  "isGroundFloorNrOffice": "On ground floor close to office",
+  "hasNearbySprinkler": "Near to sprinkler",
+  "isArsonSuitable": "Suitable for arson risks",
+  "isArsonDesignated": "Designated arson room",
+  "hasArsonInsuranceConditions": "Not insured for arson risk",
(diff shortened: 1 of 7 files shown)


server/i18n/en/characteristics.json
```diff
@@ -0,0 +1,23 @@
...
+  "isTopFloorVulnerable": "On top floor and vulnerable",
+  "isGroundFloorNrOffice": "On ground floor close to office",
+  "hasNearbySprinkler": "Near to sprinkler",
+  "isArsonSuitable": "Suitable for arson risks",
+  "isArsonDesignated": "Designated arson room",
+  "hasArsonInsuranceConditions": "Not insured for arson risk",
+  "isSuitedForSexOffenders": "Suitable for sex offenders",
+  "hasEnSuite": "En suite",
+  "isWheelchairAccessible": "Can be accessed in wheelchair",
+  "hasWideDoor": "Door >= 90cm wide",
```
patrickjfl, line 15: Are we already using this translation? It strikes me as a little technical (compared to e.g. "Door 90cm or wider")

patrickjfl (approved):
Looks good to me, nice clear improvement 👍

PR #1985, opened 2024-07-11
Allow assessors to assess request for placements

davidatkinsuk:
Previously this tab was only available to workflow_manager and matcher. This should also be available to assessors as the API will start to allow placement requests to be allocated to users with this role

# Context

Relates to JIRA ticket https://dsdmoj.atlassian.net/browse/APS-969

# Changes in this PR

## Screenshots of UI changes

Screenshots are when logged in as user with 'assessor' role

### Before

Screenshot 2024-07-11 at 15 05 03

### After

Screenshot 2024-07-11 at 15 06 50

### DIFF
--- a/server/utils/assessments/utils.test.ts
+++ b/server/utils/assessments/utils.test.ts
@@ -413,7 +413,7 @@ describe('utils', () => {
   })
 
   describe('assessmentsTabItems', () => {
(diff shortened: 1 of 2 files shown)


edavey (approved):
👍

PR #1986, opened 2024-07-11
APS 1049 - link to premises and bed from OOS bed page

patrickjfl:
# Context

Jira ticket: https://dsdmoj.atlassian.net/browse/APS-1049

# Changes in this PR

Adds links to the out of service bed page heading, so that the user can navigate to the premises and bed pages.


edavey (commented):
Can we include a unit test to describe this and prevent regression?
Or describe the navigation from OOSN to i) premises and ii) bed in an integration test?

patrickjfl:
> Can we include a unit test to describe this and prevent regression? Or describe the navigation from OOSN to i) premises and ii) bed in an integration test?

It's not unit testable as is, but we could include an integration test. I wasn't sure how much value it delivered but preventing regression is a good point.

richpjames (approved):

PR #1988, opened 2024-07-11
APS 889 - Improve OoS bed update date validation

richpjames:
# Context

[JIRA](https://dsdmoj.atlassian.net/jira/software/c/projects/APS/boards/1328?selectedIssue=APS-889&sprints=6188)

# Changes in this PR

This improves the date validation on the OoS bed inputs and corrects an error message


edavey (approved):
👍 I agree that we ought to introduce tests ('m not sure at which level?) to describe how we handle the submission of invalid dates (gracefully)

PR #1989, opened 2024-07-12
APS 1058 - enable OOS bed functionality for cru member

patrickjfl:
# Context

Jira ticket: https://dsdmoj.atlassian.net/browse/APS-1058

# Changes in this PR

Allows users with role cru_member to view, create and update out of service beds.

### DIFF
--- a/server/routes/v2Manage.test.ts
+++ b/server/routes/v2Manage.test.ts
@@ -1,0 +1,164 @@
+import { Router } from 'express'
+import { DeepMocked, createMock } from '@golevelup/ts-jest'
+import { BookingExtensionsController, BookingsController, PremisesController } from '../controllers/manage'
+import v2ManageRoutes from './v2Manage'
+import { type Controllers } from '../controllers'
+import { type Services } from '../services'
+import {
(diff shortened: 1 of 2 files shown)


edavey (approved):
thanks for restructuring this.

I think it's helpful, in principle, to exercise this journey for both the Future Manager and the CRN Member but this pattern of testing could cause the time taken to run our test suite to balloon, so in general we might what to focus more on unit testing of the `routes/*` file to describe the detailed authorisation rules around a particular journey. What do you think?

patrickjfl:
> thanks for restructuring this.
> 
> I think it's helpful, in principle, to exercise this journey for both the Future Manager and the CRN Member but this pattern of testing could cause the time taken to run our test suite to balloon, so in general we might what to focus more on unit testing of the `routes/*` file to describe the detailed authorisation rules around a particular journey. What do you think?

Yeah, good point. I agree with the approach you suggest. Do you think it's worth changing this PR to reflect that, or agreeing it as an approach in future?

richpjames (approved):

PR #1991, opened 2024-07-15
APS-1066 - add e2e tests for oos beds

patrickjfl:
# Context

Jira ticket: https://dsdmoj.atlassian.net/browse/APS-1066

# Changes in this PR

Adds an E2E test for updating an out of service bed. Also fixes a broke v1 manage test.

### DIFF
--- a/e2e/tests/manage.spec.ts
+++ b/e2e/tests/manage.spec.ts
@@ -24,6 +24,7 @@ test.describe.configure({ mode: 'parallel' })
 
 const premisesName = 'Test AP 10'
 const apArea = 'South West & South Central'
+const roomName = 'Bed 101 - 1'
 
 const navigateToPremisesPage = async (page: Page, { filterPremisesPage } = { filterPremisesPage: false }) => {
   // Given I visit the dashboard
@@ -187,7 +188,7 @@ test('Mark a bed as out of service', async ({ page, legacyManager }) => {
   await bedsPage.viewAvailableBed()
 
   // Then I should be able to mark a bed as out of service
-  const bedPage = await BedPage.initialize(page, 'Manage beds')
+  const bedPage = await BedPage.initialize(page, roomName)
   await bedPage.clickMarkBedAsOutOfService()
 
   // When I fill in and submit the form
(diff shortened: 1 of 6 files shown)


richpjames (approved):

e2e/tests/v2Manage.spec.ts
```diff
@@ -50,3 +54,56 @@ test('Future manager marks a bed as out of service in the V2 Manage area', async
...
+  // When I select 'Update record'
+  await outOfServiceBedPage.selectUpdateRecord()
+
+  // Then I should see the update out of service bed record page
+  const updateOOSBedPage = await UpdateOutOfServiceBedPage.initialize(page)
+
+  // When I update the out of service bed record
+  const update = {
+    referenceNumber: faker.string.uuid(),
+    additionalInformation: faker.lorem.sentence(),
```
edavey, line 95: would you agree that this test will be easier to debug if we use realistic content for the update?

e.g.

```ts
const update = {
  referenceNumber: 'REF-ABC123',
  additionalInformation: 'The supplier informed us of a delay in starting work'
}
```

In that way an error message might say 

> expected to find 'The supplier informed us of a delay in starting work' on the page

rather than

> expect to find 'Lorem ipsum nonsense gobbledeegook' on the page

e2e/tests/manage.spec.ts
```diff
@@ -24,6 +24,7 @@ test.describe.configure({ mode: 'parallel' })
 
 const premisesName = 'Test AP 10'
 const apArea = 'South West & South Central'
+const roomName = 'Bed 101 - 1'
```
edavey, line 27: thanks for fixing this up. I think this was my error it not updating!

edavey (approved):

e2e/tests/v2Manage.spec.ts
patrickjfl: I do agree. But having a unique string allows us to cover the case where the same update has been made in E2E tests multiple times. Then we can assert that each unique string is present on the timeline page, without having to write a more complex function to target the specific element we're looking for. To me, I thought it was worth trading off the time it would take to do that against a slightly less intuitive debugging experience. But maybe not 🤷

edavey: > I do agree. But having a unique string allows us to cover the case where the same update has been made in E2E tests multiple times. Then we can assert that each unique string is present on the timeline page, without having to write a more complex function to target the specific element we're looking for.

ah interesting, I hadn't understood that. How about combining the approaches?

e.g.

```ts
const uniqueUpdateId = faker.string.uuid()
const update = {
  referenceNumber: `UPDATED-REF [${uniqueUpdateId}]`,
  additionalInformation: `Additional info on update [${uniqueUpdateId}]`
}
```

patrickjfl: ok, good idea!

PR #1994, opened 2024-07-15
Remove unused match code

richpjames:
This PR removes some code that wasn't being used/called. It does not deprecate any existing functionality.


edavey (approved):
Thanks for clearing the ground.

PR #1998, opened 2024-07-16
Reintroduce Match transaction

richpjames:
This introduces the Match transaction in order to make the change from matching to a bed to matching to a 'space' easier. It essentially reintroduces what previously existed with very little changed.  
The journey will be accessed from the Placement Request page (placement requests are accessed via the CRU dashboard) via a link in the menu. Currently the link says 'Create placement' but if the ENABLE_V2_MATCH env var (introduced in this PR) is 'true' then it will be a replaced by a link to this new journey under the name 'Search for a space'. As can be seen in the screenshot below. 
image
[Jira](https://dsdmoj.atlassian.net/jira/software/c/projects/APS/boards/1328?assignee=62a050ba9f5d480069c97f49&selectedIssue=APS-1080)


server/utils/placementRequests/utils.test.ts
```diff
@@ -28,7 +28,7 @@ describe('utils', () => {
       searchButton(placementRequest)
 
       expect(utils.linkTo).toHaveBeenCalledWith(
-        paths.placementRequests.beds.search,
+        paths.v2Match.placementRequests.spaces.search,
```
edavey, line 31: I think we could make a good case for saying that the resource we're creating here is a "space search", or "space-search". i.e. search is the noun, rather than the verb. We're making a search and the specific type of search is a space-search...

server/views/match/placementRequests/bookings/success.njk
```diff
@@ -18,7 +18,7 @@
 
       {{ govukButton({
         text: "Return to dashboard",
-        href: paths.assessments.index({})} + "?activeTab=requests_for_placement",
+        href: paths.assessments.index({}) + "?activeTab=requests_for_placement",
```
edavey, line 21: well tracked down! 🕵️

server/utils/placementRequests/adminIdentityBar.test.ts
```diff
@@ -27,23 +40,54 @@ describe('adminIdentityBar', () => {
...
+        config.flags.v2MatchEnabled = 'true'
+      })
+      afterEach(() => {
+        config.flags.v2MatchEnabled = originalFlagValue
+      })
+
+      it('should return the v2 path for "Create placement"', () => {
+        const placementRequestDetail = placementRequestDetailFactory.build({ status: 'notMatched' })
+
+        expect(adminActions(placementRequestDetail)).toEqual([
```
edavey, line 76: could we use an "inclusion assertion" here (like https://github.com/ministryofjustice/approved-premises-ui/blob/98e2b9769263bb0896d0e94cc2e9f51861a606b5/server/testutils/jest.setup.ts#L73) to be more granular in describing which actions are included in the list of links? The problem with asserting equality on the whole list is that we don't get a helpful error if the entire set of links isn't as expected...

integration_tests/tests/match/match.cy.ts
```diff
@@ -11,75 +9,48 @@ import {
...
-
-    // Given I am logged in
-    cy.signIn()
+    cy.task('stubApAreaReferenceData')
   })
 
-  it('allows me to search for available rooms', () => {
-    // Given there is a placement request waiting for me to match
-    const person = personFactory.build()
+  it('allows me to search for an avaialble space', () => {
```
edavey, line 30: tiny typo: "available"

server/config.ts
```diff
@@ -48,6 +48,7 @@ export default {
   fliptEnabled,
   flags: {
     oasysDisabled: process.env.OASYS_DISABLED || false,
+    v2MatchEnabled: process.env.ENABLE_V2_MATCH || false,
```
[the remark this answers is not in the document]
richpjames: good spot

server/paths/match.ts
```diff
@@ -1,10 +1,24 @@
 import { path } from 'static-path'
 
+const v2MatchPath = path('/match')
+const v2PlacementRequestsPath = v2MatchPath.path('placement-requests')
+const v2PlacementRequestPath = v2PlacementRequestsPath.path(':id')
+const v2PlacementRequestSearchPath = v2PlacementRequestPath.path('search')
```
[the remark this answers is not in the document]
richpjames: yeah I think so

server/utils/placementRequests/adminIdentityBar.test.ts
richpjames: yeah absolutely - this slipped my mind

PR #2001, opened 2024-07-18
Allow OoS bed tests to be run independently

richpjames:
Previously the second test (that updates an OoS bed) depended on the first test creating an OoS bed to use. Now the second test makes its own OoS bed to use, meaning it can be run by itself. This will increase the time it takes to run the test but it should mean the suite is more durable moving forward. It also makes the first test a little harder to read as you have to instead read the function it calls rather than the test body, but I think it's a price worth paying.


patrickjfl (approved):
nice 🙌

PR #2002, opened 2024-07-18
APS-979: Change all occurances of LAO:Not found to Not Found with crn

vubalasu:
# Context

https://dsdmoj.atlassian.net/browse/APS-979

# Changes in this PR
 remove the ‘LAO: ' prefix if an applicant is not found and replace with Not Found CRN: crn

### Before
image

### After
Screenshot 2024-07-18 at 13 52 04


Boris-hmpps (approved):
Looks good to me

PR #2006, opened 2024-07-19
Chore/add new find and book factories

patrickjfl:
# Changes in this PR

To facilitate parallel working on the find and book journey, we are adding all these factories in their own PR.

### DIFF
--- a/server/testutils/factories/premisesSearchResultSummary.ts
+++ b/server/testutils/factories/premisesSearchResultSummary.ts
@@ -1,0 +1,23 @@
+import { Factory } from 'fishery'
+import { faker } from '@faker-js/faker/locale/en_GB'
+
+import type { Cas1PremisesSearchResultSummary } from '@approved-premises/api'
+import namedIdFactory from './namedId'
+import { apCharacteristicPairFactory } from './bedSearchResult'
+
+export default Factory.define<Cas1PremisesSearchResultSummary>(() => {
+  return {
+    id: faker.string.uuid(),
+    apCode: faker.string.alphanumeric(5),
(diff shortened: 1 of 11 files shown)


server/testutils/factories/premisesSearchResultSummary.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export default Factory.define<Cas1PremisesSearchResultSummary>(() => {
+  return {
+    id: faker.string.uuid(),
+    apCode: faker.string.alphanumeric(5),
+    deliusQCode: faker.string.alphanumeric(5),
+    apType: faker.helpers.arrayElement(['normal', 'pipe', 'esap', 'rfap', 'mhapStJosephs', 'mhapElliottHouse']),
+    name: faker.lorem.words(3),
+    addressLine1: faker.location.streetAddress(),
+    addressLine2: faker.location.secondaryAddress(),
```
richpjames, line 17: `secondaryAddress` 'Generates a random localized secondary address. This refers to a specific location at a given address such as an apartment or room number.' - I'm not sure this is close to what we want. Maybe `county` would do but it's not very close either.

richpjames (approved):

server/testutils/factories/premisesSearchResultSummary.ts
patrickjfl: hmmm. maybe we could hard code something that is more realistic- - any thoughts on a value? I'm assuming that there is also regularly no value for this.

richpjames: yeah I suspect it will often be undefined. Maybe `faker.location.county` would do?

PR #2010, opened 2024-07-22
[APS-1078] Fix historic out-of-service beds not displaying in dashboard

danlivings-dxw:
# Context

See [APS-1078 on the Approved Premises Jira board](https://dsdmoj.atlassian.net/browse/APS-1078).

# Changes in this PR

When using the `/manage/out-of-service-beds` dashboard, historic out-of-service beds are not displayed when the 'Historic' tab is selected.

This is because the UI is using the value `'historic'`, which is not a valid value for the `Temporality` type. The correct value is `'past'`, and using this value resolves the issue.

No user-facing changes have been made, so the tab still uses the 'Historic' label.

## Screenshots of UI changes
*No visual change.*


patrickjfl (approved):
LGTM, I assume that was just a flaky integration test failure - they're passing for me locally.

PR #2015, opened 2024-07-22
APS-1095: Remove 'Arrivals/Departures' and 'Current residents' from v2 premises page

edavey:
[APS-1095: remove elements from premises dashboard](https://dsdmoj.atlassian.net/browse/APS-1095)

## Before
premises_page_before

## After

premises_pages_after

We also wire up the v2 premises list page to provide the "future manager" with links from the premises list page to the v2 premises pages.

v2_premises_list_page


server/utils/users/homePageDashboard.ts
```diff
@@ -98,10 +107,14 @@ export const sectionsForUser = (user: UserDetails): Array<ServiceSection> => {
     items.push(sections.assess)
   }
 
-  if (managerRoles.some(role => hasRole(user, role))) {
+  if (hasRole(user, 'manager')) {
```
richpjames, line 110: Given the tiles are identical I could see this being confusing for users who have both 'manager' and 'future_manager' roles (which shouldn't happen but will). I wonder if we could either have some differentiation in the copy and/or title *or* use an `if...else` to ensure they only get one 'manage' tile

richpjames (approved):

server/utils/users/homePageDashboard.ts
edavey: yes good point, I did think about this but wasn't sure which way to go...

On reflection I think that we should show the v2 tile if the user has the future manager role. I'll tweak this...

PR #2036, opened 2024-07-31
APS-1115: If Offender Not Found when viewing /admin/placement-request…

vubalasu:
…s/{id}, name appears as 'Limited Access Offender'

# Context

https://dsdmoj.atlassian.net/browse/APS-1115

# Changes in this PR
If Offender Not Found when viewing /admin/placement-requests/{id}, name appears as 'Limited Access Offender'

### Before
image

### After
image


Boris-hmpps (approved):
Looks good to me.

PR #2037, opened 2024-07-31
APS 1082 -  Matcher space search iterations

richpjames:
# Context

[JIRA](https://dsdmoj.atlassian.net/jira/software/c/projects/APS/boards/1328?assignee=62a050ba9f5d480069c97f49&selectedIssue=APS-1082)

# Changes in this PR

Some iterations to improve the look and feel of the Search page.
There are a lot of moving util files around which is messy and causes rebase headaches but hopefully worth it for the clarity it produces.

### Before
image

### After

after


server/utils/matchUtils.ts
```diff
@@ -363,3 +365,29 @@ export const placementRequestSummaryListForMatching = (placementRequest: Placeme
 
   return rows
 }
+
+export const keyDetails = (placementRequest: PlacementRequestDetail): KeyDetailsArgs => {
+  const { person } = placementRequest
+  if (!isFullPerson(person)) throw Error('Restricted person')
```
patrickjfl, line 356: If the person is restricted, what do we want to happen to this bar? Does the error bubble up so the whole page isn't rendered?

server/views/match/search.njk
```diff
@@ -7,6 +7,7 @@
 {% from "govuk/components/radios/macro.njk" import govukRadios %}
 {% from "govuk/components/summary-list/macro.njk" import govukSummaryList %}
 {% from "govuk/components/back-link/macro.njk" import govukBackLink %}
+{% from "govuk/components/details/macro.njk" import govukDetails %}
```
patrickjfl, line 10: just the import here, maybe you included calling the component in a different commit?

patrickjfl (approved):
LGTM! nice work 🙌

server/utils/matchUtils.ts
richpjames: It's a good point - I'm not really sure what should happen, it's one of those things that shouldn't happen (but will). In this case there will be a blocking error. 
If the CRN is restricted the user won't be able to click on it on the CRU Dashboard 
image
Which means they won't see the link which takes them to the search pages which means they shouldn't be on this page.  
In time we will probably want a more helpful error but I feel like that change is out of scope for now.

server/views/match/search.njk
richpjames: good spot

PR #2040, opened 2024-08-01
Increase date range of OoS bed booking

richpjames:
When the e2e tests run against dev we're seeing test failures as the OoS bed is attempting to overbook an existing booking. By increasing the date range I hope that there are less conflicts.

### DIFF
--- a/e2e/pages/manage/v2MarkBedAsOutOfServicePage.ts
+++ b/e2e/pages/manage/v2MarkBedAsOutOfServicePage.ts
@@ -1,5 +1,5 @@
 import { Page, expect } from '@playwright/test'
-import { addDays, getDate, getMonth, getYear } from 'date-fns'
+import { addDays, addMonths, getDate, getMonth, getYear } from 'date-fns'
 import { faker } from '@faker-js/faker/locale/en_GB'
 import { BasePage } from '../basePage'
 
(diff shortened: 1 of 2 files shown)


davidatkinsuk (approved):
LGTM 👍

froddd:
Do we know how long an existing booking can be? If so, it would probably be better to ensure the date is _no sooner_ than the end of this booking, as this change will only reduce the frequency of this issue occurring, rather than prevent it.

richpjames:
@froddd an existing booking could be any length of time. The tests book them for 3 days but when running against dev/local it could encounter a booking of any length.   
I think we could see if there is a conflict error message: 
- if there is then try another date in 3 days time and see if *that* receives an error message 
- if there isn't proceed 
Is that the sort of thing you had in mind?

richpjames:
@davidatkinsuk minor but I've just noticed as part of this that the error (supplied by the API) shown when there is a conflict mentions 'bedspace' which isn't a term used in AP 
image

froddd:
I think CAS3 explicitly creates bedspaces for E2E tests, to ensure the bookings created against that don't clash with any other existing bookings. Short of that, I'm not sure how you could guarantee there would be no clash. But the faked date could at the very least enforce a _minimum_ period after the 3 days? Something along the lines of:

```js
instance.startDate = faker.date.future({ years: 1, refDate: date5DaysFromNow })
```

richpjames:
@froddd yeah if we could create a bedspace in order to start from a fresh slate that would be ideal but it's not possible in AP

PR #2042, opened 2024-08-02
Add build step to setup script

richpjames:
We don't run the build step (which compiles the SASS and copies across the views) in the setup script. This step is neccessary to ["set up a project in an initial
state"](https://github.com/github/scripts-to-rule-them-all?tab=readme-ov-file#scriptsetup) so should be included here


davidatkinsuk (approved):
LGTM 👍

PR #2045, opened 2024-08-05
Fix/Move filter header into correct div

richpjames:
This change allows the header to have a grey background instead of sitting above it

### DIFF
--- a/integration_tests/pages/match/searchPage.ts
+++ b/integration_tests/pages/match/searchPage.ts
@@ -2,8 +2,10 @@ import { SpaceSearchParametersUi, TextItem } from '@approved-premises/ui'
 import { Cas1SpaceSearchResult, Cas1SpaceSearchResults, PlacementRequestDetail } from '@approved-premises/api'
 import Page from '../page'
 import { uiObjectValue } from '../../helpers'
-import { summaryCardRows } from '../../../server/utils/match'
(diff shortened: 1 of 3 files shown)


integration_tests/pages/match/searchPage.ts
```diff
@@ -15,6 +17,29 @@ export default class SearchPage extends Page {
...
+      cy.get('span').contains(`Tier: ${placementRequest?.risks?.tier?.value.level}`)
+      cy.get('span').contains(`Date of birth: ${DateFormats.isoDateToUIDate(person.dateOfBirth, { format: 'short' })}`)
+    })
+  }
+
+  shouldShowMatchingDetails(placementRequest: PlacementRequestDetail) {
+    cy.get('.govuk-details').within(() => {
+      this.shouldContainSummaryListItems(placementRequestSummaryListForMatching(placementRequest))
+    })
+  }
```
patrickjfl, line 41: are you calling these two anywhere?

patrickjfl (approved):

integration_tests/pages/match/searchPage.ts
richpjames: line 20

patrickjfl: aha! 🙂

PR #2051, opened 2024-08-07
APS 1146 - Extend E2Es to include Match

richpjames:
This reinstates and extends the Match transaction in the E2E tests. 
This transaction was previously disabled but now development work has started on it again. 
The search page filters don't currently function so these can't be tested.
What can be tested is the information shown when the matcher performs the search: 
image
This information is pulled from Apply and Assess and we want to sure that it is populate correctly by the UI. This PR ensures that the information that is entered earlier in the journey is shown correctly. To do this we need to return the information from the earlier transactions and pass it to the Match transactions. This introduces a visible dependency between the tests but this existed previously it is just being made visible now.

### DIFF
--- a/e2e/pages/assess/listPage.ts
+++ b/e2e/pages/assess/listPage.ts
@@ -15,7 +15,7 @@ export class ListPage extends BasePage {
     const nextLink = this.page.getByRole('link', { name: 'Next' })
 
     try {
-      await expect(assessmentRow).toBeVisible()
+      await expect(assessmentRow).toBeVisible({ timeout: 1000 })
       await assessmentRow.click()
     } catch (err) {
       try {
--- a/e2e/pages/basePage.ts
+++ b/e2e/pages/basePage.ts
@@ -19,6 +19,10 @@ export class BasePage {
     await this.page.getByRole('button', { name: 'Withdraw' }).click()
   }
 
+  async clickActions() {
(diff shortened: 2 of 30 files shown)


e2e/pages/assess/listPage.ts
```diff
@@ -15,7 +15,7 @@ export class ListPage extends BasePage {
     const nextLink = this.page.getByRole('link', { name: 'Next' })
 
     try {
-      await expect(assessmentRow).toBeVisible()
+      await expect(assessmentRow).toBeVisible({ timeout: 1000 })
```
patrickjfl, line 18: nice

e2e/pages/basePage.ts
```diff
@@ -19,6 +19,10 @@ export class BasePage {
     await this.page.getByRole('button', { name: 'Withdraw' }).click()
   }
 
+  async clickActions() {
```
patrickjfl, line 23: nice

e2e/pages/basePage.ts
```diff
@@ -28,7 +28,7 @@ export class BasePage {
   }
 
   async checkRadio(label: string) {
-    await this.page.getByLabel(label, { exact: true }).check()
+    await this.page.getByLabel(label, { exact: true }).click()
```
patrickjfl, line 32: nice

patrickjfl (approved):
LGTM, good work 🙌

PR #2056, opened 2024-08-09
APS-1117: UI Implement User Version

vubalasu:
This PR forces the currently logged in user details to be fetched from the API if a call to the API returns a changed version hash for this user. The version and user id are returned from the API for any call as the `x-cas-user-version` and `x-cas-user-version` headers.

Note that, due to the user authentication and session setup happening _before_ any call to the API, it is impossible to force a user refresh on the same request cycle as the API call that returns the new version. Instead, this cycle completes, then the user is updated on the next request.

# Context

https://dsdmoj.atlassian.net/browse/APS-1117

# Changes in this PR

## Screenshots of UI changes
No uI changes

### DIFF
--- a/integration_tests/tests/admin/userManagement.cy.ts
+++ b/integration_tests/tests/admin/userManagement.cy.ts
@@ -335,4 +335,27 @@ context('User management', () => {
     // Then the page should show the results for the second page
     page.shouldShowUsers(usersForResultsPage2)
   })
+
+  it('retrieve the updated user from api if version not matched to user in session and shows the new role added report viewer', () => {
+    // Given there are users in the DB
+    const users = userFactory.buildList(10, { roles: ['assessor'] })
+    const user = userFactory.build({ version: 345, roles: ['report_viewer'] })
+    const roles = ['manager', 'matcher', 'workflow_manager']
+    cy.task('stubFindUser', { user, id: users[0].id })
+    cy.task('stubUsers', { users })
+    cy.task('stubApAreaReferenceData')
+    cy.task('stubAuthUser', { roles })
+
+    // When I visit the list page
+    const listPage = ListPage.visit()
+
+    // Then I should see the users and their details
+    listPage.shouldShowUsers(users)
+
+    // And I click on user
+    listPage.clickUser(users[0].name)
+
+    // Then I should see the user with new report role
+    listPage.shouldShowReportsMenu()
+  })
 })
--- a/integration_tests/tests/admin/userManagement.cy.ts
+++ b/integration_tests/tests/admin/userManagement.cy.ts
@@ -1,4 +1,4 @@
-import { ApprovedPremisesUserRole, UserQualification } from '../../../server/@types/shared'
+import { ApprovedPremisesUserRole, UserQualification } from '@approved-premises/api'
 import paths from '../../../server/paths/admin'
 import { userFactory } from '../../../server/testutils/factories'
 import ConfirmDeletionPage from '../../pages/admin/userManagement/confirmDeletionPage'
@@ -334,28 +334,5 @@ context('User management', () => {
 
     // Then the page should show the results for the second page
     page.shouldShowUsers(usersForResultsPage2)
-  })
-
-  it('retrieve the updated user from api if version not matched to user in session and shows the new role added report viewer', () => {
-    // Given there are users in the DB
-    const users = userFactory.buildList(10, { roles: ['assessor'] })
-    const user = userFactory.build({ version: 345, roles: ['report_viewer'] })
-    const roles = ['manager', 'matcher', 'workflow_manager']
-    cy.task('stubFindUser', { user, id: users[0].id })
-    cy.task('stubUsers', { users })
-    cy.task('stubApAreaReferenceData')
-    cy.task('stubAuthUser', { roles })
-
-    // When I visit the list page
-    const listPage = ListPage.visit()
-
-    // Then I should see the users and their details
-    listPage.shouldShowUsers(users)
-
-    // And I click on user
-    listPage.clickUser(users[0].name)
-
-    // Then I should see the user with new report role
-    listPage.shouldShowReportsMenu()
   })
 })
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "install-playwright": "npx playwright install",
     "security_audit": "npx audit-ci --config audit-ci.json",
     "int-test": "cypress run --config video=false",
-    "int-test-ui": "cypress open --e2e --browser chrome",
+    "int-test-ui": "cypress open --e2e --browser electron",
     "clean": "rm -rf dist build node_modules stylesheets",
     "start-test-wiremock": "docker-compose -f docker-compose-test.yml up -d",
     "generate-types": "script/generate-types",
--- a/server/@types/ui/index.d.ts
+++ b/server/@types/ui/index.d.ts
@@ -45,6 +45,7 @@ import {
 interface TasklistPage {
   body: Record<string, unknown>
 }
+
 interface PersonService {}
 
 // A utility type that allows us to define an object with a date attribute split into
@@ -374,7 +375,7 @@ export type UserDetails = {
   permissions: Array<ApprovedPremisesUserPermission>
   active: boolean
   apArea: ApArea
-  version: number
+  version: string
 }
 
 export type PartnerAgencyDetails = {
--- a/server/@types/ui/index.d.ts
+++ b/server/@types/ui/index.d.ts
@@ -374,6 +374,7 @@ export type UserDetails = {
   permissions: Array<ApprovedPremisesUserPermission>
   active: boolean
   apArea: ApArea
+  version: number
 }
 
 export type PartnerAgencyDetails = {
--- a/server/data/userVersionMiddleware.ts
+++ b/server/data/userVersionMiddleware.ts
@@ -1,0 +1,18 @@
+/* istanbul ignore file */
+
+import { Response, SuperAgentRequest } from 'superagent'
+import inMemoryStore from '../inMemoryStore'
+
+function userVersionMiddleware(agent: SuperAgentRequest) {
+  agent.on('request', ({ req }) => {
+    req.on('response', (res: Response, err: Error) => {
+      res.on('end', () => {
+        inMemoryStore.userVersion = res.headers['x-cas-user-version']
+      })
+    })
+  })
+
+  return agent
+}
+
+export { userVersionMiddleware }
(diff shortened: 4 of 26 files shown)


server/middleware/populateCurrentUser.ts
```diff
@@ -1,12 +1,21 @@
...
+
 import UserService, { DeliusAccountMissingStaffDetailsError } from '../services/userService'
+import inMemoryStore from '../inMemoryStore'
 
 export default function populateCurrentUser(userService: UserService): RequestHandler {
   return async (req, res, next) => {
     try {
       if (res.locals.user) {
-        const user = req.session.user || (await userService.getActingUser(res.locals.user.token))
+        let user = req.session.user || (await userService.getActingUser(res.locals.user.token))
```
richpjames, line 11: Would line 10-18 be clearer if it was written like this?
```    
let user: UserDetails | undefined = res.locals.user || req.session.user

      if (user?.version?.toString() !== inMemoryStore.userVersion) {
        user = await userService.getActingUser(res.locals.user.token)
        logger.info(`****** Updated user via API ****** from ${inMemoryStore.userVersion}  to ${user.version}`)
      }

```

server/middleware/populateCurrentUser.ts
```diff
@@ -1,12 +1,21 @@
...
+        const { userVersion } = inMemoryStore
+
+        if (user && user.version && user.version.toString() !== userVersion) {
```
richpjames, line 14: It would be good to test this functionality

server/data/restClientMetricsMiddleware.ts
```diff
@@ -33,6 +34,7 @@ function restClientMetricsMiddleware(agent: SuperAgentRequest) {
     req.on('response', (res: Response, err: Error) => {
       res.on('end', () => {
         const responseTime = Date.now() - startTime
+        inMemoryStore.userVersion = res.headers['x-cas-user-version']
```
richpjames, line 37: This doesn't seem to be related to client metrics, would it be better to create a new middleware to do this?

server/@types/ui/index.d.ts
```diff
@@ -374,6 +374,7 @@ export type UserDetails = {
   permissions: Array<ApprovedPremisesUserPermission>
   active: boolean
   apArea: ApArea
+  version: number
```
richpjames, line 377: should this be a `string`?

richpjames:
I think it would be good to write an integration test for this functionality

server/data/restClientResponseMiddleware.ts
```diff
@@ -0,0 +1,18 @@
+/* istanbul ignore file */
+
+import { Response, SuperAgentRequest } from 'superagent'
+import inMemoryStore from '../inMemoryStore'
+
+function restClientResponseMiddleware(agent: SuperAgentRequest) {
+  agent.on('request', ({ req }) => {
+    req.on('response', (res: Response, err: Error) => {
```
richpjames, line 8: Might `userVersionMiddleware` be a more accurate name? 
A lot of middleware could deal with responses, so it could be helpful to be more specific

server/data/restClientResposeMiddleware.test.ts
```diff
@@ -0,0 +1,22 @@
...
+import { restClientMetricsMiddleware } from './restClientMetricsMiddleware'
+import InMemoryStore from '../inMemoryStore'
+
+describe('restClientResponseMiddleware', () => {
+  afterEach(() => {
+    jest.clearAllMocks()
+  })
+
+  describe('response headers', () => {
+    it('headers handling', async () => {
```
richpjames, line 12: Might it be helpful to describe _how_ the function handles headers?

server/inMemoryStore.ts
```diff
@@ -0,0 +1,8 @@
+interface InMemoryStore {
+  userVersion: string
+}
+
+const inMemoryStore: InMemoryStore = {
```
richpjames, line 5: I think it would be helpful to document what sort of data we expect to keep in the `inMemoryStore`, how that differs from the data kept in `res.locals` and `req.session`

server/middleware/populateCurrentUser.test.ts
```diff
@@ -44,10 +45,25 @@ describe('populateCurrentUser', () => {
     expect(next).toHaveBeenCalled()
   })
 
+  it('should populate the current user from the API if the user version not equals', async () => {
```
richpjames, line 48: Could it be helpful to specify what the user version is not equal to?

server/middleware/populateCurrentUser.test.ts
```diff
@@ -44,10 +45,25 @@ describe('populateCurrentUser', () => {
...
+    await middleware(request, response, next)
+
+    expect(request.session.user).toEqual(user)
+    expect(response.locals.user).toEqual({ ...response.locals.user, ...user })
+
+    expect(userService.getActingUser).toHaveBeenCalledWith(token)
+    expect(next).toHaveBeenCalled()
+  })
+
   it('should fetch the user from the session if present', async () => {
```
richpjames, line 62: might `should fetch the user from the session if present and the user version matched the value in the inMemoryStore` be more accurate?

integration_tests/tests/admin/userManagement.cy.ts
```diff
@@ -335,4 +335,27 @@ context('User management', () => {
     // Then the page should show the results for the second page
     page.shouldShowUsers(usersForResultsPage2)
   })
+
+  it('retrieve the user from api if version not matched to user in session', () => {
```
richpjames, line 339: the test description doesn't describe the assertions the tests takes.   
Could we assert: 
- if the user version in the response header is different to the value we have locally then a call to `getActingUser` is made and the local version is updated?

richpjames (changes requested):
I've left some more comments. I think it's really important this work is tested thoroughly at all levels because it is going to be hit so frequently (on every request to the API) and it deals with security.

server/@types/ui/index.d.ts
davidatkinsuk: i think number is fine, we convert an integer to a string when writing this to the header in the API

vubalasu:
yes you can update user_role_assignments by removing the role or adding and check in UI that role is updated but as of now it is getting updated on for second request as agreed earlier

davidatkinsuk:
> yes you can update user_role_assignments by removing the role or adding and check in UI that role is updated but as of now it is getting updated on for second request as agreed earlier

i guess you could also login in a different browser and modify your user roles and then see menu options appear/disappear in your original browser when you refresh the page?

server/data/userVersionMiddleware.ts
```diff
@@ -0,0 +1,18 @@
+/* istanbul ignore file */
+
+import { Response, SuperAgentRequest } from 'superagent'
+import inMemoryStore from '../inMemoryStore'
+
+function userVersionMiddleware(agent: SuperAgentRequest) {
+  agent.on('request', ({ req }) => {
+    req.on('response', (res: Response, err: Error) => {
+      res.on('end', () => {
+        inMemoryStore.userVersion = res.headers['x-cas-user-version']
```
froddd, line 10: As discussed, it seems in-memory storage is shared between all sessions, so the value above would be unique for all current users and continually change, which would result in excessive calls to refresh user details. Suggestion would be to use a different session key which does not get updated with user details:

```suggestion
        req.session.userVersion = res.headers['x-cas-user-version']
```

froddd (changes requested):

server/inMemoryStore.ts
froddd: Also discussed is that this approach does not work, as the in-memory storage is shared between all sessions. Will be updated.

integration_tests/tests/admin/userManagement.cy.ts
```diff
@@ -335,4 +335,27 @@ context('User management', () => {
     // Then the page should show the results for the second page
     page.shouldShowUsers(usersForResultsPage2)
   })
+
+  it('retrieve the updated user from api if version not matched to user in session and shows the new role added report viewer', () => {
```
froddd, line 339: This test might be clearer if 2 successive API calls are stubbed with a different `x-cas-user-version` header, to show that there's a change. At the moment there is so much interaction around users it muddies the readability -- I would suggest 2 successive visits of the same page (even the home page would be enough) should be sufficient.

server/@types/ui/index.d.ts
froddd: If the UI is only dealing with the header value it would be better to have this typed as string, to avoid excessive type conversions.

server/data/userVersionMiddleware.ts
froddd: As discussed, the above isn't doable directly -- this is a call to the API using Superagent, so the `req` object here is not the same as the user session!

package.json
```diff
@@ -34,7 +34,7 @@
     "install-playwright": "npx playwright install",
     "security_audit": "npx audit-ci --config audit-ci.json",
     "int-test": "cypress run --config video=false",
-    "int-test-ui": "cypress open --e2e --browser chrome",
+    "int-test-ui": "cypress open --e2e --browser electron",
```
bobmeredith, line 37: Is this intentional?  It seems a big step as I assume chrome is our top of our supported browser list.

aliuk2012 (approved):
I haven't tested these changes locally but from my limited understanding of the service and what has been changed here does make sense.

server/middleware/populateCurrentUser.ts
```diff
@@ -1,14 +1,27 @@
...
+          if (user) {
+            logger.info(
+              `Updated user from API${currentVersion !== user.version ? ` (version: ${currentVersion} -> ${user.version})` : ''}`,
+            )
+          }
+        }
+
         req.session.user = user
-        res.locals.user = { ...user, ...res.locals.user }
+        res.locals.user = { ...res.locals.user, ...user }
```
bobmeredith, line 24: Shouldn't these two lines be inside the user version changed test block? I guess it would just be overwriting the session and local user with the same data but it seems a bit pointless.

bobmeredith (approved):
I made a couple of minor comments but looks OK to me.

package.json
froddd: Yes -- as per commit message here: https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/2056/commits/144054bc891db4ce4e6a35c75f03f9eeeeaade53

Electron uses Chromium anyway, so any rendering and functionality should be on par with Chrome. This is mostly a convenience change, which prevents Cypress from refocusing the test window whenever a new test starts running.

server/middleware/populateCurrentUser.ts
froddd: I guess that's correct for assigning `user` to `req.session.user`, yes!

But for the assignment to locals, if we're populating the user details from what is available in the session, we still want this data to be added to `res.locals` so it can be used in templates.

PR #2058, opened 2024-08-09
APS-1167 Enable out of service beds report

davidatkinsuk:
The out of service beds report will replace lost beds. As such, we have deprecated that report in this commit.

# Context

# Changes in this PR

### Before

Screenshot 2024-08-09 at 09 18 52

### After

Screenshot 2024-08-09 at 09 17 42


RobBoothMOJ (approved):
Approved

PR #2064, opened 2024-08-12
Reinstate short-notice-application test

richpjames:
Previously we were trying to inspect the textContent of an element that was sometimes undefined which caused an error. Now we check if the item is defined before extracting the text content. 
This enables us to reinstate the short notice application test previously removed #2063


patrickjfl (approved):
LGTM 👍

PR #2068, opened 2024-08-12
Aps 1145 -  Search screen iterations

richpjames:
Small iterations and improvements to search screen 
[JIRA](https://dsdmoj.atlassian.net/jira/software/c/projects/APS/boards/1328?assignee=62a050ba9f5d480069c97f49&selectedIssue=APS-1145)

### Before
image

### After
after (3)

### DIFF
--- a/server/testutils/factories/spaceSearchResults.ts
+++ b/server/testutils/factories/spaceSearchResults.ts
@@ -6,7 +6,7 @@ import spaceSearchParameters from './spaceSearchParameters'
 import spaceSearchResult from './spaceSearchResult'
 
 export default Factory.define<Cas1SpaceSearchResults>(() => {
-  const numberOfResults = faker.number.int({ min: 1, max: 50 })
+  const numberOfResults = faker.number.int({ min: 1, max: 10 })
   return {
     searchCriteria: spaceSearchParameters.build(),
     resultsCount: numberOfResults,
--- a/server/utils/match/index.ts
+++ b/server/utils/match/index.ts
@@ -282,7 +282,7 @@ export const groupedCriteria = {
     inputName: 'spaceCharacteristics',
   },
   accessNeeds: {
-    title: 'AP & room characteristics',
+    title: 'Access needs and additional features',
     items: placementRequirementCriteriaLabels,
     inputName: 'spaceCharacteristics',
   },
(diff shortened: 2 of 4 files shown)


server/utils/match/index.ts
```diff
@@ -205,21 +204,15 @@ export const apTypeRow = (apType: ApType) => ({
...
-  },
-})
-
 export const addressRow = (spaceSearchResult: SpaceSearchResult) => ({
   key: {
     text: 'Address',
   },
   value: {
-    text: `${spaceSearchResult.premises.addressLine1} ${spaceSearchResult.premises.addressLine2}`,
+    html: `<p>
```
patrickjfl, line 213: I was going to suggest this the other day - good approach I reckon

server/utils/match/index.ts
```diff
@@ -216,14 +216,17 @@ export const addressRow = (spaceSearchResult: SpaceSearchResult) => ({
...
-export const distanceRow = (spaceSearchResult: SpaceSearchResult, postcodeArea?: string) => ({
-  key: {
-    text: 'Distance',
-  },
-  value: {
-    text: `${spaceSearchResult.distanceInMiles} miles from ${postcodeArea || 'the desired location'}`,
-  },
-})
+export const distanceRow = (spaceSearchResult: SpaceSearchResult, postcodeArea?: string) => {
+  const roundedDistanceInMiles = Math.round(spaceSearchResult.distanceInMiles * 10) / 10
```
patrickjfl, line 230: I think this is fine but I guess the flipside is that a postcode area can probably cover a significant area itself, so I'm not completely sold on the value of the decimal point.

server/testutils/factories/spaceSearchResults.ts
```diff
@@ -6,7 +6,7 @@ import spaceSearchParameters from './spaceSearchParameters'
 import spaceSearchResult from './spaceSearchResult'
 
 export default Factory.define<Cas1SpaceSearchResults>(() => {
-  const numberOfResults = faker.number.int({ min: 1, max: 50 })
+  const numberOfResults = faker.number.int({ min: 1, max: 10 })
```
patrickjfl, line 9: nice

patrickjfl (approved):
LGTM

server/utils/match/index.ts
richpjames: true. I think if it came down to it being such a similar distance then the matcher would use other factors than distance to make the decision too

PR #2069, opened 2024-08-13
Remove 'Create a placement' link for users with future_manager role

richpjames:
This link was showing in 'V2 Manage' which isn't needed and could be confusing to users.
We need to remove it.
Only users with the role 'future_manager' will see V2 Manage and those with that role will only see V2 Manage and not V1 so this logic should work for what we need.


davidatkinsuk (approved):
LGTM 👍

PR #2070, opened 2024-08-13
Remove 'Create a placement' link entirely

richpjames:
#2069 removed the 'Create a placement' link for users with the 'workflow_manager' role but without the 'future_manager' role. We now want to remove the link for all users.


richpjames:
This will need to include a rewrite of how we do the manage.spec.ts e2e tests

PR #2072, opened 2024-08-13
APS-1051: LAO E2E integration tests

vubalasu:
# Context

https://dsdmoj.atlassian.net/browse/APS-1051

# Changes in this PR
LAO E2E integration tests

## Screenshots of UI changes
No ui changes

### DIFF
--- a/integration_tests/tests/apply/dashboard.cy.ts
+++ b/integration_tests/tests/apply/dashboard.cy.ts
@@ -3,6 +3,7 @@ import {
   applicationSummaryFactory,
   placementApplicationFactory,
   requestForPlacementFactory,
+  restrictedPersonFactory,
 } from '../../../server/testutils/factories'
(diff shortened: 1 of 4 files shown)


integration_tests/tests/apply/dashboard.cy.ts
```diff
@@ -60,6 +61,24 @@ context('All applications', () => {
     page3.shouldShowApplications()
   })
 
+  it('lists all applications for lao', () => {
+    // Given I am logged in
+    cy.signIn()
+
+    // Given there are multiple pages of applications
```
richpjames, line 68: There's only 1 page here

richpjames (approved):
It's not clear from these tests if the behaviour for LAOs is any different to the behaviour for `FullPerson`s. 
We could assert in the tests the differences between the two. EG on the application dashboard an LAO will show the CRN instead of the name and a link to the application. Do you think that would be worthwhile?

PR #2076, opened 2024-08-14
Remove Flipt feature flags 

richpjames:
The instance of Flipt we were using [has been deprecated](https://mojdt.slack.com/archives/C03K0HB0LBE/p1716212368386669).  
[We have removed our integration with Flipt in the API](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/2094) but we still call the service in the UI. 
There were two remaining flags: 
- `show-search-by-CRN-timeline-navigation` - this one has been _on_ in prod for a while. It has successfully bedded in and can be safely removed. We should show this nav tile for all users. 
- `allow-sufficient-information-request-without-confirmation` - this one never made it to production. Whilst testing it we found it did not successfully avoid the infinite loop it was put in to prevent so it can be removed with the code now permanently behaving as if the feature switch was off.


server/controllers/admin/cruDashboardController.test.ts
```diff
@@ -4,7 +4,7 @@ import { DeepMocked, createMock } from '@golevelup/ts-jest'
 import { when } from 'jest-when'
 import CruDashboardController from './cruDashboardController'
 
-import { ApAreaService, ApplicationService, FeatureFlagService, PlacementRequestService } from '../../services'
```
patrickjfl, line 7: I assume the FeatureFlagService is still being used elsewhere, and that's why you aren't removing it entirely?

patrickjfl (approved):
Ah sorry, you answered my question in the following commit. Nice job 👍

PR #2080, opened 2024-08-15
APS-1029 : Iterate CRU Staff Role

RobBoothMOJ:
Utilise added permissions for CRU, Janitor and Future Manager roles to provide / allow role-specific functionality within the UI.   The CRU and Janitor roles have the same permissions, the Future Manager role has permissions relating to out of service beds, to maintain consistency with current functionality.

New permissions added to control displaying of tiles on CAS1 Dashboard :
CAS1_VIEW_CRU_DASHBOARD - display the 'CRU Dashboard' tile
CAS1_VIEW_MANAGE_TASKS - display the 'Manage task Allocations' tile
CAS1_VIEW_OUT_OF_SERVICE_BEDS - display the 'view out of service beds' tile

New permissions governing permitted behaviour within the application :
CAS1_ADHOC_BOOKING_CREATE - user will be able to create an ad-hoc booking
CAS1_BOOKING_CREATE - user will be able to create a booking
CAS1_BOOKING_WITHDRAW - user will be able to withdraw a booking

CAS1_OUT_OF_SERVICE_BED_CREATE - user will have access to forms allowing creation of out of service beds

[Jira Ticket](https://dsdmoj.atlassian.net/browse/APS-1029)

# Changes in this PR
Implementing changes to permissions which incorporated tiles being displayed / not displayed in the CAS1 Dashboard as outlined above. 
Additionally, action options were removed and also changed to authentication which takes place as part of the routing.  Instead of checking for roles, some routes are now checking for permissions.

# E2E Tests running locally 
The tests required a manual change to the 'assessor' user in the DB who needed to have the 'CAS1_CRU_MEMBER' role added.   This shouldn't affect the e2e tests in dev as the corresponding user - AP_USER_TEST_1 has the 'CRU Member' role 

## Screenshots of UI changes

Restricted tiles (missing permissions for user)
no-access

Displayed tiles (permissions on user record (CRU-Member role))

CRU-member role permissions added

No permission to allow creation of bookings
cant_book

With permission to be able to create bookings
can_book

No permission to allow creation of ad-hoc bookings
cant_book_adhoc

Permission added which allows creation of ad-hoc bookings 
can_book_adhoc

No permission to withdraw placements
cannot_withdraw_placement

Added permission to withdraw placements
can_withdraw_placement


server/middleware/applicationAuthMiddleware.ts
```diff
@@ -1,12 +1,16 @@
...
+import { hasPermission } from '../utils/users'
 
 export default function applicationAuthMiddleware(requestHandler: RequestHandler, middlewareSpec?: MiddlewareSpec) {
-  if (middlewareSpec?.allowedRoles) {
+  if (middlewareSpec?.allowedRoles || middlewareSpec?.allowedPermissions) {
     return async (req: Request, res: Response, next: NextFunction) => {
-      if (res.locals.user.roles.some((role: ApprovedPremisesUserRole) => middlewareSpec.allowedRoles.includes(role))) {
+      if (
+        res.locals.user.roles.some((role: ApprovedPremisesUserRole) => middlewareSpec.allowedRoles.includes(role)) ||
+        hasPermission(res.locals.user, middlewareSpec.allowedPermissions ?? new Array<ApprovedPremisesUserPermission>())
```
richpjames, line 12: This isn't a blocking comment as I know that the function wasn't added in this PR but it seems a little off to me that `hasPermission` takes an array of Permissions but it's name implies that it tests only one permission. It is also only ever called with an array containing 1 item. This difference in its signature with `hasRole` which makes this `if` harder to read than it needs to be.  
Separately, could remove  `?? new Array<ApprovedPremisesUserPermission>()` it isn't necessary as we check that `middlewareSpec.allowedPermissions` is truthy on line 8, therefore (according to its type) it must be an array.

integration_tests/tests/tasks/list.cy.ts
```diff
@@ -16,6 +17,21 @@ context('Task Allocation', () => {
...
+  })
+
+  it('returns unauthorised if user does not have the cas1 view manage task permission', () => {
+    // Given I am logged in without the required permissions
+    const me = userFactory.build({ apArea })
+    cy.task('stubAuthUser', { roles: [], permissions: [], userId: me.id, apArea: me.apArea })
+    cy.signIn()
+
+    const path = paths.tasks.index({})
+    cy.request({ url: path, failOnStatusCode: false }).should(response => {
```
richpjames, line 32: 👏

server/middleware/applicationAuthMiddleware.test.ts
```diff
@@ -47,4 +47,36 @@ describe('applicationAuthMiddleware', () => {
     expect(response.status).toHaveBeenCalledWith(401)
     expect(response.render).toHaveBeenCalledWith('roleError')
   })
+
+  it('returns the handler when there are allowedPermissions specified and the user has one of those permissions', async () => {
+    const user = userDetailsFactory.build({ permissions: ['cas1_booking_create'] })
+    const response = createMock<Response>({ locals: { user } })
+
+    const auditedhandler = applicationAuthMiddleware(handler, {
+      allowedRoles: ['manager', 'matcher'],
```
richpjames, line 56: do you think it's worth doing `allowedRoles: []` for clarity?

server/middleware/applicationAuthMiddleware.test.ts
```diff
@@ -47,4 +47,36 @@ describe('applicationAuthMiddleware', () => {
...
+      allowedRoles: [],
+      allowedPermissions: ['cas1_booking_withdraw', 'cas1_process_an_appeal'],
+    })
+
+    await auditedhandler(request, response, next)
+
+    expect(loggerSpy).toHaveBeenCalledWith('User is not authorised to access this')
+    expect(response.status).toHaveBeenCalledWith(401)
+    expect(response.render).toHaveBeenCalledWith('roleError')
+  })
```
richpjames, line 81: Do you think it's worth adding a test that calls `applicationAuthMiddleware` with just the `handler`? 
The second parameter is optional, it might be worth demonstrating that the function can handle the `allowedPermissions` and `allowedRoles` being `undefined`.

richpjames (commented):
This is really good Rob, thorough and clear testing makes the review straight forward.   
I've left some comments

server/middleware/applicationAuthMiddleware.ts
RobBoothMOJ: Taken action to rename `hasPermission` to hasPermissions in a separate PR.
Taken action to improve readability of code to cover role check as well as permissions.

richpjames (approved):

PR #2087, opened 2024-08-21
Remove ‘manager’ and ‘legacy_manager’ options

davidatkinsuk:
These two roles should no longer be used going forward. This commit removes them as options from the manage user page to ensure they are no longer assigned to users. Once this change is in production the roles will be removed from the API and we can then remove all references to them in the UI code.

# Context

# Changes in this PR

### Before

Screenshot 2024-08-21 at 14 40 54

### After

Screenshot 2024-08-21 at 14 39 35


RobBoothMOJ (approved):
Approved

RobBoothMOJ (approved):
Approved

PR #2089, opened 2024-08-22
APS-1174 - Replace the content on the AP list page with a link to the new destination

Boris-hmpps:
Replace the content on the AP list page with a link to the new destination

# Context
We do not want anyone to access the old screens
We cannot guarantee that users will not find the link to the old screen via bookmarks, browser history, or shared links
Instead of a redirect we’re using a link so that users know that things have moved and changed

# Changes in this PR
Updated view content
Removed integration tests as it is not needed and same test is run in another test with the new url

### Before
Screenshot 2024-08-22 at 15 54 26

### After
Screenshot 2024-08-22 at 15 54 37


RobBoothMOJ (approved):
Approved

PR #2091, opened 2024-08-23
APS-1176 AP Beds replacement

Boris-hmpps:
Replace the content on the AP Beds page with a link to the new destination
# Context
We do not want anyone to access the old screens
We cannot guarantee that users will not find the link to the old screen via bookmarks, browser history, or shared links
Instead of a redirect we’re using a link so that users know that things have moved and changed

# Changes in this PR
Update view

### Before
Screenshot 2024-08-23 at 16 50 09

### After
Screenshot 2024-08-23 at 16 48 28


Rosie-Brigham (approved):
LGTM

PR #2107, opened 2024-09-04
Update generate types to get cas1 specific types

davidatkinsuk:
The API definitions have recently changes to start moving cas1 specific API types into a cas1 specific API file. This commit changes the generate types script to include that file when generating types

This PR also pulls in the latest API definitions


Boris-hmpps (approved):
Looks good to me.

PR #2113, opened 2024-09-06
APS-1245 - Fix e2e issues

davidatkinsuk:
* Update README to reflect changes to test users, and to explain how to test emails locally
* Adds an option to disable checking emails when running locally to avoid needing to configure the API to send emails to notify. Also adds more information when assertions fail to help debug issues


RobBoothMOJ (approved):
Approved

PR #2118, opened 2024-09-09
APS-1275 enable legacy placements alongside space booking

davidatkinsuk:
# Context

# Changes in this PR

Whilst we develop the ‘search a space functionality’, we should ensure that the ‘Create placement’ functionality is always available to allow us to test the functionality that is currently in production

Given this change, we can now enable match and manage v2 functionality in pre-production, as it no longer blocks access to the placement functionality provided in production

### Before

Screenshot 2024-09-09 at 15 27 34

### After

Screenshot 2024-09-09 at 15 20 28


RobBoothMOJ (approved):
approved

PR #2120, opened 2024-09-10
APS-1277 - Re-instate change booking date on view placement

davidatkinsuk:
This was removed as part of migration to future manager. This is still a required function so has been re-added via a permission assigned to future manager. Note that this functonality can also be accessed via the CRU Dashboard Placement Request View

# Context

### Before

Screenshot 2024-09-10 at 08 52 57

### After

Screenshot 2024-09-10 at 08 51 00


RobBoothMOJ (approved):
Approved

RobBoothMOJ (approved):
Approved

PR #2122, opened 2024-09-10
Remove LEGACY_MANAGER from E2E tests

davidatkinsuk:
This role is no longer in use.

Instead, the tests making use of this role should instead use a CRU Member

Screenshot 2024-09-10 at 15 20 49


RobBoothMOJ (approved):
Approved

PR #2128, opened 2024-09-11
APS-1259:M&M V2 - View Premises - UI - Use API /cas1/premises/$premis…

vubalasu:
…esId to retrieve premise info for /manage/premises/$premisesId

# Context

https://dsdmoj.atlassian.net/browse/APS-1259

# Changes in this PR
View Premises - UI - Use API /cas1/premises/$premisesId to retrieve premise info for /manage/premises/$premisesId
‘Available Beds’ fields should be removed from the UI

### Before
Screenshot 2024-09-11 at 14 17 55

### After
Screenshot 2024-09-11 at 14 13 57


froddd:
Could you rebase to keep the types updates separate? If necessary get those types updates merged into main first.

vubalasu:
Generally If we have any updates form types we will do as part of the pr instead of creating new p

froddd:
Ok -- we do the opposite on CAS3, we ensure any types required are merged into master ahead of the feature work to avoid having a PR with changes which are not directly relevant to it. I'll probably suggest we do this from now on, I'll also change the way the type updates PRs are created: we have a single branch for it which gets updated, which removes the need to close loads of out-of-date branches all the time.

froddd (approved):
A couple of extra things can be cleaned up but this should be good to go

PR #2129, opened 2024-09-11
APS-1206 Allow bookings with no characteristics defined

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/APS-1206

# Changes in this PR

Sends empty arrays to spaceBookingCreate if no essential or desirable criteria are defined, rather than arrays containing an empty string.
Parameterised the test to test both with populated criteria and blanks.  


bobmeredith:
I'll add a jest test to cover this change.

froddd (approved):
👍

PR #2140, opened 2024-09-13
APS-1227 Conflict error for out-of-service beds with correct link

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/APS-1227

# Changes in this PR

The handling errors in this area appears to be a bit sensitive to changes. There is code that detects the entity causing the conflict and the id of that entity by parsing the returned error details. It seems that the error details have changed from citing 'Lost beds' to 'out-of-service beds' which caused the error handling to default to its 'bookings' mode.
Also, the path needed updating for M&M v2.

I switched to regexp to do the parsing and switched some literal strings to enums, but this whole mechanism remains a bit fragile.

Note, the error texts on the date fields are wrong, but this was not mentioned in the ticket so I'll create a new ticket for this fix.

### After
chrome-capture-2024-9-16

### DIFF
--- a/integration_tests/components/bedspaceConflictErrorComponent.ts
+++ b/integration_tests/components/bedspaceConflictErrorComponent.ts
@@ -3,17 +3,18 @@ import errorLookups from '../../server/i18n/en/errors.json'
 import Page from '../pages/page'
 import BookingShowPage from '../pages/manage/booking/show'
 import { OutOfServiceBedShowPage } from '../pages/v2Manage/outOfServiceBeds/outOfServiceBedShow'
+import { EntityType } from '../../server/@types/ui'
 
 export default class BedspaceConflictErrorComponent {
   constructor(
     private readonly premisesId: string,
-    private readonly source: 'booking' | 'lost-bed',
+    private readonly source: EntityType,
   ) {}
 
   shouldShowDateConflictErrorMessages(
     fields: Array<string>,
     conflictingEntity: Booking | LostBed | Cas1OutOfServiceBed,
-    conflictingEntityType: 'booking' | 'lost-bed',
+    conflictingEntityType: EntityType,
   ): void {
     fields.forEach(field => {
       cy.get(`[data-cy-error-${field}]`).should('contain', errorLookups[field].conflict)
@@ -36,7 +37,7 @@ export default class BedspaceConflictErrorComponent {
     cy.go('back')
   }
 
-  private getTitle(fields: Array<string>, conflictingEntityType: 'booking' | 'lost-bed'): string {
+  private getTitle(fields: Array<string>, conflictingEntityType: EntityType): string {
     return (
       conflictingEntityType === 'lost-bed'
         ? 'Out of service bed record cannot be created for the $date$ entered'
@@ -44,7 +45,7 @@ export default class BedspaceConflictErrorComponent {
     ).replace('$date$', fields.length === 1 ? 'date' : 'dates')
   }
 
-  private getMessage(fields: Array<string>, conflictingEntityType: 'booking' | 'lost-bed'): string {
+  private getMessage(fields: Array<string>, conflictingEntityType: EntityType): string {
     const noun = conflictingEntityType === 'booking' ? 'booking' : 'out of service beds record'
 
     return fields.length === 1 ? `It conflicts with an existing ${noun}` : `They conflict with an existing ${noun}`
--- a/integration_tests/components/bedspaceConflictErrorComponent.ts
+++ b/integration_tests/components/bedspaceConflictErrorComponent.ts
@@ -30,7 +30,7 @@ export default class BedspaceConflictErrorComponent {
     if (conflictingEntityType === 'booking') {
       Page.verifyOnPage(BookingShowPage, [this.premisesId, conflictingEntity as Booking])
     } else {
-      Page.verifyOnPage(OutOfServiceBedShowPage, this.premisesId, conflictingEntity as OutOfServiceBed)
+      Page.verifyOnPage(OutOfServiceBedShowPage, this.premisesId, conflictingEntity as Cas1OutOfServiceBed)
     }
 
     cy.go('back')
--- a/integration_tests/components/bedspaceConflictErrorComponent.ts
+++ b/integration_tests/components/bedspaceConflictErrorComponent.ts
@@ -2,7 +2,7 @@ import type { Booking, Cas1OutOfServiceBed, LostBed } from '@approved-premises/a
 import errorLookups from '../../server/i18n/en/errors.json'
 import Page from '../pages/page'
 import BookingShowPage from '../pages/manage/booking/show'
-import LostBedShowPage from '../pages/manage/lostBedShow'
+import { OutOfServiceBedShowPage } from '../pages/v2Manage/outOfServiceBeds/outOfServiceBedShow'
 
 export default class BedspaceConflictErrorComponent {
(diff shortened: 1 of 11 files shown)


server/utils/bookings/index.ts
```diff
@@ -21,9 +21,14 @@ import { bookingActions, v1BookingActions, v2BookingActions } from './bookingAct
 
 const UPCOMING_WINDOW_IN_DAYS = 365 * 10
 
+enum ConflictingEntityType {
+  booking,
+  lostBed,
+}
+
```
froddd, line 28: I don't think the codebase uses enums anywhere, because 'enums are evil' or whatever -- not sure I want to go down this debate, but, best keep to union types, if anything, for consistency.

server/utils/bookings/index.ts
```diff
@@ -179,33 +184,34 @@ export const generateConflictBespokeError = (
...
-
+  /**
+   *  Return the entity type and id by parsing an error detail string
+   *  @param detail - string is text containing the entity id at the end preceded by ': '
+   *    e.g. "An out-of-service bed already exists for dates from 2024-10-01 to 2024-10-14 which overlaps with the desired dates: 220a71da-bf5c-424d-94ff-254ecac5b857"
+   */
+  const matchResult = /([0-9a-z-]*)$/.exec(detail)
+  const conflictingEntityId = matchResult && matchResult[1]
+  const isOutOfServiceBedEntity = /(?:lost bed)|(?:out-of-service bed)/i.test(detail)
+  const conflictingEntityType = isOutOfServiceBedEntity ? ConflictingEntityType.lostBed : ConflictingEntityType.booking
```
froddd, line 214: This seems complex, and we shouldn't need to handle the 'Lost bed' scenario anymore.

CAS3 simply splits the string by `:` to parse the entity ID, then searches for a simpler match in the detail:

```suggestion
  /**
   *  Return the entity type and id by parsing an error detail string
   *  @param detail - string is text containing the entity id at the end preceded by ': '
   *    e.g. "An out-of-service bed already exists for dates from 2024-10-01 to 2024-10-14 which overlaps with the desired dates: 220a71da-bf5c-424d-94ff-254ecac5b857"
   */
  const [ message, conflictingEntityId ] = details.split(':').map(trim)
  const conflictingEntityType = message.includes('out-of-service bed') ? ConflictingEntityType.lostBed : ConflictingEntityType.booking
```

With the comment above about not using enums, this last line would then return either `lost-bed` or `booking` instead of the enum syntax.

server/utils/bookings/index.test.ts
```diff
@@ -220,22 +220,23 @@ describe('bookingUtils', () => {
       })
     })
 
-    it('generates a bespoke error when there is a conflicting lost bed', () => {
+    it('generates a bespoke error when there is a conflicting out-of-service bed', () => {
       const err = {
         data: {
-          detail: `Conflicting Lost Bed: ${lostBedId}`,
+          detail: `Conflicting out-of-service bed: ${lostBedId}`,
```
froddd, line 226: Is this the actual error detail that comes back from the API?

bobmeredith: No, actually I abbreviated it. You're right - would probably be better with an actual cut'n'paste from the API.

server/utils/bookings/index.ts
bobmeredith: I'm not 100% convinced - mainly because I like regex so much. but I've changed to use your approach.

integration_tests/components/bedspaceConflictErrorComponent.ts
```diff
@@ -30,20 +30,22 @@ export default class BedspaceConflictErrorComponent {
...
-      : 'This bedspace is not available for the dates entered'
+  private getTitle(fields: Array<string>, conflictingEntityType: 'booking' | 'lost-bed'): string {
+    return (
+      conflictingEntityType === 'lost-bed'
+        ? 'Out of service bed record cannot be created for the $date$ entered'
+        : 'This bedspace is not available for the $date$ entered'
+    ).replace('$date$', fields.length === 1 ? 'date' : 'dates')
   }
 
   private getMessage(fields: Array<string>, conflictingEntityType: 'booking' | 'lost-bed'): string {
```
froddd, line 47: The type definition for `ConflictingEntityType` already exists elsewhere, doesn't it? It shouldn't have to be defined here

bobmeredith:
> There's an integration test for a conflict with a booking when creating an out of service bed -- could we add one for other OOSB conflict, too?

Yup, there was actually a test for this but it didn't fail when I made the changes (which it should have). Because the API mock didn't change. As the test mimics the behaviour of the code, it reverted to generating and checking a booking error with a link to a booking. I guess this is the kind of issue caused by using complex code in tests rather than keeping it simple.
I've refactored that so it tests this OOSB conflict and link to the OOSB view page.

integration_tests/components/bedspaceConflictErrorComponent.ts
```diff
@@ -30,20 +30,22 @@ export default class BedspaceConflictErrorComponent {
...
     }
 
     cy.go('back')
   }
 
-  private getTitle(fields: Array<string>): string {
-    return fields.length === 1
-      ? 'This bedspace is not available for the date entered'
-      : 'This bedspace is not available for the dates entered'
+  private getTitle(fields: Array<string>, conflictingEntityType: 'booking' | 'lost-bed'): string {
```
froddd, line 39: The type definition for `ConflictingEntityType` already exists elsewhere, doesn't it? It shouldn't have to be defined here, or in the `getMessage` signature below.

integration_tests/mockApis/utils.ts
```diff
@@ -54,7 +54,7 @@ const errorStub = (fields: Array<string>, pattern: string, method: 'PUT' | 'POST
 const bedspaceConflictResponseBody = (entityId: string | LostBed, entityType: 'booking' | 'lost-bed') => ({
```
froddd, line 54: Again, `entityType` should surely use an existing type here? Also what's going on with `entityId`? That's an odd union 🤔

integration_tests/tests/v2Manage/future_manager/createsOutOfServiceBed.cy.ts
```diff
@@ -85,25 +84,33 @@ context('OutOfServiceBeds', () => {
...
-    signIn([], ['cas1_out_of_service_bed_create'])
+    signIn([], ['cas1_out_of_service_bed_create', 'cas1_view_out_of_service_beds'])
 
+    const bed = { name: 'abc', id: '123' }
     const premises = extendedPremisesSummaryFactory.build()
-    const conflictingBooking = bookingFactory.build()
+    const conflictingOutOfServiceBed = outOfServiceBedFactory.build({
+      bed,
+      startDate: '2022-02-11',
+      endDate: '2022-03-11',
```
froddd, line 96: I don't think it matters that the OOSB built here is defined -- it could be a blank factory build. All we care about is that the response returns a conflict error, I think?

froddd (commented):
Mostly fine now, but the type for conflict entity seems to be recreated in a few places -- would be useful to use a single definition for it.

integration_tests/mockApis/utils.ts
bobmeredith: That's a very strange union as LostBed is an object.
I'm not touching it in this ticket though.

froddd (approved):
👍

PR #2144, opened 2024-09-23
APS-1355: clean up manage v1

froddd:
# Context

This PR removes a lot of unused code, belonging to what is commonly referred to as 'Manage v1'. For a while, 'Manage v1' was running alongside 'Manage v2' (which is current) for some regions, but this is no longer the case.

A lot of the URIs used for 'Manage v1' have equivalent URIs in 'Manage v2': where possible, these have been redirected, and any reference to the old paths have been updated to the new one. The redirect controller logs any redirection.

https://dsdmoj.atlassian.net/browse/APS-1355

Another separate PR will follow to remove unused roles and any code still associated to those


server/utils/bedUtils.test.ts
```diff
@@ -71,57 +66,19 @@ describe('bedUtils', () => {
...
-    describe('when no user is given', () => {
-      const undefinedUser: UserDetails = undefined
-
-      it('returns a "V1" link to manage the room', () => {
-        expect(actionCell(bed, premisesId, undefinedUser)).toEqual({
-          html: v1BedLink(bed, premisesId),
-        })
+    it('returns a link to manage the room', () => {
+      expect(actionCell(bed, premisesId)).toEqual({
+        html: v2BedLink(bed, premisesId),
```
froddd, line 71: In this case, the role has little effect, as the v1 version now produces a redirect to the v2 version, so most of these tests are redundant, and can be removed.

server/utils/calendarUtils.test.ts
```diff
@@ -188,7 +188,7 @@ describe('calendarUtils', () => {
 
       expect(labelForScheduleItem(lostBedOccupancyEntry, premisesId, bedId)).toMatchStringIgnoringWhitespace(
         `<span title="${expectedText}" class="tooltip">
-          <a href="/premises/${premisesId}/beds/${bedId}/lost-beds/${lostBedOccupancyEntry.lostBedId}" class="govuk-link govuk-link--lost_bed">${expectedText}</a>
+          <a href="/manage/premises/${premisesId}/beds/${bedId}/out-of-service-beds/${lostBedOccupancyEntry.lostBedId}/details" class="govuk-link govuk-link--lost_bed">${expectedText}</a>
```
froddd, line 191: These tests could be updated to use the `paths.v2Manage.premises...` construct as the first test above, but this whole implementation will soon be removed, so it shouldn't matter much right now.

server/routes/v2Manage.ts
```diff
@@ -46,31 +45,27 @@ export default function routes(controllers: Controllers, router: Router, service
...
-  get(paths.v2Manage.premises.beds.overbookings.show.pattern, bedsController.overbookings(), {
-    auditEvent: 'SHOW_OVERBOOKINGS',
-    allowedRoles: ['future_manager'],
-  })
 
   // Bookings
   get(paths.v2Manage.bookings.show.pattern, bookingsController.show(), {
     auditEvent: 'SHOW_BOOKING',
-    allowedRoles: ['future_manager'],
+    allowedRoles: ['workflow_manager', 'future_manager'],
```
froddd, line 52: The V1 Manage version was accessible by the role `workflow_manager`, but is now redirected. No user has this role in prod, but we'll remove it completely in a separate PR, so for now we add it to this endpoint to satisfy the tests.

server/paths/manage.ts
```diff
@@ -100,19 +98,15 @@ const v2DateChangesPath = v2BookingPath.path('date-changes')
 const v2ExtensionsPath = v2BookingPath.path('extensions')
 const outOfServiceBedsPath = v2SinglePremisesPath.path('beds/:bedId/out-of-service-beds')
 const outOfServiceBedPath = outOfServiceBedsPath.path(':id')
-const outOfServiceBedsIndexPath = v2ManagePath.path('/out-of-service-beds')
+const outOfServiceBedsIndexPath = v2ManagePath.path('out-of-service-beds')
 
 const v2Manage = {
```
bobmeredith, line 103: As all the old manage paths are going, could this just be `manage` rather than v2Manage in the path tree.

froddd: Yes, in fact it could be squashed some more, but this may have to be done at a later point as long as we still have _some_ references to v2 in the codebase.

froddd: In fact, this should really be done -- however we do need to retain historical `v1` paths, so they can be redirected. How about we move those to `paths.manageV1`, and move any current `paths.v2Manage` to the root of `paths`?

bobmeredith (approved):
Phew!
Nice one. Lots of good reformatting and file renaming.
Looks good as far as I can see.

PR #2148, opened 2024-09-25
Only rebase Renovate branches when conflicted

froddd:
This should prevent excessive test/CI runs due to Renovate branches rebasing whenever an update to `main` is done. Instead, only conflicts with main will trigger a rebase and a test run, which shoudl be rare, given dependencies are seldom updated manually.


davidatkinsuk (approved):
LGM 👍

PR #2149, opened 2024-09-26
APS-1364: clean up unused roles

froddd:
Clean up any reference to `manager` and `legacy_manager` roles from the codebase. Any functionality linked to these roles is removed for now, as no user is currently linked to these. Any relevant functionality will then be reintroduced based on permissions, or more recent roles.

Refactors utilities around roles, so basic methods like `hasRole` and `hasPermission` are now separate from utilities to build lists of checkboxes or select options.

https://dsdmoj.atlassian.net/browse/APS-1364


server/controllers/manage/cancellationsController.test.ts
```diff
@@ -51,9 +51,7 @@ describe('cancellationsController', () => {
       await requestHandler({ ...request, params: { premisesId, bookingId } }, response, next)
 
       expect(response.render).toHaveBeenCalledWith('cancellations/new', {
-        apManager: false,
```
froddd, line 54: These are not used in the template.

integration_tests/tests/dashboard.cy.ts
```diff
@@ -8,13 +8,27 @@ context('Dashboard', () => {
   })
 
   it('displays all services when a user has all roles and permissions', () => {
-    signInWithRolesAndPermissions(['assessor', 'manager', 'future_manager'], ['cas1_view_assigned_assessments'])
+    signInWithRolesAndPermissions(
+      ['assessor', 'future_manager', 'report_viewer', 'role_admin'],
+      [
+        'cas1_view_assigned_assessments',
```
bobmeredith, line 14: I thought future_manager would have all those roles already?

bobmeredith (approved):
Looks really good.
One tiny remark

PR #2150, opened 2024-09-26
APS-1361: Prevent error if selected oasys section is null

froddd:
# Context

https://dsdmoj.atlassian.net/browse/APS-1361

# Changes in this PR

This stops the user seeing an error if the application contains some `null` values in the optional Oasys sections. When submitting the page again, those null sections are then fully cleared.


bobmeredith (approved):
LGTM

PR #2158, opened 2024-10-03
APS-1391: Only check user permissions if necessary

froddd:
There is no need to check the user's permissions if the middleware does not specify allowed permissions. This prevents passing an `undefined` list of permissions to `hasPermission`, which would result in the `TypeError` mentioned in the ticket.

Similarly, only check if the user has the role for each specified allowed role, not the other way round, for clarity.

https://dsdmoj.atlassian.net/browse/APS-1391


bobmeredith (approved):
LGTM

PR #2162, opened 2024-10-08
APS-1378 Document upload restriction

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/APS-1378

# Changes in this PR

- Limits the number of documents that can be selected for attachment to 5 in both the application and create-placement journeys.
- Adds a text area on both of the document attachment pages.
- The text area content (if populated) is shown on the corresponding 'check-your-answers' pages.

Also:
- Corrects the name of the document attachment page in the application flow.
- Fixes incorrect 'change' links on the check-your-answers page in the create-placement flow.
- change button titles to 'save and continue' on the create-placement flow rather than 'submit'


server/form-pages/apply/add-documents/attachDocuments.test.ts
```diff
@@ -124,6 +124,15 @@ describe('attachDocuments', () => {
         [`selectedDocuments_${selectedDocuments[1].id}`]: `You must enter a description for the document '${selectedDocuments[1].fileName}'`,
       })
     })
+    it('should return an error if the user has selected more than 5 documents', () => {
+      const selectedDocuments = Array(6)
+        .fill(null)
+        .map((_, idx) => documentFactory.build({ fileName: `file${idx}.pdf`, description: 'document description' }))
```
froddd, line 130: The factories have a `buildList` method to build... lists 😄  So you should be able to create the array of docs like this:

```suggestion
      const selectedDocuments = documentFactory.buildList(6)
```

This will return an array, so the rest of the test should be able to remain unchanged.

server/form-pages/apply/add-documents/attachDocuments.ts
```diff
@@ -68,23 +72,28 @@ export default class AttachDocuments implements TasklistPage {
...
   errors() {
     const errors: TaskListErrors<this> = {}
-
-    this.body.selectedDocuments.forEach(document => {
-      if (!document.description) {
-        errors[`selectedDocuments_${document.id}`] =
-          `You must enter a description for the document '${document.fileName}'`
-      }
-    })
+    // Check number of docs selected is docs is OK.
```
froddd, line 83: Code is quite self-explanatory here, this comment doesn't feel needed.

integration_tests/pages/match/placementRequestForm/additionalDocuments.ts
```diff
@@ -35,5 +37,7 @@ export default class AdditionalDocuments extends Page {
       cy.get(textareaSelector).clear()
       cy.get(textareaSelector).type(d.description)
     })
+    // Complete the additional text area
```
froddd, line 40: Comment is maybe unnecessary

server/form-pages/placement-application/request-a-placement/additionalDocuments.test.ts
```diff
@@ -111,6 +111,15 @@ describe('additionalDocuments', () => {
         [`selectedDocuments_${selectedDocuments[1].id}`]: `You must enter a description for the document '${selectedDocuments[1].fileName}'`,
       })
     })
+    it('should return an error if the user has selected more than 5 documents', () => {
+      const selectedDocuments = Array(6)
+        .fill(null)
+        .map((_, idx) => documentFactory.build({ fileName: `file${idx}.pdf`, description: 'document description' }))
```
froddd, line 117: Same as above -- can use `documentFactory.buildList(6)`

server/form-pages/apply/add-documents/attachDocuments.ts
```diff
@@ -68,23 +72,28 @@ export default class AttachDocuments implements TasklistPage {
...
+    if (this.body.selectedDocuments.length > maxDocumentsToUpload) {
+      errors[`selectedDocuments_${this.body.selectedDocuments[maxDocumentsToUpload].id}`] =
+        `You can only select ${maxDocumentsToUpload} documents, remove ${this.body.selectedDocuments.length - maxDocumentsToUpload} to continue.`
+    }
+    if (!Object.keys(errors).length) {
+      // Only check descriptions if number of docs is OK so that the user doesn't waste their time adding descriptions.
```
froddd, line 89: A test to cover this behaviour would be nice, and remove the need for this comment -- it could be covered in the test to show errors for too many docs. When using the `buildList` factory method, you can still provide default params, i.e. `documentFactory.buildList(6, { description: undefined })`

server/utils/applications/summaryListUtils.ts
```diff
@@ -55,38 +58,80 @@ const taskResponsesAsSummaryListItems = (
...
+export const getDocumentSummaryListItems = (
+  documents: Array<Document>,
+  personCrn: string,
+  changePath: string,
+  otherDocumentDetails: string,
+): Array<SummaryListItem> => {
   const items: Array<SummaryListItem> = []
 
-  documentsFromApplication(application).forEach(document => {
+  const getAction = (visuallyHiddenText = null as string) =>
```
froddd, line 76: Could the type casting be avoided here?

```suggestion
  const getAction = (visuallyHiddenText = '') =>
```

server/utils/applications/summaryListUtils.ts
```diff
@@ -55,38 +58,80 @@ const taskResponsesAsSummaryListItems = (
...
+  const changePath: string = showActions
+    ? applyPaths.applications.pages.show({ task: task.id, page: pageName, id: application.id })
+    : undefined
+  let otherDocumentDetails: string
+  try {
+    otherDocumentDetails = retrieveQuestionResponseFromFormArtifact(
+      application,
+      AttachDocuments,
+      'otherDocumentDetails',
+    ) as string
```
froddd, line 126: Not sure this type casting is useful?

server/utils/placementRequests/checkYourAnswersUtils.ts
```diff
@@ -1,8 +1,8 @@
-import { ApprovedPremisesApplication as Application, Document, PlacementApplication } from '../../@types/shared'
-import { HtmlItem, SummaryListItem, TextItem } from '../../@types/ui'
+import { ApprovedPremisesApplication as Application, Document, PlacementApplication } from '@approved-premises/api'
+import { HtmlItem, SummaryListItem, TextItem } from '@approved-premises/ui'
```
froddd, line 2: Nice 😃

server/utils/placementRequests/checkYourAnswersUtils.ts
```diff
@@ -123,34 +123,30 @@ export const attachDocumentsSummaryListItems = (
   pageName: string,
   showActions: boolean,
 ) => {
-  const items: Array<SummaryListItem> = []
+  const personCrn: string = application.person.crn
```
froddd, line 126: This now seems very similar to `attachDocumentsSummaryListItems` in the summaryListUtils above -- is this actual duplication?

server/views/placement-applications/pages/request-a-placement/additional-documents.njk
```diff
@@ -5,11 +5,19 @@
 
 {% set columnClasses = "govuk-grid-column-full applications--pages--attach-documents" %}
 
+{% set textboxLabel %}
+  <p class="govuk-body">If there are any other relevant documents on Ndelius that you would like the assessor to know about, add details here.</p>
```
froddd, line 9: NDelius seems to usually be spelled with both capital N and D across the codebase -- worth checking to make it consistent.

froddd (commented):
Quite the PR in the end -- good work! A few comments on factories, and a couple of other things I may not be understanding correctly...

server/form-pages/apply/add-documents/attachDocuments.ts
bobmeredith: True

server/form-pages/apply/add-documents/attachDocuments.test.ts
bobmeredith: Nice!

server/utils/applications/summaryListUtils.ts
bobmeredith: Fair enough. The only downside is that it would rely on checking for an empty string rather than null

server/utils/applications/summaryListUtils.ts
bobmeredith: True. At one stage I was declaring inside the try/catch.
Heh - I just noticed there is a routine `retrieveOptionalQuestionResponseFromFormArtifact` that absorbs the exception if the value is not populated. That should simplify things a bit.

server/utils/placementRequests/checkYourAnswersUtils.ts
bobmeredith: The entire files were clearly copied and pasted at some stage so yes, there is lots of duplication. I merged the core of the summary list generation to `getDocumentSummaryListItems` inside summaryListUtils so that both `attachDocumentsSummaryListItems` here and `attachDocumentsSummaryListItems` in the application journey can share it. 
Maybe there is scope to move more code into the shared routines - but document lists are coming from different entity types so pushing it much further will probably hit type issues.
Actually, using the fn that absorbs exceptions, these two fns become very small.

server/form-pages/apply/add-documents/attachDocuments.ts
bobmeredith: Fair enough.

froddd (approved):
Looks good, nice!

PR #2163, opened 2024-10-08
APS-1334: task allocation Women's Estate

froddd:
# Context

https://dsdmoj.atlassian.net/browse/APS-1334

# Changes in this PR

Fetches the list of CRU management areas instead of AP areas for the 'AP area' filter on the Task allocation page. This lets the user filter the list for applications to the Women's Estate only.

The new CAS1-specific Reference Data endpoint has been added as a service, and CRU information has been added to the user's session data to enable pre-selecting the filter.

The label for the filter has been changed to singular form for consistency with other filters.

#### Showing _Women's Estate_ option in AP area filter

Screenshot 2024-10-09 at 09 26 45

### DIFF
--- a/server/data/userClient.ts
+++ b/server/data/userClient.ts
@@ -30,9 +30,10 @@ export default class UserClient {
   }
 
   async getUserList(roles: Array<UserRole> = []): Promise<Array<UserSummary>> {
-    return this.restClient.get({ path: paths.users.summary({}), query: { roles: roles.join(',') } }) as Promise<
-      Array<User>
-    >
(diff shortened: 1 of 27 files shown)


server/data/userClient.ts
bobmeredith, line 1: This looks the same.  I guess it's just a prettify.

bobmeredith (approved):
Looks good.

server/data/userClient.ts
froddd: Added `await` to quash IDE suggestion, but mostly a prettify, yes 🙂

PR #2164, opened 2024-10-09
APS-1339: Use CRU Management data to populate CRU dashboard area filter

froddd:
# Context

https://dsdmoj.atlassian.net/browse/APS-1339

# Changes in this PR

On the CRU dashboard, populate the AP area filter using CRU management areas instead of simply AP areas, so the applications can be filtered for Women's Estate.

#### Showing _Women's Estate_ option in AP area filter

Screenshot 2024-10-09 at 16 47 15


bobmeredith (approved):
LGTM

PR #2167, opened 2024-10-14
APS-1402 Women's estate - Preferred AP selection

bobmeredith:
# Context
https://dsdmoj.atlassian.net/browse/APS-1402

As discussed in the ticket, this PR does not implement the checkbox page indicated in the design, but rather applies only a slight modification to the `Preferred AP` page for the women's estate.

- The title is modified as per the design
- The two-level selects are not applied. i.e. the user will see all the Women's estate APS in each select rather than being pre-filtered by area. This is because there are relatively few APs in the women's estate.

The ticket also implements two features needed to drive the above:

- The field **isWomensApplication** in the application is populated using the sex of the applicant and the selection of the Men's AP or Women's AP from the exceptional case board findings.
- The field **isInApplicable** in the application is no longer set if the exceptional case board find for a woman's AP, nor if the person's sex is female.
- Following from the exceptional case board findings page, the application will no longer abort with the 'refer to delius' page if the WE feature flag is set.

# Changes in this PR

### After

image

image

### DIFF
--- a/integration_tests/pages/apply/preferredAps.ts
+++ b/integration_tests/pages/apply/preferredAps.ts
@@ -4,9 +4,14 @@ import paths from '../../../server/paths/apply'
 import ApplyPage from './applyPage'
 
 export default class PreferredAps extends ApplyPage {
+  isWomensApplication: boolean
+
   constructor(application: ApprovedPremisesApplication) {
+    this.isWomensApplication = application.isWomensApplication
     super(
-      'Select a preferred AP',
+      this.isWomensApplication
+        ? 'Select all preferred properties for your women’s AP application'
+        : 'Select a preferred AP',
       application,
       'location-factors',
       'preferred-aps',
@@ -19,11 +24,17 @@ export default class PreferredAps extends ApplyPage {
   }
 
   completeForm() {
-    this.getSelectInputByIdAndSelectAnEntry('area0', 'area1')
-    this.selectSelectOptionFromPageBody('preferredAp1')
-    this.getSelectInputByIdAndSelectAnEntry('area1', 'area2')
-    this.selectSelectOptionFromPageBody('preferredAp2')
-    this.getSelectInputByIdAndSelectAnEntry('area2', 'area3')
-    this.selectSelectOptionFromPageBody('preferredAp3')
+    if (this.isWomensApplication) {
+      this.selectSelectOptionFromPageBody('preferredAp1')
+      this.selectSelectOptionFromPageBody('preferredAp2')
+      this.selectSelectOptionFromPageBody('preferredAp3')
+    } else {
+      this.getSelectInputByIdAndSelectAnEntry('area0', 'area1')
+      this.selectSelectOptionFromPageBody('preferredAp1')
+      this.getSelectInputByIdAndSelectAnEntry('area1', 'area2')
+      this.selectSelectOptionFromPageBody('preferredAp2')
+      this.getSelectInputByIdAndSelectAnEntry('area2', 'area3')
+      this.selectSelectOptionFromPageBody('preferredAp3')
+    }
   }
 }
--- a/integration_tests/pages/apply/preferredAps.ts
+++ b/integration_tests/pages/apply/preferredAps.ts
@@ -4,9 +4,7 @@ import paths from '../../../server/paths/apply'
 import ApplyPage from './applyPage'
 
 export default class PreferredAps extends ApplyPage {
-  private readonly isWomensApplication: boolean
-
-  constructor(application: ApprovedPremisesApplication) {
+  constructor(private readonly application: ApprovedPremisesApplication) {
     super(
       application.isWomensApplication
         ? 'Select all preferred properties for your women’s AP application'
@@ -20,11 +18,10 @@ export default class PreferredAps extends ApplyPage {
         page: 'describe-location-factors',
       }),
     )
-    this.isWomensApplication = application.isWomensApplication
   }
 
   completeForm() {
-    if (this.isWomensApplication) {
+    if (this.application.isWomensApplication) {
       this.selectSelectOptionFromPageBody('preferredAp1')
       this.selectSelectOptionFromPageBody('preferredAp2')
       this.selectSelectOptionFromPageBody('preferredAp3')
--- a/integration_tests/pages/apply/preferredAps.ts
+++ b/integration_tests/pages/apply/preferredAps.ts
@@ -7,9 +7,8 @@ export default class PreferredAps extends ApplyPage {
   isWomensApplication: boolean
 
   constructor(application: ApprovedPremisesApplication) {
-    this.isWomensApplication = application.isWomensApplication
     super(
-      this.isWomensApplication
+      application.isWomensApplication
         ? 'Select all preferred properties for your women’s AP application'
         : 'Select a preferred AP',
       application,
@@ -21,6 +20,7 @@ export default class PreferredAps extends ApplyPage {
         page: 'describe-location-factors',
       }),
     )
+    this.isWomensApplication = application.isWomensApplication
   }
 
   completeForm() {
--- a/integration_tests/pages/apply/preferredAps.ts
+++ b/integration_tests/pages/apply/preferredAps.ts
@@ -4,7 +4,7 @@ import paths from '../../../server/paths/apply'
 import ApplyPage from './applyPage'
 
 export default class PreferredAps extends ApplyPage {
-  isWomensApplication: boolean
+  private readonly isWomensApplication: boolean
 
   constructor(application: ApprovedPremisesApplication) {
     super(
--- a/integration_tests/support/commands.ts
+++ b/integration_tests/support/commands.ts
@@ -1,6 +1,4 @@
 Cypress.Commands.add('signIn', (options = { failOnStatusCode: true }) => {
-  cy.task('stubAllPremises', [])
-  cy.task('stubCas1AllPremises', [])
   cy.request('/')
   return cy.task('getSignInUrl').then((url: string) => cy.visit(url, options))
 })
--- a/integration_tests/support/commands.ts
+++ b/integration_tests/support/commands.ts
@@ -1,5 +1,6 @@
 Cypress.Commands.add('signIn', (options = { failOnStatusCode: true }) => {
   cy.task('stubAllPremises', [])
+  cy.task('stubCas1AllPremises', [])
   cy.request('/')
   return cy.task('getSignInUrl').then((url: string) => cy.visit(url, options))
 })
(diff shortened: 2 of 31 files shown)


bobmeredith:
@froddd This really needs some integration tests over the women's estate but I can't see how to do that with the feature flag being an environment variable.
I've switched the old premises summary endpoint for the new one in all but the one case (in manage) where the bedcount and filter by area is needed.

froddd:
@bobmeredith given this will become the new default, should the integration tests all run with the Women's Estate enabled from now?

server/services/premisesService.ts
```diff
@@ -18,15 +19,14 @@ export default class PremisesService {
...
-      return 0
-    })
+    return premises.sort((a, b) => (a.name < b.name ? -1 : 1))
+  }
+
+  async getCas1All(token: string, gender: ManWoman = null): Promise<Array<Cas1PremisesSummary>> {
+    const premisesClient = this.premisesClientFactory(token)
+    const premises = await premisesClient.allCas1(gender)
+
+    return premises.sort((a, b) => (a.name < b.name ? -1 : 1))
```
froddd, line 29: `localeCompare` might be more robust for sorting strings:

```suggestion
    return premises.sort((a, b) => a.name.localeCompare(b.name))
```

server/services/premisesService.ts
```diff
@@ -18,15 +19,14 @@ export default class PremisesService {
...
-    return premises.sort((a, b) => {
-      if (a.name < b.name) {
-        return -1
-      }
-      if (a.name > b.name) {
-        return 1
-      }
-      return 0
-    })
+    return premises.sort((a, b) => (a.name < b.name ? -1 : 1))
```
froddd, line 22: Same as below comment -- would favour using `localeCompare`

server/form-pages/apply/risk-and-need-factors/location-factors/preferredAps.ts
```diff
@@ -12,43 +12,50 @@ export type PreferredApsBody = {
...
+  title: string
 
-  allPremises: Array<ApprovedPremises> = []
+  allPremises: Array<Cas1PremisesSummary> = []
 
   preferredApOptions = preferredAps
 
   preferredApLabels = this.preferredApOptions.map((_, i) => `${numberToOrdinal(i)} choice AP`)
 
+  isWomensAp: boolean
```
froddd, line 37: Is this used anywhere? I can't find any direct usage, and it seems `application.isWomensApplication` should cover all bases? 

In any case, the abbreviation `ap` tends to refer to Approved Premises around this codebase, so to avoid confusion I'd recommend using the full `isWomensApplication` if this is still needed.

server/utils/applications/getApplicationData.test.ts
```diff
@@ -168,6 +170,7 @@ describe('getApplicationData', () => {
     it('returns all the defined attributes', () => {
       ;(arrivalDateFromApplication as jest.Mock).mockReturnValue('2023-01-01')
       ;(isInapplicable as jest.Mock).mockReturnValue(false)
+      ;(isWomensAp as jest.Mock).mockReturnValue(false)
```
froddd, line 173: Probably worth testing `isWomensApplication: true` in at least one test case?

server/utils/applications/utils.test.ts
```diff
@@ -594,22 +596,36 @@ describe('utils', () => {
...
 
-      expect(isInapplicable(application)).toEqual(true)
+      expect(isInapplicable(application)).toEqual(false)
     })
+  })
 
-    it('should return false if the applicant has answered yes to the shouldPersonBePlacedInMaleAp question', () => {
-      mockOptionalQuestionResponse({ shouldPersonBePlacedInMaleAp: 'yes' })
+  describe('isWomensAp', () => {
+    it(`should return true if the person from nDelius has a sex of 'Male'`, () => {
```
froddd, line 607: Test title seems the wrong way round here!

Also, given we rely on the answer to `shouldPersonBePlacedInMaleAp` to determine the output on this, it would be nice to be explicit and include it in the test setup.

server/utils/applications/utils.test.ts
```diff
@@ -594,22 +596,36 @@ describe('utils', () => {
...
+      const application = applicationFactory.build()
+      const fp = application.person as FullPerson
+      fp.sex = 'Male'
 
-      expect(isInapplicable(application)).toEqual(false)
+      expect(isWomensAp(application)).toEqual(false)
     })
+    it(`should return false if the person from nDelius has a sex of 'Female'`, () => {
```
froddd, line 614: Test title wrong way round

server/utils/applications/utils.ts
```diff
@@ -199,13 +194,19 @@ const isInapplicable = (application: Application): boolean => {
...
   }
 
-  if (shouldPersonBePlacedInMaleAp === 'no') {
-    return true
-  }
-
   return false
 }
 
+const isWomensAp = (application: Application): boolean => {
```
froddd, line 200: As per previous comment, `ap` could be ambiguous here -- let's be explicit and call this `isWomensApplication`

server/utils/premises/index.test.ts
```diff
@@ -240,6 +241,50 @@ describe('premisesUtils', () => {
...
+          items: [
+            { selected: false, text: area2Premises[0].name, value: area2Premises[0].id },
+            { selected: false, text: area2Premises[1].name, value: area2Premises[1].id },
+          ],
+          label: 'Area 2',
+        },
+      ])
+    })
+
+    it('should support a field name', () => {
```
froddd, line 268: What does the field name do? Can this be made a bit more explicit, i.e. 'marks the correct option as selected'

server/utils/premises/index.test.ts
```diff
@@ -240,6 +241,50 @@ describe('premisesUtils', () => {
     })
   })
 
+  describe('groupCas1SummaryPremisesSelectOptions', () => {
+    const area1Premises = cas1PremisesSummaryFactory.buildList(2, { apArea: { id: 'a1', name: 'Area 1' } })
+    const area2Premises = cas1PremisesSummaryFactory.buildList(2, { apArea: { id: 'a2', name: 'Area 2' } })
+    const premises = [...area1Premises, ...area2Premises]
+
+    it('should group premises by region', () => {
+      expect(groupCas1SummaryPremisesSelectOptions(premises, { premisesId: area1Premises[1].id })).toEqual([
```
froddd, line 250: Probably no need to test the selected option here -- test case seems to be basic render.

integration_tests/pages/apply/preferredAps.ts
```diff
@@ -16,14 +20,21 @@ export default class PreferredAps extends ApplyPage {
         page: 'describe-location-factors',
       }),
     )
+    this.isWomensApplication = application.isWomensApplication
```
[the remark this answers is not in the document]
bobmeredith: true

integration_tests/support/commands.ts
```diff
@@ -1,5 +1,6 @@
 Cypress.Commands.add('signIn', (options = { failOnStatusCode: true }) => {
   cy.task('stubAllPremises', [])
+  cy.task('stubCas1AllPremises', [])
```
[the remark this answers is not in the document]
bobmeredith: Not sure. I'll try clearing them up and see what I get

server/services/premisesService.ts
bobmeredith: True, would be better if the strings may contain numerics.

server/services/premisesService.ts
bobmeredith: true

server/utils/premises/index.test.ts
bobmeredith: It was a copy'n'paste from the original so yes, it's not very precise.

server/utils/premises/index.test.ts
bobmeredith: It was a copy'n'paste from existing code but I think it's valid. First test case shows selection of the default selection fieldname while second shows an explicit one.

froddd: The context could be empty, so I'd expect to see a test for:
```suggestion
      expect(groupCas1SummaryPremisesSelectOptions(premises, {})).toEqual([
```

server/data/premisesClient.ts
```diff
@@ -28,6 +29,14 @@ export default class PremisesClient {
     })) as Array<PremisesSummary>
   }
 
+  async allCas1(gender: ManWoman = null): Promise<Array<Cas1PremisesSummary>> {
+    const query = gender ? { gender } : {}
+    return (await this.restClient.get({
+      path: paths.premises.indexCas1({}),
+      query,
+    })) as Array<Cas1PremisesSummary>
+  }
```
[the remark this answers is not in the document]
bobmeredith: Well a params argument would maybe better but it's not a pattern that seems to be used in the services. It does make type checking simpler though.

integration_tests/tests/manage/departure.cy.ts
```diff
@@ -33,6 +33,7 @@ context('Departures', () => {
     })
 
     cy.task('stubAllPremises', premises)
+    cy.task('stubCas1AllPremises', premises)
```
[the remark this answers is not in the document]
bobmeredith: yes indeed - good spot

server/utils/applications/utils.test.ts
bobmeredith: It is.

integration_tests/pages/apply/preferredAps.ts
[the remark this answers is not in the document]
froddd: Of course -- it would need marking as `private` in the constructor signature, which would expose as a member of the class

server/utils/premises/index.test.ts
bobmeredith: That's covered off in the first test - which is testing for the grouping. I thought that specifically testing for the selected flags, it wouldn't need to check the entire list content.

bobmeredith:
> @bobmeredith given this will become the new default, should the integration tests all run with the Women's Estate enabled from now?

This I will look into although I'm not quite sure how to make it work. If I try setting the flag in the integration test environment, I can't get the tests to pass because they compare the rendered page with page objects created within the tests - which return old values as if the env flag is not set (because they're running off different environments). IMHO, it's pretty crazy linking tests and code like this as it prevents legitimate test failures but reworking would be a big task.
I'll have another rummage.

integration_tests/pages/apply/preferredAps.ts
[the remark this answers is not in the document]
bobmeredith: True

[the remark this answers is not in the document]
froddd: Sorry, I meant making the `application` argument `private readonly` in the constructor:

```js
constructor(private readonly application: ApprovedPremisesApplication) {
  // ...
}
```

Which would make `application` available as a member of the class:

```js
if (this.application.isWomensApplication) {
  // ...
}
```

froddd (approved):
Looks good!

PR #2169, opened 2024-10-17
Separate shellcheck from other linting

froddd:
Ensures shellcheck can also run if other linting fails.


package-lock.json
aliuk2012, line 1: Wasn't expecting to see these changes.

aliuk2012 (approved):

PR #2172, opened 2024-10-21
APS-1341 WE - Record a placement

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/APS-1341

# Changes in this PR
In the create placement flow, if the `isWomensApplication` flag is set on the associated application, the AP/date selection page shows only women's estate APs and displays them as a set of radio buttons, rather than a nested drop-down. If the application is not a women's application, the page is unchanged but only men's APs are shown in the nested select.  

### Before

image

### After

image

### DIFF
--- a/integration_tests/pages/admin/placementApplications/createPlacementPage.ts
+++ b/integration_tests/pages/admin/placementApplications/createPlacementPage.ts
@@ -1,10 +1,18 @@
-import { Cas1PremisesSummary, PlacementRequestDetail } from '../../../../server/@types/shared'
+import {
+  ApprovedPremisesApplication,
+  Cas1PremisesSummary,
+  PlacementRequestDetail,
+} from '../../../../server/@types/shared'
 import { placementDates } from '../../../../server/utils/match'
 import Page from '../../page'
 
 export default class CreatePlacementPage extends Page {
   constructor(private readonly placementRequest: PlacementRequestDetail) {
-    super('Record an Approved Premises (AP) placement')
+    super(
+      (placementRequest.application as ApprovedPremisesApplication).isWomensApplication
+        ? `Record a women’s Approved Premises placement`
+        : 'Record an Approved Premises (AP) placement',
+    )
   }
 
   dateInputsShouldBePrepopulated(): void {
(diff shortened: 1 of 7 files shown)


integration_tests/pages/admin/placementApplications/createPlacementPage.ts
```diff
@@ -1,10 +1,18 @@
-import { Cas1PremisesSummary, PlacementRequestDetail } from '../../../../server/@types/shared'
+import {
+  ApprovedPremisesApplication,
+  Cas1PremisesSummary,
+  PlacementRequestDetail,
+} from '../../../../server/@types/shared'
```
froddd, line 5: Import can be shortened:

```suggestion
} from '@approved-premises/api'
```

integration_tests/tests/admin/placementRequests.cy.ts
```diff
@@ -83,9 +78,43 @@ context('Placement Requests', () => {
...
   it('allows me to view a placement request', () => {
+    const {
+      unmatchedPlacementRequest,
+      unmatchedPlacementRequests,
+      unableToMatchPlacementRequests,
+      matchedPlacementRequests,
+      preferredAps,
+      matchedPlacementRequest,
+      parolePlacementRequest,
+    } = stubArtifacts()
```
froddd, line 117: Very nice 😍

froddd (commented):
All good with what's there, but the 'Create placement' page doesn't handle errors very nicely: error summary doesn't link to the error field, and I guess the error message should be different for a women's placement, as it currently references the 'AP area':

Screenshot 2024-10-21 at 12 29 03

Could we improve the errors on this?

integration_tests/pages/admin/placementApplications/createPlacementPage.ts
bobmeredith: It can indeed. We should probably replace all these relative paths with simpler path aliases at some stage.

froddd: Yeah, I've only been doing them as and when I find them, I think the IDE does this for me on save, in fact.

froddd (approved):
As discussed, approved to go as is!

PR #2175, opened 2024-10-21
APS-1435: Add women's AP qualification

froddd:
# Context

# Changes in this PR

This adds the qualification to the task reallocation filters, as well as the user management, so it can be assigned to users.

## Todo

- [ ] Get types updated from API

#### Task allocation filters:

Screenshot 2024-10-21 at 16 28 01
 
---

Screenshot 2024-10-21 at 16 25 40

---

#### User management edit screen: 

Screenshot 2024-10-21 at 16 25 58

### DIFF
--- a/server/@types/shared/models/UserQualification.ts
+++ b/server/@types/shared/models/UserQualification.ts
@@ -2,4 +2,12 @@
 /* istanbul ignore file */
 /* tslint:disable */
 /* eslint-disable */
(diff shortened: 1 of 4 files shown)


froddd:
Not implementing this: instead we'll continue using the CRU Management Area for filtering all entities in relation to the Women's Estate.

PR #2178, opened 2024-10-22
APS-1072 add user manager role checkbox

froddd:
# Context

https://dsdmoj.atlassian.net/browse/APS-1072

# Changes in this PR

Adds a checkbox for the 'User manager' role to the User roles edit screen.

## Screenshots of UI changes

Screenshot 2024-10-22 at 15 26 03

---

Screenshot 2024-10-22 at 15 26 33

### DIFF
--- a/server/views/admin/users/edit.njk
+++ b/server/views/admin/users/edit.njk
@@ -4,83 +4,88 @@
 {% from "govuk/components/checkboxes/macro.njk" import govukCheckboxes %}
 {% from "govuk/components/button/macro.njk" import govukButton %}
 
-{% set pageTitle = applicationName + " - " + pageHeading  %}
+{% set pageTitle = applicationName + " - " + pageHeading %}
 {% set mainClasses = "app-container govuk-body assessments--index" %}
+{% set allocationPreferencesHint %}
+    <p class="govuk-body govuk-hint">Select all that apply</p>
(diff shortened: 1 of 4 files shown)


server/views/admin/users/edit.njk
```diff
@@ -4,83 +4,90 @@
...
+                            classes: "govuk-fieldset__legend--m"
+                        }
+                    },
+                    hint: {
+                        text: "Select all that apply"
+                    },
+                    items: UserUtils.rolesToCheckboxItems(roles, user.roles)
+                }) }}
+
+                {% set allocationPreferencesHint %}
```
bobmeredith, line 45: Nice to move this to a variable, but how about moving the declarations to the top of the file so they don't interrupt the flow.

server/views/admin/users/edit.njk
```diff
@@ -89,6 +96,7 @@
                         preventDoubleClick: true
                     }) }}
                 </div>
-            </div>
+            </form>
```
bobmeredith, line 99: Whoa - so there was no form close!

bobmeredith (approved):
Looks good. Just one minor suggestion

server/views/admin/users/edit.njk
froddd: I know, right 😨

bobmeredith (approved):
Nice!

PR #2190, opened 2024-10-31
APS-1455: remove unused bookings v1 functionality

froddd:
# Context

https://dsdmoj.atlassian.net/browse/APS-1455

# Changes in this PR

Removes all unused implementations from Match & Manage V1:

- Removes the following booking implementations and related controllers:
  - Arrivals
  - Non-arrivals
  - Departures
  - Moves
- Removes implementation and routes for calendar
- Removes implementation and related routes for overbooking
- Removes any unused Premises utilities
- Removes unused types


bobmeredith (approved):
Looks good - that's a helluva lot of code to delete.
Slightly confused about `server/views/arrivals/_non-arrival_form.njk` as it's showing as an empty file.  Is that being deleted?

froddd:
Yes, that was an empty file -- it was leftover when the template was moved to a new, separate location: https://github.com/ministryofjustice/hmpps-approved-premises-ui/commit/2f0e6bb7b21af13ec4d3f7cead10626864274867

PR #2194, opened 2024-11-05
Feature/aps 1267 view placement details

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/APS-1267

# Changes in this PR

This PR adds a page to show the details for a single placement.
The page contains a header containing information about the person.
Things to note:

- Some data are still missing from the API
-- Status - this could be calculated in the UI perhaps.
-- Non arrival reason
-- Non arrival any other information
-- Breach or recall (isn’t this the departure reason?)
-- More information
- Greyed-out information is not as light as in the design due to text contrast/accessibility issues. These data are to be hidden anyway I believe.
- The back link uses logic to route either to the premises page or the full referrer if it matches the premises details page of the correct premises. This is to be reviewed and perhaps needs specific labelling. 
- The actions menu is present and depends on user permissions as well as the context of the placement i.e. if the `actualArrivalDate` is set, the option to record arrival is removed. The links are however blank at this point.

## Screenshots of UI changes

image


server/testutils/factories/cas1SpaceBooking.ts
```diff
@@ -0,0 +1,48 @@
...
+import userFactory from './user'
+
+import staffMemberFactory from './staffMember'
+import { DateFormats } from '../../utils/dateUtils'
+
+const addTime = (date: Date): Date => new Date(date.getTime() + faker.number.int({ max: 60 * 60 * 24 * 1000 }))
+
+export default Factory.define<Cas1SpaceBooking>(() => {
+  const startDateTime = addTime(faker.date.soon({ days: 90 }))
+  const endDateTime = addTime(faker.date.soon({ days: 365, refDate: startDateTime }))
```
froddd, line 16: What's the purpose of using `addTime` here? I'm not sure I get it... If this is to push the dates into the future a bit, there is an `addDays` utility somewhere, I believe!

server/testutils/factories/cas1SpaceBookingDates.ts
```diff
@@ -0,0 +1,18 @@
...
+import { faker } from '@faker-js/faker'
+import { Cas1SpaceBookingDates } from '@approved-premises/api'
+import { DateFormats } from '../../utils/dateUtils'
+
+export default Factory.define<Cas1SpaceBookingDates>(() => {
+  const startDate = faker.date.anytime()
+  const [canonicalArrivalDate, canonicalDepartureDate] = [
+    startDate,
+    faker.date.soon({ days: 365, refDate: startDate }),
+  ].map(DateFormats.dateObjToIsoDate)
```
froddd, line 11: Not convinced this reads better than:

```js
const canonicalArrivalDate = DateFormats.dateObjToIsoDate(startDate)
const canonicalDepartureDate = DateFormats.dateObjToIsoDate(faker.date.soon({ days: 365, refDate: startDate }))
```

server/utils/placements/index.test.ts
```diff
@@ -0,0 +1,208 @@
...
+        actualArrivalDate: undefined,
+        actualDepartureDate: undefined,
+        keyWorkerAllocation: undefined,
+      })
+      expect(actions(placement, userDetails)).toEqual([
+        { classes: 'govuk-button--secondary', href: '', text: 'Allocate keyworker' },
+        { classes: 'govuk-button--secondary', href: '', text: 'Record arrival' },
+      ])
+    })
+    it('should include with record arrival option before arrival', () => {
```
froddd, line 37: 'should render'?

server/utils/placements/index.test.ts
```diff
@@ -0,0 +1,208 @@
...
+      expect(getBackLink(urlWithQuery, premises)).toEqual(urlWithQuery)
+      expect(getBackLink('some string', premises)).toEqual(bareUrl)
+      expect(getBackLink('', premises)).toEqual(bareUrl)
+      expect(getBackLink(null, premises)).toEqual(bareUrl)
+      expect(getBackLink(urlOtherId, premises)).toEqual(bareUrl)
+    })
+  })
+
+  describe('getKeyDetail', () => {
+    it('should allow setting of keyworker when placement in initial state', () => {
```
froddd, line 91: This test name seems incorrect?

server/utils/placements/index.test.ts
```diff
@@ -0,0 +1,208 @@
...
+        { id: 'id2', canonicalArrivalDate: '2024-09-20', canonicalDepartureDate: '2025-03-20' },
+      ]
+
+      const placement = cas1SpaceBookingFactory.build({ otherBookingsInPremisesForCrn: placementList })
+      expect(otherBookings(placement)).toEqual({
+        rows: [
+          {
+            key: { text: 'Other placement bookings at this premises' },
+            value: {
+              html: '<a href="id1">Placement 10 Sep 2024 to 04 Jun 2025</a><br/><a href="id2">Placement 20 Sep 2024 to 20 Mar 2025</a>',
```
froddd, line 201: Should this `href` be the actual path?

Also it would be good to use an actual list for accessibility here.

server/controllers/manage/placementController.ts
```diff
@@ -0,0 +1,21 @@
...
+export default class PlacementController {
+  constructor(private readonly premisesService: PremisesService) {}
+
+  show(): RequestHandler {
+    return async (req: Request, res: Response) => {
+      const { premisesId, placementId } = req.params
+      const referrer = req.headers.referer
+      const { user } = res.locals
+      const placement = await this.premisesService.getPlacement({ token: req.user.token, premisesId, placementId })
+      const premises = await this.premisesService.find(req.user.token, req.params.premisesId)
```
froddd, line 15: It seems all the premises info the template ever uses is either the premises `id` or the `name`. I believe these are available directly from `placement.premises`, so we can save an API request here.

integration_tests/mockApis/spaceBooking.ts
```diff
@@ -40,7 +55,7 @@ export default {
     return stubFor({
       request: {
         method: 'GET',
-        urlPattern: `${paths.premises.placements({ premisesId: args.premisesId })}.*`,
+        urlPattern: `${paths.premises.placements.index({ premisesId: args.premisesId })}.*`,
```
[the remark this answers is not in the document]
bobmeredith: It's a GET so it needs the `.*` to soak up the query params.

server/paths/manage.ts
```diff
@@ -72,7 +72,7 @@ const paths = {
     },
     placements: {
       index: singlePremisesPath.path('space-bookings'),
-      show: singlePremisesPath.path('space-bookings').path(':bookingId'),
+      show: singlePremisesPath.path('placements').path(':placementId'),
```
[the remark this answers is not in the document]
bobmeredith: No, it seems not. I'll remove it

server/testutils/factories/cas1SpaceBooking.ts
bobmeredith: The `actual` dates are actually date/times. This adds a random time of up to 24 hours to the dates.

server/testutils/factories/cas1SpaceBookingDates.ts
bobmeredith: Fair enough.

server/utils/placements/index.test.ts
bobmeredith: Very true

integration_tests/pages/manage/placementShow.ts
```diff
@@ -0,0 +1,36 @@
...
+    ;[
+      ['Delius Event Number', placement.deliusEventNumber],
+      ['AP name', premises.name],
+      ['Key worker', placement.keyWorkerAllocation.keyWorker.name],
+      ['Expected arrival date', DateFormats.isoDateToUIDate(placement.expectedArrivalDate)],
+      ['Expected departure date', DateFormats.isoDateToUIDate(placement.expectedDepartureDate)],
+    ].forEach(([title, expectedText]) => {
+      cy.contains(title).closest('.govuk-summary-list__row').should('contain', expectedText)
+    })
+  }
```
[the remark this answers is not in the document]
bobmeredith: Now there's interesting! It feels wrong to use the same code that drives the template in an integration test, but I guess that's tested in the unit tests so it's OK and is a lot easier.

server/controllers/manage/placementController.ts
bobmeredith: True. It does seem architecturally wrong for the api calls to be so tailored for the precise requirements of the UI but I guess as the information is there... 
It tidies up quite a lot of parameter passing in the tests too.

server/utils/placements/index.test.ts
bobmeredith: Actually, the id alone is a valid relative path, given that we are already on a placement detail page. It would probably be better to include a full path though.

integration_tests/pages/manage/placementShow.ts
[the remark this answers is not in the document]
bobmeredith: Actually, it turned out to be a bit messy because that method doesn't handle html in both the key and value. I've split out the tests to cope and had to add a bodgy filter - but that can go when we remove the grey text.

server/utils/placements/index.test.ts
froddd: Oh, I see -- yeah, full path would be better for discoverability if anything (finding all places a certain path is used in code).

froddd (approved):
Looks good!

PR #2195, opened 2024-11-06
APS-1411: record placement arrival

froddd:
# Context

https://dsdmoj.atlassian.net/browse/APS-1411

# Changes in this PR

Implements the 'Record arrival' flow from the placement screen. For now this uses the standard, GOVUK-style date input field -- the datepicker will be implemented in a separate ticket. The confirmation banner used is also consistent with the rest of the service and uses the GOVUK version, rather than the MoJ version shown in the design -- this has been agreed with the designer and will be reviewed service-wide at a later stage.

The 'Time of arrival' field is validated, and expects a valid 24hr-format time, with or without leading zero for the hours.

### Arrival form

Screenshot 2024-11-07 at 10 17 17

---

### Arrival form with 'empty' errors

Screenshot 2024-11-07 at 10 17 33

---

### Arrival form with 'invalid' errors

Screenshot 2024-11-07 at 10 18 14

---

### Confirmation on placement screen

Screenshot 2024-11-07 at 10 18 59

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,9 +40,9 @@
     "clean": "rm -rf dist build node_modules stylesheets",
     "start-test-wiremock": "docker compose -f docker-compose-test.yml up -d",
     "generate-types": "script/generate-types",
-    "create-application:standard": "APPLICATION_TYPE='standard' npx playwright test --config ./e2e/playwright.config.ts --project=dev utils/createApplication.spec.ts",
-    "create-application:emergency": "APPLICATION_TYPE='emergency' npx playwright test --config ./e2e/playwright.config.ts --project=dev utils/createApplication.spec.ts",
-    "create-application:short-notice": "APPLICATION_TYPE='shortNotice' npx playwright test --config ./e2e/playwright.config.ts --project=dev utils/createApplication.spec.ts"
+    "create-application:standard": "APPLICATION_TYPE='standard' npx playwright test --config ./e2e/playwright.config.ts --project=local utils/createApplication.spec.ts",
+    "create-application:emergency": "APPLICATION_TYPE='emergency' npx playwright test --config ./e2e/playwright.config.ts --project=local utils/createApplication.spec.ts",
+    "create-application:short-notice": "APPLICATION_TYPE='shortNotice' npx playwright test --config ./e2e/playwright.config.ts --project=local utils/createApplication.spec.ts"
   },
   "engines": {
     "node": "^20.0.0",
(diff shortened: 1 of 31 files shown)


server/utils/dateUtils.ts
```diff
@@ -339,3 +339,12 @@ export const daysToWeeksAndDays = (days: string | number): { days: number; weeks
     weeks: durationWeeks,
   }
 }
+
```
bobmeredith, line 342: You could do this with regex - `^(2[0-3]|[01]?[0-9]):([0-5]?[0-9])$` and then it wouldn't need a try/catch

server/testutils/factories/cas1SpaceBooking.ts
```diff
@@ -11,7 +11,18 @@ import { DateFormats } from '../../utils/dateUtils'
 
 const addTime = (date: Date): Date => new Date(date.getTime() + faker.number.int({ max: 60 * 60 * 24 * 1000 }))
 
-export default Factory.define<Cas1SpaceBooking>(() => {
+class Cas1SpaceBookingFactory extends Factory<Cas1SpaceBooking> {
```
bobmeredith, line 14: Nice!

package.json
```diff
@@ -40,9 +40,9 @@
     "clean": "rm -rf dist build node_modules stylesheets",
     "start-test-wiremock": "docker compose -f docker-compose-test.yml up -d",
     "generate-types": "script/generate-types",
-    "create-application:standard": "APPLICATION_TYPE='standard' npx playwright test --config ./e2e/playwright.config.ts --project=dev utils/createApplication.spec.ts",
-    "create-application:emergency": "APPLICATION_TYPE='emergency' npx playwright test --config ./e2e/playwright.config.ts --project=dev utils/createApplication.spec.ts",
-    "create-application:short-notice": "APPLICATION_TYPE='shortNotice' npx playwright test --config ./e2e/playwright.config.ts --project=dev utils/createApplication.spec.ts"
+    "create-application:standard": "APPLICATION_TYPE='standard' npx playwright test --config ./e2e/playwright.config.ts --project=local utils/createApplication.spec.ts",
```
bobmeredith, line 43: Has this crept in to the PR by accident? 
Seems useful though - so you can create an application using the e2e tests in local DB?

bobmeredith (approved):
Nice

server/utils/dateUtils.ts
froddd: You’re right, nice to avoid the try/catch. Though the second question mark probably isn’t needed, we do want 2 digits for the minutes!

PR #2197, opened 2024-11-07
Fix flaky AP Type test

froddd:
Target correct estate for AP type tests, by specifying the `sex` property of the person for the application. This ensures the output for expected radio buttons does not fluctuate between test runs, removing some flakiness.


bobmeredith (approved):
LGTM

PR #2201, opened 2024-11-13
APS-1412 Record non-arrivals

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/APS-1412

# Changes in this PR

Adds 'Record non-arrival' action to placement details page.
The option and page routes require the permission `cas1_space_booking_record_non_arrival` to be enabled

## Screenshots of UI changes
Placement detail page with open actions menu
image

Record non-arrival page
image

Checks on reason and character count
image

Completion showing confirmation and updated details
image

### DIFF
--- a/integration_tests/mockApis/spaceBooking.ts
+++ b/integration_tests/mockApis/spaceBooking.ts
@@ -93,4 +93,18 @@ export default {
         status: 200,
       },
     }),
+
+  stubSpaceBookingNonArrival: placement =>
+    stubFor({
+      request: {
+        method: 'POST',
+        urlPattern: paths.premises.placements.nonArrival({
+          premisesId: placement.premises.id,
+          placementId: placement.id,
+        }),
+      },
+      response: {
+        status: 200,
+      },
+    }),
 }
--- a/integration_tests/mockApis/spaceBooking.ts
+++ b/integration_tests/mockApis/spaceBooking.ts
@@ -107,4 +107,13 @@ export default {
         status: 200,
       },
     }),
+
+  getApiPost: async (path: string): Promise<Record<string, unknown>> => {
+    const result = await getMatchingRequests({
+      method: 'POST',
+      url: path,
+    })
+    const { requests } = result.body
+    return JSON.parse(requests[0].body)
+  },
 }
--- a/integration_tests/mockApis/spaceBooking.ts
+++ b/integration_tests/mockApis/spaceBooking.ts
@@ -107,13 +107,4 @@ export default {
         status: 200,
       },
     }),
-
-  getApiPost: async (path: string): Promise<Record<string, unknown>> => {
-    const result = await getMatchingRequests({
-      method: 'POST',
-      url: path,
-    })
-    const { requests } = result.body
-    return JSON.parse(requests[0].body)
-  },
 }
--- a/integration_tests/pages/manage/placements/nonArrival.ts
+++ b/integration_tests/pages/manage/placements/nonArrival.ts
@@ -1,19 +1,17 @@
 import { faker } from '@faker-js/faker'
-import type { Cas1SpaceBooking } from '@approved-premises/api'
+import type { Cas1NonArrival, Cas1SpaceBooking } from '@approved-premises/api'
 import Page from '../../page'
 import paths from '../../../../server/paths/manage'
+import apiPaths from '../../../../server/paths/api'
 
 export class RecordNonArrivalPage extends Page {
   constructor(title: string = 'Record someone as not arrived') {
     super(title)
   }
 
-  shouldShowReasonError(): void {
-    cy.get('.govuk-error-summary__list').should('contain', 'You must select a reason for non-arrival')
-  }
-
-  shouldShowTextError(): void {
-    cy.get('.govuk-error-summary__list').should('contain', 'You have exceeded 200 characters')
+  nonArrivalDetails: Cas1NonArrival = {
+    notes: faker.lorem.words(20).substring(0, 200),
+    reason: null,
   }
 
   completeNotesBad(): void {
@@ -24,7 +22,7 @@ export class RecordNonArrivalPage extends Page {
   }
 
   completeNotesGood(): void {
-    this.completeTextArea('notes', faker.lorem.words(20).substring(0, 200))
+    this.completeTextArea('notes', this.nonArrivalDetails.notes)
   }
 
   checkForCharacterCountError(): void {
@@ -35,6 +33,7 @@ export class RecordNonArrivalPage extends Page {
     cy.get('input[name=reason]')
       .invoke('val')
       .then(val => {
+        this.nonArrivalDetails.reason = String(val)
         this.checkRadioByNameAndValue('reason', String(val))
       })
   }
@@ -44,5 +43,14 @@ export class RecordNonArrivalPage extends Page {
       failOnStatusCode: false,
     })
     return new RecordNonArrivalPage(`Authorisation Error`)
+  }
+
+  checkApiCalled(placement: Cas1SpaceBooking): void {
+    cy.task(
+      'getApiPost',
+      apiPaths.premises.placements.nonArrival({ premisesId: placement.premises.id, placementId: placement.id }),
+    ).then(body => {
+      expect(body).to.deep.equal(this.nonArrivalDetails)
+    })
   }
 }
--- a/integration_tests/pages/manage/placements/nonArrival.ts
+++ b/integration_tests/pages/manage/placements/nonArrival.ts
@@ -1,0 +1,48 @@
+import { faker } from '@faker-js/faker'
+import type { Cas1SpaceBooking } from '@approved-premises/api'
+import Page from '../../page'
+import paths from '../../../../server/paths/manage'
+
+export class RecordNonArrivalPage extends Page {
+  constructor(title: string = 'Record someone as not arrived') {
+    super(title)
+  }
+
+  shouldShowReasonError(): void {
+    cy.get('.govuk-error-summary__list').should('contain', 'You must select a reason for non-arrival')
+  }
+
+  shouldShowTextError(): void {
+    cy.get('.govuk-error-summary__list').should('contain', 'You have exceeded 200 characters')
+  }
+
+  completeNotesBad(): void {
+    // Generate a text string of exactly 250 chars with no trailing space - so we have a repeatable error message
+    const words = faker.lorem.words(100).substring(0, 250).trim()
+    const text = words + 'abc'.substring(0, 250 - words.length)
+    this.completeTextArea('notes', text)
+  }
+
+  completeNotesGood(): void {
+    this.completeTextArea('notes', faker.lorem.words(20).substring(0, 200))
+  }
+
+  checkForCharacterCountError(): void {
+    cy.contains('You have 50 characters too many')
+  }
+
+  completeReason(): void {
+    cy.get('input[name=reason]')
+      .invoke('val')
+      .then(val => {
+        this.checkRadioByNameAndValue('reason', String(val))
+      })
+  }
+
+  static visitUnauthorised(placement: Cas1SpaceBooking): RecordNonArrivalPage {
+    cy.visit(paths.premises.placements.nonArrival({ premisesId: placement.premises.id, placementId: placement.id }), {
+      failOnStatusCode: false,
+    })
+    return new RecordNonArrivalPage(`Authorisation Error`)
+  }
+}
--- a/integration_tests/pages/manage/placements/nonArrival.ts
+++ b/integration_tests/pages/manage/placements/nonArrival.ts
@@ -47,7 +47,7 @@ export class RecordNonArrivalPage extends Page {
 
   checkApiCalled(placement: Cas1SpaceBooking): void {
     cy.task(
-      'getApiPost',
+      'verifyApiPost',
       apiPaths.premises.placements.nonArrival({ premisesId: placement.premises.id, placementId: placement.id }),
     ).then(body => {
       expect(body).to.deep.equal(this.nonArrivalDetails)
--- a/integration_tests/tests/manage/placements/keyworker.cy.ts
+++ b/integration_tests/tests/manage/placements/keyworker.cy.ts
@@ -1,23 +1,25 @@
 import { signIn } from '../../signIn'
 import { KeyworkerAssignmentPage } from '../../../pages/manage/placements/keyworker'
+import { PlacementShowPage } from '../../../pages/manage'
 import {
   cas1PremisesSummaryFactory,
   cas1SpaceBookingFactory,
   staffMemberFactory,
 } from '../../../../server/testutils/factories'
-import { PlacementShowPage } from '../../../pages/manage'
+
+const premises = cas1PremisesSummaryFactory.build()
+const placement = cas1SpaceBookingFactory.upcoming().build({ premises })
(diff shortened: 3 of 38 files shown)


integration_tests/pages/manage/placements/nonArrival.ts
```diff
@@ -0,0 +1,48 @@
...
+    super(title)
+  }
+
+  shouldShowReasonError(): void {
+    cy.get('.govuk-error-summary__list').should('contain', 'You must select a reason for non-arrival')
+  }
+
+  shouldShowTextError(): void {
+    cy.get('.govuk-error-summary__list').should('contain', 'You have exceeded 200 characters')
+  }
```
froddd, line 17: The parent class `Page` offers a `shouldShowErrorMessagesForFields` method which means the error messages can be seen directly in the cypress test file, and also tests both the error summary and the error by the field -- e.g.:

```js
// nonArrival.cy.ts

const page = new RecordNonArrivalPage()
// submit the form, etc
page.shouldShowErrorMessagesForFields(['notes'], {
  notes: 'You have exceeded 200 characters'
})
```

As an aside, is there any reason this field has a limit on character count? This seems a bit of an odd outlier.

integration_tests/tests/manage/placements/keyworker.cy.ts
```diff
@@ -45,4 +47,19 @@ context('Keyworker', () => {
...
+    // And I am on the placement page
+    const placementPage = PlacementShowPage.visit(placement)
+
+    // Then the record non-arrival option should not be present
+    placementPage.actionMenuShouldNotExist()
+
+    // When I navigate to the keyworker page directly
+    // Then I see an authorisation error
+    KeyworkerAssignmentPage.visitUnauthorised(placement)
+  })
```
froddd, line 64: Yup, this is really nice

server/controllers/manage/premises/placements/nonArrivalsController.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  new(): RequestHandler {
+    return async (req: Request, res: Response) => {
+      const { token } = req.user
+      const { premisesId, placementId } = req.params
+      const { errors, errorSummary, userInput, errorTitle } = fetchErrorsAndUserInput(req)
+      const placement = await this.premisesService.getPlacement({ token, premisesId, placementId })
+      const nonArrivalReasons: Array<NonArrivalReason> = await this.placementService.getNonArrivalReasons(token)
+      return res.render('manage/premises/placements/non-arrival', {
+        nonArrivalReasons: nonArrivalReasons.filter(({ isActive }) => isActive),
```
froddd, line 22: Is this necessary? I believe the API filters inactive reference data out, so while the type specifies `isActive` as boolean it should never be `false` -- makes me feel I should really check this!

server/services/placementService.test.ts
```diff
@@ -35,14 +46,34 @@ describe('PlacementService', () => {
...
   })
+
+  describe('recordNonArrival', () => {
+    it('calls the recordNonArrival method of the placement client and returns a response', async () => {
+      const nonArrival = cas1NonArrivalFactory.build()
+      placementClient.recordNonArrival.mockResolvedValue({})
+      const result = await placementService.recordNonArrival(token, premisesId, placementId, nonArrival)
+      expect(result).toEqual({})
+      expect(placementClientFactory).toHaveBeenCalledWith(token)
+      expect(placementClient.recordNonArrival).toHaveBeenCalledWith(premisesId, placementId, nonArrival)
```
froddd, line 67: Blank lines help to separate the Given/When/Then parts of tests to make them a bit more readable -- not a deal-breaker, but makes development a bit easier on the eye 🙂 

```suggestion
      const nonArrival = cas1NonArrivalFactory.build()
      placementClient.recordNonArrival.mockResolvedValue({})
      
      const result = await placementService.recordNonArrival(token, premisesId, placementId, nonArrival)
      
      expect(result).toEqual({})
      expect(placementClientFactory).toHaveBeenCalledWith(token)
      expect(placementClient.recordNonArrival).toHaveBeenCalledWith(premisesId, placementId, nonArrival)
```

server/testutils/factories/cas1NonArrival.ts
```diff
@@ -0,0 +1,12 @@
+import { Factory } from 'fishery'
+import { Cas1NonArrival } from '@approved-premises/api'
+import { faker } from '@faker-js/faker'
+import { DateFormats } from '../../utils/dateUtils'
+
+export default Factory.define<Cas1NonArrival>(() => ({
+  reason: faker.word.words(3),
+  notes: faker.lorem.words(50),
```
froddd, line 8: Might that break the 200 character limit in some cases?

server/testutils/factories/cas1NonArrival.ts
```diff
@@ -0,0 +1,12 @@
+import { Factory } from 'fishery'
+import { Cas1NonArrival } from '@approved-premises/api'
+import { faker } from '@faker-js/faker'
+import { DateFormats } from '../../utils/dateUtils'
+
+export default Factory.define<Cas1NonArrival>(() => ({
+  reason: faker.word.words(3),
```
froddd, line 7: Is this supposed to be an id? If so maybe using `faker.string.uuid()` might be more appropriate.

server/testutils/factories/cas1NonArrival.ts
```diff
@@ -0,0 +1,12 @@
...
+  notes: faker.lorem.words(50),
+  confirmedAt: DateFormats.dateObjToIsoDateTime(
+    new Date(faker.date.recent().getTime() + faker.number.int({ max: 1000 * 60 * 60 * 24 })),
```
froddd, line 10: Still don't quite get why this manipulation is needed -- what's wrong with simply using `faker.date.recent()`, which does add time info?

server/testutils/factories/cas1SpaceBooking.ts
```diff
@@ -8,6 +8,7 @@ import userFactory from './user'
 
 import staffMemberFactory from './staffMember'
 import { DateFormats } from '../../utils/dateUtils'
+import cas1SpaceBookingNonArrivalFactory from './cas1SpaceBookingNonArrival'
 
 const addTime = (date: Date): Date => new Date(date.getTime() + faker.number.int({ max: 60 * 60 * 24 * 1000 }))
```
froddd, line 13: I still don't think this is needed -- the faker date methods always return full dateTime objects...

server/testutils/factories/cas1SpaceBookingNonArrival.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export default Cas1SpaceBookingNonArrivalFactory.define(() => {
+  return {
+    notes: faker.word.words(20),
+    reason: {
+      id: faker.string.uuid(),
+      name: `${faker.word.noun()} ${faker.word.verb()}`,
+    },
+    confirmedAt: DateFormats.dateObjToIsoDate(
+      new Date(faker.date.recent().getTime() + faker.number.int({ max: 60 * 60 * 24 * 1000 })),
```
froddd, line 16: Again, `faker.date.recent()` does return a time, so adding this time shouldn't be necessary

server/utils/placements/index.test.ts
```diff
@@ -180,6 +206,34 @@ describe('placementUtils', () => {
...
+      expect(arrivalInformation(nonArrivalplacement)).toEqual({
+        rows: [
+          {
+            key: { text: 'Expected arrival date' },
+            value: { text: DateFormats.isoDateToUIDate(expectedArrivalDate) },
+          },
+          {
+            key: { text: 'Non arrival recorded at' },
+            value: {
+              text: `${DateFormats.isoDateToUIDate(confirmedAt)} ${DateFormats.timeFromDate(DateFormats.isoToDateObj(confirmedAt))}`,
```
froddd, line 222: Wasn't aware there was a `timeFromDate` utility -- might rewrite a couple of things!

server/utils/placements/index.ts
```diff
@@ -7,29 +7,40 @@ import paths from '../../paths/manage'
...
+  if (!departed && !nonArrival && hasPermission(user, ['cas1_space_booking_record_keyworker'])) {
+    actionList.push({
       text: 'Assign keyworker',
       classes: 'govuk-button--secondary',
       href: paths.premises.placements.keyworker({ premisesId: placement.premises.id, placementId: placement.id }),
     })
   }
-  if (!placement.actualArrivalDate && hasPermission(user, ['cas1_space_booking_record_arrival'])) {
-    out.push({
+  if (!arrived && !nonArrival && hasPermission(user, ['cas1_space_booking_record_arrival'])) {
```
froddd, line 22: I guess this and the next if/else conditional should both satisfy `!arrived && !nonArrival && !departed` so could be combined into one?

server/views/manage/premises/placements/non-arrival.njk
```diff
@@ -0,0 +1,76 @@
...
+
+{% block content %}
+
+    <form action="{{ paths.premises.placements.nonArrival({ premisesId: placement.premises.id, placementId: placement.id }) }}"
+          method="post">
+        <input type="hidden" name="_csrf" value="{{ csrfToken }}" />
+
+        {{ showErrorSummary(errorSummary, errorTitle) }}
+
+        {% include "../../../_messages.njk" %}
```
froddd, line 33: I don't think there should ever be a message on this page?

froddd (commented):
A few comments, looks good!

server/controllers/manage/premises/placements/nonArrivalsController.ts
bobmeredith: Oh, OK. I kinda assumed that if the API was filtering them out, it wouldn't bother returning the flag. I'll test this - I think all the test data has the flag set true.

integration_tests/mockApis/spaceBooking.ts
```diff
@@ -93,4 +93,18 @@ export default {
         status: 200,
       },
     }),
+
+  stubSpaceBookingNonArrival: placement =>
```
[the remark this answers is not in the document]
bobmeredith: OK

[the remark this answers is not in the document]
bobmeredith: Probably simpler to have a generic getApiPost task taking the path. Then it can be used by everything

server/testutils/factories/cas1NonArrival.ts
bobmeredith: Oops - yes, it's an ID, not the reason name. Good spot!

server/testutils/factories/cas1SpaceBookingNonArrival.ts
bobmeredith: You're quite right

server/utils/placements/index.ts
bobmeredith: Combining the tests probably makes the code less readable but OK.

server/views/manage/premises/placements/non-arrival.njk
bobmeredith: True

integration_tests/mockApis/spaceBooking.ts
```diff
@@ -93,4 +93,27 @@ export default {
...
+    }),
+
+  getApiPost: async (path: string): Promise<Record<string, unknown>> => {
+    const result = await getMatchingRequests({
+      method: 'POST',
+      url: path,
+    })
+    const { requests } = result.body
+    return JSON.parse(requests[0].body)
+  },
```
froddd, line 118: That's great, may I suggest moving this to `integration_tests/mockApis/utils.ts`? Seems generic enough. Also maybe keep the `verify...` naming, so `verifyApiPost`? Would align with existing implementations.

server/testutils/factories/cas1NonArrival.ts
```diff
@@ -0,0 +1,10 @@
+import { Factory } from 'fishery'
+import { Cas1NonArrival } from '@approved-premises/api'
+import { faker } from '@faker-js/faker'
+import { DateFormats } from '../../utils/dateUtils'
+
+export default Factory.define<Cas1NonArrival>(() => ({
+  reason: faker.string.uuid(),
+  notes: faker.lorem.words(20),
+  confirmedAt: DateFormats.dateObjToIsoDateTime(new Date(faker.date.recent())),
```
froddd, line 9: Faker does return a `Date` object so the `new Date()` is unnecessary here:

```suggestion
  confirmedAt: DateFormats.dateObjToIsoDateTime(faker.date.recent()),
```

server/utils/placements/index.ts
froddd: I think I was thinking of these conditionals as matching 'states' for the placement, effectively:

- Upcoming: no arrival or non-arrival recorded (show 'Record arrival' and 'Record non-arrival')
- Current: arrival recorded, no departure recorded (show 'Record departure')
- Historic: departure or non-arrival recorded (don't show anything?)

Similarly, there's a further conditional that checks `arrived && !nonArrival && !departed` -- surely if a placement has an arrival, it can't ever have a non-arrival as well?

server/controllers/manage/premises/placements/nonArrivalsController.ts
froddd: If that's the case, that seems to be a job for the service rather than the controller? I don't think we'll ever make use of inactive items, so let's remove them as early as possible.

server/testutils/factories/cas1NonArrival.ts
bobmeredith: True

server/utils/placements/index.ts
bobmeredith: True. I was making the tests independent so that it would be easier to re-arrange the menu options in the future if required. You're quite right - you can only depart a placement that's arrived.

server/controllers/manage/premises/placements/nonArrivalsController.ts
bobmeredith: OK

server/services/placementService.ts
```diff
@@ -21,4 +24,17 @@ export default class PlacementService {
 
     return placementClient.assignKeyworker(premisesId, placementId, keyworkerAssignment)
   }
+
+  async getNonArrivalReasons(token: string): Promise<Array<NonArrivalReason>> {
+    const client = this.referenceDataClientFactory(token)
+    const allNonArrivalReasons = await client.getNonArrivalReasons()
+
+    return allNonArrivalReasons.filter(({ isActive }) => isActive)
```
froddd, line 32: As discussed, the API should only ever return reference data where `isActive = true`, so we'll be able to remove this when the API is fixed for this endpoint.

server/testutils/factories/cas1SpaceBookingNonArrival.ts
```diff
@@ -0,0 +1,17 @@
...
+class Cas1SpaceBookingNonArrivalFactory extends Factory<Cas1SpaceBookingNonArrival> {}
+
+export default Cas1SpaceBookingNonArrivalFactory.define(() => {
+  return {
+    notes: faker.word.words(20),
+    reason: {
+      id: faker.string.uuid(),
+      name: `${faker.word.noun()} ${faker.word.verb()}`,
+    },
+    confirmedAt: DateFormats.dateObjToIsoDateTime(new Date(faker.date.recent())),
```
froddd, line 15: Don't need `new Date()` here

froddd (approved):
👍

PR #2204, opened 2024-11-18
APS-1548 Control create booking type by permission

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/APS-1548

# Changes in this PR
the actions menu on the placement request page will now allow users to create legacy bookings or to search for an AP (and create a space booking/placement) depending on user permissions:

- `cas1_booking_create` enables the legacy create placement option.
- `cas1_space_booking_create` enables the creation of a space booking via 'search for a space'

As it includes the latest set of API permissions and roles, I've also needed to handle a new role `cru_member_find_and_book_beta`

## Screenshots of UI changes
With permission `cas1_booking_create` but without `cas1_space_booking_create`
image

With permission `cas1_space_booking_create` but without `cas1_booking_create`
image

### DIFF
--- a/server/utils/users/roles.ts
+++ b/server/utils/users/roles.ts
@@ -72,6 +72,10 @@ export const roleLabelDictionary: RoleLabelDictionary = {
     label: 'CRU member',
     hint: 'Manage out of service beds across all premises and areas',
   },
+  cru_member_find_and_book_beta: {
+    label: 'CRU member beta (find and book)',
+    hint: 'Create and amend space bookings',
+  },
   report_viewer: {
     label: 'Report Viewer',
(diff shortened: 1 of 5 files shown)


server/utils/users/roles.ts
froddd, line 1: Given we're not using this role for any of the functionality covered in this PR, I'd prefer keeping these changes out for now. This would make it easier to revert single changes if needed.

bobmeredith: Yeah, makes sense.  I didn't realise the power of the unused list to keep a new role quiet.

froddd (approved):
👍

PR #2206, opened 2024-11-19
CHORE: Add option to assess to application create script

froddd:
This adds the following scripts to help create test data locally:

- `create-application:standard:assess`
- `create-application:emergency:assess`
- `create-application:short-notice:assess`

In addition to creating the application like the existing scripts, these also go through the assessment so the application is ready to match under the CRU dashboard.


bobmeredith (approved):
Luverly!
That's going to save some time.

PR #2208, opened 2024-11-20
APS-1562 Legacy bookings only allow legacy premises

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/APS-1562

# Changes in this PR
Filters out premises that have `supportsSpaceBookings` set true when on the legacy bookings creation page within CRU dashboard.

image
Note: the SWSC region has been removed because all its premises allow space bookings.

### DIFF
--- a/server/controllers/admin/placementRequests/bookingsController.test.ts
+++ b/server/controllers/admin/placementRequests/bookingsController.test.ts
@@ -1,6 +1,6 @@
 import type { NextFunction, Request, Response } from 'express'
 import { DeepMocked, createMock } from '@golevelup/ts-jest'
-import type { ApprovedPremisesApplication } from '@approved-premises/api'
+import type { ApprovedPremisesApplication, Cas1PremisesBasicSummary } from '@approved-premises/api'
 import type { ErrorsAndUserInput } from '@approved-premises/ui'
 import { PlacementRequestService, PremisesService } from '../../../services'
 import {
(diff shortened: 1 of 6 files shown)


server/controllers/admin/placementRequests/bookingsController.test.ts
```diff
@@ -52,22 +62,32 @@ describe('PlacementRequestsController', () => {
...
+
+      const requestHandler = bookingsController.new()
+
+      await requestHandler(request, response, next)
+
+      expect(response.render).toHaveBeenCalledWith('admin/placementRequests/bookings/new', {
+        ...expectedRenderParameters,
+        premises: mixedPremises.filter(({ supportsSpaceBookings }) => !supportsSpaceBookings),
+      })
+    })
```
froddd, line 89: Given this is a very focused test, the assert would maybe benefit from using the `expect.objectContaining()` approach:

```suggestion
      expect(response.render).toHaveBeenCalledWith('admin/placementRequests/bookings/new', expect.objectContaining({
        premises: mixedPremises.filter(({ supportsSpaceBookings }) => !supportsSpaceBookings),
      })
    })
```

froddd (approved):
👍

server/controllers/admin/placementRequests/bookingsController.test.ts
bobmeredith: Yeah, makes sense.

PR #2213, opened 2024-11-22
Navigate to placeholder occupancy view

gregkhawkins:
# Context

* This PR relates to: https://dsdmoj.atlassian.net/browse/APS-1577
* As discussed with Fred and Bob, this PR does enough to unblock this and forthcoming `Occupancy View` tickets

# Changes in this PR
* Navigate to new `/view-spaces` route when click on the `View Spaces` link on the `Find a space page`
* Render a bare bones view (for this new route) with banner and header 
* The result of the above is that we hit this road block view (instead of navigating the `Confirm Booking` page to continue and finish the user journey) - **_please shout if I misunderstood and this is bad!_**
* Changes to e2e tests so they do not fail when they hit the new road block
* All e2e tests run and pass locally against AP tools

## Screenshots of UI changes
occupancy_view

### DIFF
--- a/e2e/steps/match.ts
+++ b/e2e/steps/match.ts
@@ -1,22 +1,12 @@
 import { Page } from '@playwright/test'
 import { visitDashboard } from './apply'
-import { ConfirmPage, ConfirmationPage } from '../pages/match'
 import { E2EDatesOfPlacement } from './assess'
 import { ListPage, PlacementRequestPage } from '../pages/workflow'
 import { ApprovedPremisesApplication as Application, Premises } from '../../server/@types/shared'
 import { ApTypeLabel } from '../../server/utils/apTypeLabels'
 import { SearchScreen } from '../pages/match/searchScreen'
 import { BookingScreen } from '../pages/match/bookingScreen'
-
-export const confirmBooking = async (page: Page) => {
-  const confirmPage = new ConfirmPage(page)
-  await confirmPage.clickConfirm()
-}
-
-export const shouldShowBookingConfirmation = async (page: Page) => {
-  const confirmationPage = new ConfirmationPage(page)
-  await confirmationPage.shouldShowSuccessMessage()
-}
+import { OccupancyViewScreen } from '../pages/match/occupancyViewScreen'
 
 export const matchAndBookApplication = async ({
   applicationId,
@@ -78,6 +68,11 @@ export const matchAndBookApplication = async ({
   // And I select an AP
   const premisesName = await searchScreen.retrieveFirstAPName()
   await searchScreen.selectFirstAP()
+
+  // Then I should see the occupancy view screen for that AP
+  const occupancyViewScreen = await OccupancyViewScreen.initialize(page, premisesName)
+  // And I continue to booking
+  await occupancyViewScreen.clickContinue()
 
   // Then I should see the booking screen for that AP
   const bookingScreen = await BookingScreen.initialize(page, premisesName)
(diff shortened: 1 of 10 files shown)


e2e/steps/match.ts
```diff
@@ -79,15 +68,6 @@ export const matchAndBookApplication = async ({
   const premisesName = await searchScreen.retrieveFirstAPName()
   await searchScreen.selectFirstAP()
 
-  // Then I should see the booking screen for that AP
-  const bookingScreen = await BookingScreen.initialize(page, premisesName)
```
bobmeredith, line 83: I take it that the confirmation page is no longer reachable. This would be a bit annoying as it means we can no longer create space-bookings through the UI.

How about adding a button on the new occupancy view page that jumps to the confirmation page, then you can add this to the e2e tests without needing to remove the confirmation page tests.

server/controllers/match/placementRequests/occupancyViewController.ts
```diff
@@ -0,0 +1,23 @@
+import type { Request, Response, TypedRequestHandler } from 'express'
+import { PlacementRequestService } from '../../../services'
+
+interface NewRequest extends Request {
+  params: { id: string }
+  query: { premisesName: string }
```
bobmeredith, line 6: I think this page will be needing lots of information from the premises - so you'll be needing the id passing in and make API requests to get everything including the name.

froddd (commented):
Looks good, couple of comments!

e2e/steps/match.ts
gregkhawkins: okay great! I was a bit surprised we were blocking this journey entirely. I will add in the button to navigate to the confirmation page as suggest - thanks both :)

server/controllers/match/placementRequests/occupancyViewController.ts
gregkhawkins: yep, thanks. I've got that in a local branch that gets the matching details data so that can follow if okay?

server/views/match/placementRequests/occupancyView/view.njk
```diff
@@ -0,0 +1,24 @@
...
+{% endblock %}
+
+{% block content %}
+    <div class="govuk-width-container">
+        <div class="govuk-grid-row">
+            <div class="govuk-grid-column-two-thirds">
+                <h1 class="govuk-heading-l">{{ pageHeading }}</h1>
+            </div>
+        </div>
+    </div>
```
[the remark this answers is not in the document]
gregkhawkins: Okay perfect! Will make this change also

server/controllers/match/placementRequests/occupancyViewController.ts
gregkhawkins: actually, in adding the new `Continue` button (as per convo above) I needed to pass all this info now so done as part of this PR afterall. Thanks!

server/views/match/placementRequests/occupancyView/view.njk
```diff
@@ -0,0 +1,21 @@
...
+        text: "Back to other APs",
+        href: paths.v2Match.placementRequests.search.spaces({ id: placementRequest.id })
+    }) }}
+{% endblock %}
+
+{% block content %}
+    <h1 class="govuk-heading-l">{{ pageHeading }}</h1>
+    <a class="govuk-button" href="{{ MatchUtils.redirectToSpaceBookingsNew({placementRequestId: placementRequest.id, premisesName: premisesName, premisesId: premisesId, apType: apType, startDate: startDate, durationDays: durationDays }) }}">
+        Continue
+    </a>
```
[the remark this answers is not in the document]
froddd: Absolutely fine, yes! It doesn't matter what it looks like right now, only that the journey can be completed for test purposes :)

froddd (approved):
LGTM!

bobmeredith (approved):
Looks good to me

PR #2214, opened 2024-11-25
APS-1615: E2E tests for Manage flow

froddd:
# Context

https://dsdmoj.atlassian.net/browse/APS-1615

# Changes in this PR

This adds the end to end tests to cover the standard Manage flow at the end of the Standard application test:

- [x] Edit keyworker
- [x] Record arrival
- [x] Record departure

NOTE: it may be possible to separate those actions and have them as standalone tests, not tacked on the end of the Standard application flow: at the moment the application returns useful info (placement date, premises booked in, etc) to select the correct placement on the premises page, but it should be possible to select the _first_ available placement without those arguments passed in. This should enable testing other paths, like recording non-arrival, etc. May also speed up tests, but would require solid seed data!


froddd:
Superseded by https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/2217 to avoid having to deal with a gazillion conflicts 😄

PR #2215, opened 2024-11-26
Remove placement matching outcome reports

davidatkinsuk:
These are not currently used and will be replaced with a general placements report as part of find and book

# Context

# Changes in this PR


froddd (approved):
👍

PR #2222, opened 2024-11-28
APS-1410 Status in placements table

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/APS-1410

# Changes in this PR
Adds a status column to the ends of the placement table in the premises details view for match&manage.
The statuses are translated to text via a lookup table so that the UI can be changed in the future although the current status keys could be transformed programatically at the moment

## Screenshots of UI changes

image

image


froddd (approved):
👍

PR #2225, opened 2024-11-29
Occupancy view - add matching details section

gregkhawkins:
# Context

- Related JIRA: https://dsdmoj.atlassian.net/browse/APS-1577
- Do you need to add any environment variables?: Not this time

# Changes in this PR

- Introduces the `Matching details` section to the `Occupancy View page` UI
- Integrates with the `GET /cas1/premises/{premiseId}/capacity` endpoint to get the `Total capacity` data needed for the UI
- All test coverage required (inc. unit / integration / e2e)

### After
occupncy_view_manage_details


server/views/match/placementRequests/occupancyView/view.njk
```diff
@@ -15,6 +16,16 @@
 
 {% block content %}
     <h1 class="govuk-heading-l">{{ pageHeading }}</h1>
+    {% set detailsHTML %}
+        {{ govukSummaryList({
+            rows: MatchUtils.occupancyViewSummaryListForMatchingDetails(totalCapacity, dates, placementRequest, essentialCharacteristics)
```
froddd, line 21: This pattern of creating a utility that's used directly in the template has been used everywhere historically, but I think we should move away from it: it relies on the integration tests to be covered properly, and these are somewhat slow, and it doesn't help discoverability as one has to do a manual search for that utility method name to figure out if it is used in a template, versus having the IDE let you discover usages. 

A more testable approach would be to call those utilities in the controller directly, and simply return directly usable data/variables to the template. 

Not necessarily one to change right now, but let's have a chat about this when we can!

froddd (commented):
Looks good -- couple of comments!

e2e/pages/match/matchBasePage.ts
```diff
@@ -0,0 +1,19 @@
+import { Page, expect } from '@playwright/test'
+import { E2EDatesOfPlacement } from '../../steps/assess'
+import { DateFormats } from '../../../server/utils/dateUtils'
+
+export class MatchBasePage {
```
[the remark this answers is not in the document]
gregkhawkins: ah yes, good spot! Thanks

e2e/pages/match/searchScreen.ts
```diff
@@ -1,11 +1,10 @@
 import { Page, expect } from '@playwright/test'
-import { BasePage } from '../basePage'
+import { MatchBasePage } from './matchBasePage'
 import { Premises } from '../../../server/@types/shared'
 import { ApTypeLabel } from '../../../server/utils/apTypeLabels'
 import { E2EDatesOfPlacement } from '../../steps/assess'
-import { DateFormats } from '../../../server/utils/dateUtils'
 
-export class SearchScreen extends BasePage {
+export class SearchScreen extends MatchBasePage {
```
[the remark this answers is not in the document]
gregkhawkins: perfect, I noticed that there was another one `BookingScreen` so will change that too

integration_tests/tests/match/match.cy.ts
```diff
@@ -102,6 +104,39 @@ context('Placement Requests', () => {
...
+  it('allows me to view spaces and occupancy capacity', () => {
+    const premisesName = 'Hope House'
+    const apType = 'normal'
+    const durationDays = 15
+    const startDate = '2024-07-23'
+    const endDate = '2024-08-07'
+    const totalCapacity = 10
+
+    // Given I am signed in as a cru_member
+    signIn(['cru_member'])
```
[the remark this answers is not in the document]
gregkhawkins: thanks for raising this, I will ask the question on the #cas-dev slack channel (once I've fixed up the rest)

server/views/match/placementRequests/occupancyView/view.njk
gregkhawkins: have refacored as suggested, agree this is much nicer! Keep the views as dumb as possible

integration_tests/tests/match/match.cy.ts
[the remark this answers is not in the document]
gregkhawkins: baed on what was discussed on the slack thread: https://mojdt.slack.com/archives/C03K0HB0LBE/p1733155155157929

I have changed from `signIn(['cru_member'])` to `signIn(['cru_member'], ['cas1_space_booking_create'])`

Hopefully I interpreted this correctly?

e2e/pages/match/occupancyViewPage.ts
```diff
@@ -0,0 +1,34 @@
...
+  async shouldShowMatchingDetails({
+    datesOfPlacement,
+    duration,
+    releaseType,
+  }: {
+    datesOfPlacement: E2EDatesOfPlacement
+    duration: string
+    releaseType: ReleaseTypeLabel
+  }): Promise<void> {
+    await super.shouldShowDatesOfPlacement(datesOfPlacement, duration)
```
froddd, line 23: Again, I'm not sure about starting to use `super` instead of simply `this` -- not on a fundamental level, but for consistency, as it's not something we've used anywhere else.

e2e/pages/match/searchPage.ts
```diff
@@ -33,7 +32,7 @@ export class SearchScreen extends BasePage {
     apType: ApTypeLabel
     preferredPostcode: string
   }): void {
-    this.shouldShowDatesOfPlacement(datesOfPlacement, duration)
+    super.shouldShowDatesOfPlacement(datesOfPlacement, duration)
```
froddd, line 35: Another use of `super` instead of `this` I would get rid of!

Also, all the methods in `shouldShowApplicationDetails` are Promises, so they should all be `await`-ed and shouldShowApplicationDetails` should be async (aware this wasn't your code but let's fix if we can!)

e2e/steps/match.ts
```diff
@@ -71,12 +74,20 @@ export const matchAndBookApplication = async ({
...
+    duration,
+    releaseType,
+  })
+
   // And I continue to booking
   await occupancyViewScreen.clickContinue()
 
   // Then I should see the booking screen for that AP
-  const bookingScreen = await BookingScreen.initialize(page, premisesName)
+  const bookingScreen = await BookingPage.initialize(page, premisesName)
```
froddd, line 90: Same as above -- let's rename this `bookingPage`

integration_tests/tests/match/match.cy.ts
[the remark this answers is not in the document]
froddd: Yup, that seems fine!

e2e/pages/match/searchPage.ts
gregkhawkins: sure, no worries

froddd (approved):
Perfect!

PR #2230, opened 2024-12-03
APS-1646: filter keyworkers

froddd:
# Context

https://dsdmoj.atlassian.net/browse/APS-1646

# Changes in this PR

Ensure list of keyworker options in placements list filter is only showing keyworkers. Introduces a new service method that returns a list of staff members filtered for keyworkers only.

Also fixes an 'undefined' appearing in document title for placement page -- now renders as 'Approved Premises - {placement dates}'.

### Before

Screenshot 2024-12-03 at 10 59 50

### After

Screenshot 2024-12-03 at 10 59 20


bobmeredith (approved):
Nice,
Makes sense moving the keyworker filtering to the service.

PR #2231, opened 2024-12-03
Re-add the v1 placement matching outcomes report

davidatkinsuk:
This is still in use, so is being readded

# Context

# Changes in this PR

### After

Screenshot 2024-12-03 at 11 11 27


RobBoothMOJ (approved):
Approved

PR #2235, opened 2024-12-03
APS-1410 status in placement view

bobmeredith:
# Context
https://dsdmoj.atlassian.net/browse/APS-1410

# Changes in this PR
Adds the placement status to the placement details page.
Also, switches to use split date/time values for actual arrival/departure dates. Once this PR is merged, the deprecated values `actualArrivalDate` and `actualDepartureDate` are no longer used by the UI and can be removed.

## Screenshots of UI changes

image

### DIFF
--- a/server/testutils/factories/cas1SpaceBooking.ts
+++ b/server/testutils/factories/cas1SpaceBooking.ts
@@ -1,6 +1,11 @@
 import { Factory } from 'fishery'
 import { faker } from '@faker-js/faker'
-import { Cas1KeyWorkerAllocation, Cas1SpaceBooking, Person } from '@approved-premises/api'
+import type {
+  Cas1KeyWorkerAllocation,
+  Cas1SpaceBooking,
+  Cas1SpaceBookingSummaryStatus,
+  Person,
+} from '@approved-premises/api'
 import { fullPersonFactory } from './person'
 import cas1SpaceBookingDatesFactory from './cas1SpaceBookingDates'
 import cas1SpaceBookingRequirementsFactory from './spaceBookingRequirements'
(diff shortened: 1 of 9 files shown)


server/testutils/factories/cas1SpaceBooking.ts
```diff
@@ -107,5 +129,6 @@ export default Cas1SpaceBookingFactory.define(() => {
     otherBookingsInPremisesForCrn: cas1SpaceBookingDatesFactory.buildList(4),
     deliusEventNumber: String(faker.number.int()),
     nonArrival: undefined,
-  }
+    status: 'departed',
+  } as Cas1SpaceBooking
```
froddd, line 133: Would casting `status` above as `Cas1SpaceBookingSummaryStatus` remove the need to type cast the whole faked booking?

server/utils/dateUtils.ts
```diff
@@ -50,6 +50,14 @@ export class DateFormats {
     return formatISO(date)
   }
 
+  /**
+   * @param date JS Date object.
+   * @returns the time in the format '19:05'.
+   */
+  static dateObjToIsoTime(date: Date) {
```
froddd, line 57: Pedantic nitpick of the day: `19:05` isn't actually ISO time -- that would be `19:05:00(.000)Z`, while `19:05` is 24hr time 😄

server/utils/placements/index.test.ts
```diff
@@ -197,11 +197,11 @@ describe('placementUtils', () => {
           },
           {
             key: { text: 'Actual arrival date' },
-            value: { text: DateFormats.isoDateToUIDate(placement.actualArrivalDate) },
+            value: { text: DateFormats.isoDateToUIDate(placement.actualArrivalDateOnly) },
           },
           {
             key: { text: 'Arrival time' },
-            value: { text: DateFormats.timeFromDate(DateFormats.isoToDateObj(placement.actualArrivalDate)) },
+            value: { text: (placement.actualArrivalTime || '').substring(0, 5) },
```
froddd, line 204: Is this `substring` needed here? The factory already returns time in 24hr format, I think? 🤔

froddd (approved):
Minor comments, seems good to go!

server/utils/placements/index.test.ts
bobmeredith: Yes, you're right. The API currently includes seconds so I originally made the factory do likewise.  Not any more though.

server/testutils/factories/cas1SpaceBooking.ts
bobmeredith: thanks, I'll try that.

PR #2236, opened 2024-12-04
APS-1578: Occupancy summary box

froddd:
# Context

https://dsdmoj.atlassian.net/browse/APS-1578

# Changes in this PR

Adds the occupancy summary banner to the occupancy view page. 

### Dates available

Screenshot 2024-12-04 at 11 37 13

---

### No dates available

Screenshot 2024-12-04 at 11 36 40

---

### Mix of available and overbooked

Screenshot 2024-12-04 at 11 33 30

### DIFF
--- a/server/utils/match/occupancySummary.ts
+++ b/server/utils/match/occupancySummary.ts
@@ -1,0 +1,74 @@
+import type { Cas1PremiseCapacity, Cas1PremiseCapacityForDay } from '@approved-premises/api'
+import { differenceInDays } from 'date-fns'
+import { DateFormats, daysToWeeksAndDays } from '../dateUtils'
+
+export type DateRange = {
+  start: string
+  end: string
+}
+
+const daysToRanges = (days: Array<Cas1PremiseCapacityForDay>) =>
+  days.reduce((ranges: Array<DateRange>, capacityForDay) => {
+    if (!ranges.length) {
+      ranges.push({ start: capacityForDay.date, end: capacityForDay.date })
+    } else {
+      const lastRange = ranges[ranges.length - 1]
+      if (differenceInDays(capacityForDay.date, lastRange.end) === 1) {
+        lastRange.end = capacityForDay.date
+      } else {
+        ranges.push({ start: capacityForDay.date, end: capacityForDay.date })
+      }
+    }
+    return ranges
+  }, [])
+
+export const renderDateRange = (dateRange: DateRange): string => {
+  let render = `${DateFormats.isoDateToUIDate(dateRange.start)}`
+
+  const totalDays =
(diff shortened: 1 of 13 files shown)


server/utils/match/occupancySummary.ts
```diff
@@ -0,0 +1,76 @@
...
+      } else {
+        ranges.push({ start: capacityForDay.date, end: capacityForDay.date })
+      }
+    }
+    return ranges
+  }, [])
+
+export const renderDateRange = (dateRange: DateRange): string => {
+  let render = `${DateFormats.isoDateToUIDate(dateRange.start)}`
+  const totalDays = DateFormats.differenceInDays(
```
gregkhawkins, line 27: can we not use the same `differenceInDays` function as on line 16? If not would reduce the need to import `DateFormats`

server/utils/match/occupancySummary.ts
```diff
@@ -0,0 +1,76 @@
...
+      overbookedDays.push(capacityForDay)
+    }
+  })
+
+  if (availableDays.length === 0) {
+    return `<p class="govuk-heading-m">There are no spaces available for the dates you have selected.</p>`
+  }
+  if (overbookedDays.length === 0) {
+    return `<p class="govuk-heading-m">The placement dates you have selected are available.</p>`
+  }
```
gregkhawkins, line 59: this lead to some head scratching but my brain caught up in the end. A coffee and the tests helped the penny drop - can see we are covered :)

gregkhawkins (approved):

gregkhawkins:
That PR took longer than it should have - still getting used to the stack
one minor suggestion re `DateFormats` but not doing any harm :)

server/utils/match/occupancySummary.ts
froddd: Yeah, this whole utility might not be hugely optimised (and it certainly won't win any code golf award) but it does the job!

server/utils/match/occupancySummary.ts
froddd: Yes, that's right actually -- I imported this one before I implemented the week/days format, which means we can now use the other one!

froddd: Addressed as suggested -- `DateFormats` is imported anyway, but this does clean up the syntax a bit. 

Also refactored the calculation of the difference slightly: now simply adding `1` to the difference, rather than checking difference between `addDays(endDate, 1)` and `startDate` -- I think this is more readable now, and removes the need for `addDays` to be imported at all.

gregkhawkins: ah yes, of course DateFormats is used for other stuff. Cool, all sounds good

server/utils/match/occupancySummary.ts
gregkhawkins: it's fine! Just my brain slowly getting into gear after lunch

PR #2238, opened 2024-12-05
CAS-1236 Remove HTML5Shiv as it is no longer used

aliuk2012:
# Context
HTML5Shiv is old and outdated. It was used to polyfill older browsers to help them support html5 elements like nav/section but based on our browser support model we are no longer support IE or old versions of chrome/firefox which may need this polyfill

# Changes in this PR


Rosie-Brigham (approved):
🦖

PR #2239, opened 2024-12-05
Increase max POST body size and parameter count

froddd:
# Context

https://dsdmoj.atlassian.net/browse/APS-1663

# Changes in this PR

This increases the max POST body size for form from the default of `100kb` to `1mb` (https://expressjs.com/en/resources/middleware/body-parser.html#bodyparserurlencodedoptions). It also increases the max parameter count to `2500` to cover rare cases where forms may have in excess of 1000 checkboxes.

## Screenshots of UI changes

n/a


bobmeredith (approved):
LGTM

PR #2242, opened 2024-12-09
APS-1630 Placement view info

bobmeredith:
# Context
https://dsdmoj.atlassian.net/browse/APS-1630

# Changes in this PR

## Screenshots of UI changes

<details>
  <summary>Application tab</summary>

</details>
<details>
  <summary>Assessment tab</summary>

</details>
<details>
  <summary>Placement request tab</summary>

</details>
<details>
  <summary>Placement details tab (default)</summary>

</details>
<details>
  <summary>Timeline tab</summary>

</details>

### DIFF
--- a/integration_tests/pages/page.ts
+++ b/integration_tests/pages/page.ts
@@ -8,6 +8,7 @@ import {
   Cas1OutOfServiceBedRevision as OutOfServiceBedRevision,
   Person,
   PersonAcctAlert,
+  PersonStatus,
   PrisonCaseNote,
   SortOrder,
   TimelineEvent,
@@ -501,6 +502,10 @@ export default abstract class Page {
     cy.get('a.moj-sub-navigation__link').contains(tabName).should('have.attr', 'aria-current', 'page')
   }
 
+  clickTab(tabTitle: string): void {
+    cy.get('.moj-sub-navigation__list').contains(tabTitle).click()
+  }
+
   shouldHaveSelectText(id: string, text: string): void {
     cy.get(`#${id}`).find('option:selected').should('have.text', text)
   }
@@ -529,16 +534,18 @@ export default abstract class Page {
                 timelineEvents[i].createdBy.name,
               )
             }
-            cy.get('.govuk-link').should('have.attr', { time: timelineEvents[i].associatedUrls[0].url })
-            cy.get('.govuk-link').should('contain', timelineEvents[i].associatedUrls[0].type)
+            if (timelineEvents[i].associatedUrls?.length) {
+              cy.get('.govuk-link').should('have.attr', { time: timelineEvents[i].associatedUrls[0].url })
+              cy.get('.govuk-link').should('contain', timelineEvents[i].associatedUrls[0].type)
+            }
             cy.get('time').should('contain', DateFormats.isoDateTimeToUIDateTime(timelineEvents[i].occurredAt))
           })
         })
       })
   }
 
-  shouldShowPersonDetails(person: FullPerson): void {
-    cy.get('dl[data-cy-person-info]').within(() => {
+  shouldShowPersonDetails(person: FullPerson, expectedStatus?: PersonStatus): void {
+    cy.get('dl[data-cy-person-info],div[data-cy-section="person-details"]').within(() => {
       this.assertDefinition('Name', person.name)
       this.assertDefinition('CRN', person.crn)
       this.assertDefinition('Date of Birth', DateFormats.isoDateToUIDate(person.dateOfBirth, { format: 'short' }))
@@ -546,7 +553,9 @@ export default abstract class Page {
       this.assertDefinition('Nationality', person.nationality)
       this.assertDefinition('Religion or belief', person.religionOrBelief)
       this.assertDefinition('Sex', person.sex)
-      cy.get(`[data-cy-status]`).should('have.attr', 'data-cy-status').and('equal', person.status)
+      cy.get(`[data-cy-status]`)
+        .should('have.attr', 'data-cy-status')
+        .and('equal', expectedStatus || person.status)
       this.assertDefinition('Prison', person.prisonName)
     })
   }
--- a/server/controllers/manage/placementController.test.ts
+++ b/server/controllers/manage/placementController.test.ts
@@ -1,41 +1,82 @@
 import type { NextFunction, Request, Response } from 'express'
 import { DeepMocked, createMock } from '@golevelup/ts-jest'
 
-import { cas1SpaceBookingFactory } from '../../testutils/factories'
-import { PremisesService } from '../../services'
+import {
+  applicationFactory,
+  assessmentFactory,
+  cas1SpaceBookingFactory,
+  placementRequestDetailFactory,
+  timelineEventFactory,
+} from '../../testutils/factories'
+import type {
+  ApplicationService,
+  AssessmentService,
+  PlacementRequestService,
+  PlacementService,
+  PremisesService,
+} from '../../services'
 import PlacementController from './placementController'
+import { mapApplicationTimelineEventsForUi } from '../../utils/applications/utils'
 
 describe('placementController', () => {
-  const token = 'SOME_TOKEN'
-
-  let request: DeepMocked<Request>
-  let response: DeepMocked<Response>
+  const token = 'TEST_TOKEN'
   const next: DeepMocked<NextFunction> = createMock<NextFunction>({})
 
+  const placementService = createMock<PlacementService>({})
+  const applicationService = createMock<ApplicationService>({})
+  const assessmentService = createMock<AssessmentService>({})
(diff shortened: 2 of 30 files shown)


davidatkinsuk:
@bobmeredith would you mind checking what happens on this page if the offender is restricted (i.e. an LAO) and the person viewing the page doesn't have access to that user? They shouldn't be able to see application or assessment information or the person's name. This may be better handled as a separate ticket as I don't think the UI design says what to do here. 

CRN X400001 has a restriction locally. If you login as LAORESTRICTED they shouldn't be able to view details for the person

bobmeredith:
So there is no security issue with this - but the behaviour is not nice. ATM, if a user without access to the LAO offender, selects the application or assessment tab, the API returns a 403 and they are immediately logged out.
We just need to disable these links so the only way users can get to the application/assessment tabs is by tampering with the url.

davidatkinsuk:
> So there is no security issue with this - but the behaviour is not nice. ATM, if a user without access to the LAO offender, selects the application or assessment tab, the API returns a 403 and they are immediately logged out. We just need to disable these links so the only way users can get to the application/assessment tabs is by tampering with the url.

Thanks for looking into this @bobmeredith , that doesn't block merging the PR at least. I'll leave the actual reviewing to a UI dev 👍

server/controllers/manage/placementController.ts
```diff
@@ -1,24 +1,74 @@
...
+    private readonly applicationService: ApplicationService,
+    private readonly assessmentService: AssessmentService,
+    private readonly placementRequestService: PlacementRequestService,
+    private readonly placementService: PlacementService,
+    private readonly premisesService: PremisesService,
+  ) {}
 
   show(): RequestHandler {
     return async (req: Request, res: Response) => {
+      const { activeTab = 'placement' } = req.query as Record<string, PlacementTab>
```
froddd, line 31: I'm not a fan of using a query string for the tab -- I'm not sure where the pattern was first introduced, but the routing framework lets us define paths easily, which can point to the same controller method and provide the `req.params` to determine what 'tab' should be shown, so we should be able to use clearly defined paths for each tab, e.g. `/manage/premises/:premisesId/placements/:placementId/application`, etc. At the moment, visiting `?activeTab=foo` does render the page but doesn't highlight any of the tabs as active. The routes also provide a way to define permissions, which can come in handy!

server/utils/placements/index.ts
```diff
@@ -211,5 +213,27 @@ export const renderKeyworkersSelectOptions = (
     })),
 ]
 
+export type PlacementTab = 'application' | 'assessment' | 'placementrequest' | 'placement' | 'timeline'
+
+export const tabTextMap: Record<PlacementTab, string> = {
+  application: 'Application',
+  assessment: 'Assessment',
+  placementrequest: 'Request for placement',
```
froddd, line 221: CamelCase?

server/controllers/manage/placementController.test.ts
```diff
@@ -44,8 +86,76 @@ describe('placementController', () => {
           pageHeading: '16 Nov 2024 to 26 Mar 2025',
           user,
           backLink: null,
+          activeTab: 'placement',
+        }),
+      )
+      expect(premisesService.getPlacement).toHaveBeenCalledWith({ token, placementId: placement.id, premisesId })
+      expect(applicationService.findApplication).not.toHaveBeenCalled()
```
froddd, line 93: Would be nice to assert _all_ the endpoints that have _not_ been called!

server/views/manage/premises/placements/show.njk
```diff
@@ -41,27 +44,57 @@
...
+        {{ applicationReadonlyView(application) }}
+    {% elseif activeTab==='assessment' %}
+      {% for section in SummaryListUtils.summaryListSections(assessment, false) %}
+        <h2 class="govuk-heading-l">{{ section.title }}</h3>
+        {% for task in section.tasks %}
+          {{
+            govukSummaryList(task)
+          }}
+        {% endfor %}
+      {% endfor %}
```
froddd, line 64: There's a mismatched `<h2>` and `</h3>` closing tag in there.

Feels like there should be a macro for this!

server/views/manage/premises/placements/show.njk
```diff
@@ -41,27 +44,57 @@
...
+    {% elseif activeTab==='placementrequest' %}
+        {{ govukSummaryList({
+            attributes: {
+            'data-cy-section': "placement-request-summary"
+            },
+            rows:PlacementRequestUtils.adminSummary(placementRequestDetail).rows
+            }) }}
```
froddd, line 71: Formatting is a bit off on that

froddd (commented):
Couple of small comments, and a rogue mismatched closing tag!

server/controllers/manage/placementController.ts
bobmeredith: True although the query string is used elsewhere. I'll refactor this to use explicit routes then.

server/utils/placements/index.ts
bobmeredith: I used lowercase as this literal goes into the querystring. Probably should be kebab-case but that's way too inconvenient.

server/controllers/manage/placementController.test.ts
bobmeredith: OK

server/views/manage/premises/placements/show.njk
bobmeredith: Also, why didn't the accessibility checker spot this?

server/views/manage/premises/placements/show.njk
bobmeredith: It is indeed.
Do you have an Intelij formatter/highlighter for nunjucks. I've not found one that works. I've noticed you sometimes 'prettify' templates so I'm guessing you have.

bobmeredith: Heh - you use Twig - I saw you left a review on the Twig plugin site!

integration_tests/pages/page.ts
```diff
@@ -538,7 +542,7 @@ export default abstract class Page {
   }
 
   shouldShowPersonDetails(person: FullPerson): void {
-    cy.get('dl[data-cy-person-info]').within(() => {
+    cy.get('dl[data-cy-person-info],div[data-cy-section="person-details"]').within(() => {
```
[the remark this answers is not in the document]
bobmeredith: True.
Have removed the duplication but not looked at the attribute yet. One slight annoyance is that some of these blocks show the person's status, while others show the status at the time of application.

server/views/manage/premises/placements/show.njk
froddd: That is a good question 🤔

integration_tests/pages/page.ts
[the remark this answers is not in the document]
froddd: Very nice!

server/testutils/factories/person.ts
```diff
@@ -39,5 +39,5 @@ export const fullPersonSummaryFactory = Factory.define<FullPersonSummary>(() =>
   crn: getCrn(),
   personType: 'FullPersonSummary',
   name: faker.person.fullName(),
-  isRestricted: false,
+  isRestricted: Boolean(Math.floor(Math.random())),
```
froddd, line 42: There's a faker utility for boolean -- `faker.datatype.boolean()`

froddd (approved):
👍 -- couple of comments but non-blockers!

server/testutils/factories/person.ts
bobmeredith: Yes, there is! I'm pretty sure I didn't write that code actually - it came in with a rebase and I remember resolving the conflict. Very odd!

server/utils/personUtils.ts
```diff
@@ -23,6 +23,14 @@ const isUnknownPerson = (person?: Person): boolean => person?.type === 'UnknownP
 
 const laoName = (person: FullPerson) => (person.isRestricted ? `LAO: ${person.name}` : person.name)
 
+const laoSummaryName = (personSummary: PersonSummary) => {
+  if (personSummary.personType === 'FullPersonSummary') {
+    const { name, isRestricted } = personSummary as FullPersonSummary
+    return isRestricted ? `LAO: ${name}` : name
+  }
+  return personSummary.personType === 'RestrictedPersonSummary' ? 'LAO' : 'Unknown'
+}
```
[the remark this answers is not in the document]
bobmeredith: Fair comments.
Yes, the code is duplicated but we have a `Person` type and a `PersonSummary` type which are mirror images.  They contain a `type`/ `personType` respectively which are themselves different union types and select between a [`RestrictedPerson` and `FullPerson`] and [`RestrictedPersonSummary` and `FullPersonSummary`] types. I thought it simpler to duplicate the code rather than try to handle all that type casting shenanigans in one routine.

PR #2245, opened 2024-12-12
Remove govuk_frontend_toolkit

aliuk2012:
This is ancient and before govuk_frontend. Its also not included in package.json

# Context

# Changes in this PR


froddd (approved):
This is a straightforward removal, but a lot more could be done with the way static resources are setup -- as it stands I believe this setup exposes loads of files directly from node_modules...

PR #2247, opened 2024-12-13
updates for switch over to auto generated swagger docs

daveawc:
This PR updates the code so that we can successfully generate the UI code when using auto-generated swagger documentation. The changes are:

- The AnyValue class now generates as Unit, so this has been updated where used.
- There are a number of ArrayOfOasys.... classes that are just a wrapper for a list of OASysQuestion with different names, so these have renamed to use Array<OasysQuestion> (which is what the API uses... it does not use these wrappers)


froddd (approved):
👍

PR #2251, opened 2024-12-16
APS-1618 Licence expiry date

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/APS-1618

# Changes in this PR

This change makes `licenceExpiryDate` a first-class field in the `approved_premises_applications` table (`ApprovedPremisesApplication` object )
Note that the object type `SubmitApprovedPremisesApplication` uses the American spelling of `licenseExpiryDate` while everywhere else uses the English 'licence' spelling.
The field value is read from the application object and passed on application submission. The field is only populated if the user both populates the licence expiry date and selects it as a relevant date.

## Screenshots of UI changes
No UI change


froddd (approved):
LGTM 👍

PR #2256, opened 2024-12-18
APS-1713 - Remove incorrect use of out of service bed permission

davidatkinsuk:
This commit removes usage of `cas1_out_of_service_bed_create` permisson to determine if a user can view a bed, replacing it with the new `cas1_premises_view` permission which is assigned to CRU Members and Future Managers

I've tested this by ensuring a future manager can still view a bed


froddd (approved):
LGTM 👍

PR #481, opened 2023-01-12
(873) Confirm if theres enough information in assessment

pezholio:
This adds a new step to confirm if there is enough information in the assessment, as well as displaying the applicant's information, so the reviewer knows who to contact if they need more information. At the moment, this only tests the happy path. The next PR will implement what happens if the user selects "No"

## Screenshots

image

image

PR #2259, opened 2024-12-19
Moving away from csurf and to csrf-sync

aliuk2012:
# Related
- CAS-2: https://github.com/ministryofjustice/hmpps-community-accommodation-tier-2-ui/pull/840
- CAS-3: https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/1179

# Context
Fixes a security vulnerability https://github.com/advisories/GHSA-pxg6-pf52-xh8x and brings our UI more inline with HMPPS Typescript template

> csurf has been deprecated for some time and this removes that dependency and implements the synchronizer token pattern using csrf-sync.
>
> Note: Previously csurf used to generate new tokens on every request. The new library generates tokens once per session which is preferrable due to the extra calls to redis that per-request would generate. It is possible to force a refresh/revocation of a token by explicitly calling: req.csrfToken(true)

See PR #481

Quote from https://github.com/ministryofjustice/hmpps-template-typescript/blob/2e139f5e9e3a6dbfaa0c6bbbc240513556102a8f/CHANGELOG.md 

PR: https://github.com/ministryofjustice/hmpps-template-typescript/pull/481/files


froddd (approved):
LGTM

PR #2263, opened 2024-12-20
Occupancy View - add license expiry date to matching details section

gregkhawkins:
# Context

* This relates to this ticket: https://dsdmoj.atlassian.net/browse/APS-1624
* We previously de-scoped including the `Licence expiry date` data from the `Matching details` section on the `Occupancy View` page because the data was not available from the API
* The `Licence expiry date` data has since been made available from the `GET /placement-requests/{id}` API endpoint

# Changes in this PR
- Add the `Licence expiry date` data into the `Matching details` section on the `Occupancy View`

### Before
before

### After
after

### DIFF
--- a/integration_tests/pages/match/occupancyViewPage.ts
+++ b/integration_tests/pages/match/occupancyViewPage.ts
@@ -3,7 +3,6 @@ import {
   Cas1PremiseCapacity,
   Cas1PremisesSummary,
   Cas1SpaceBookingCharacteristic,
-  PlacementRequest,
   PlacementRequestDetail,
 } from '@approved-premises/api'
 import Page from '../page'
@@ -32,7 +31,7 @@ export default class OccupancyViewPage extends Page {
     totalCapacity: number,
     startDate: string,
     durationDays: number,
-    placementRequest: PlacementRequest,
+    placementRequest: PlacementRequestDetail,
     managerDetails: string,
   ) {
     cy.get('.govuk-details').within(() => {
--- a/server/utils/match/index.ts
+++ b/server/utils/match/index.ts
@@ -1,6 +1,7 @@
 import { addDays } from 'date-fns'
 import type {
   ApType,
+  ApprovedPremisesApplication,
   Cas1SpaceCharacteristic,
   Gender,
   PlacementCriteria,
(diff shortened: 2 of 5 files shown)


server/utils/match/index.ts
```diff
@@ -174,17 +175,19 @@ export const spaceBookingPersonNeedsSummaryCardRows = (
 
 export const occupancyViewSummaryListForMatchingDetails = (
   totalCapacity: number,
-  placementRequest: PlacementRequest,
+  placementRequest: PlacementRequestDetail,
   managerDetails: string,
 ): Array<SummaryListItem> => {
   const placementRequestDates = placementDates(placementRequest.expectedArrival, placementRequest.duration)
   const essentialCharacteristics = filterOutAPTypes(placementRequest.essentialCriteria)
+  const application = placementRequest.application as ApprovedPremisesApplication
```
bobmeredith, line 183: Does this need to check the `type`, or can we guarantee that this `Application` is always an `ApprovedPremisesApplication`

bobmeredith (commented):
Looks good.
A couple of comments.

Has there been an API change to return the whole application along with a placement request?  If so, that doesn't sound like a good thing from a performance POV. The page controller could always get the application in a separate call if needed.

integration_tests/pages/match/occupancyViewPage.ts
```diff
@@ -32,7 +31,7 @@ export default class OccupancyViewPage extends Page {
     totalCapacity: number,
     startDate: string,
     durationDays: number,
-    placementRequest: PlacementRequest,
+    placementRequest: PlacementRequestDetail,
```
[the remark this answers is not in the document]
gregkhawkins: thanks for looking at this @bobmeredith was half expecting this to be an after xmas review. Appreciate it :)

So, in this test code, this function goes on to call the `occupancyViewSummaryListForMatchingDetails()` function to assert some expectations. To achieve this it needs to pass the correct type - which has changed because of the data-type change late on in the PR:
https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/2263/files#diff-cc8326f12e1aed4a0d54a6c3199c6dac2af0450fec3103000188aa31f6df42cb

The justification for that data type change is that, in actual fact, we have always been passing an object of type `PlacementRequestDetail` into that function but the data type was set incorrectly to it's parent type (i.e. `PlacementRequest`). It wasn't flagging issues (as we were passing a child of that parent) but the change is now needed so I can access the application from the `PlacementRequestDetail` (so i can go on to grab the `licenceExpiryDate` later).

Furthermore, in the occupancy view controller, we make the below call:

placement-request-detail-call

As you can see above, from the IDE hint, what we get back has always been of type `PlacementRequestDetail` (not `PlacementRequest`). And back to the test:
here

As you can see we were always stubbing `PlacementRequestDetail` (not the parent `PlacementRequestDetail`) which is why there are no side effects of this data type change also

server/utils/match/index.ts
gregkhawkins: I had a quick look into the BE code and saw that when we hit the `GET /placement-requests/{id}` endpoint, it essentially goes off and build the placement request including the application. To build the correct application type it executes this code:
Screenshot 2024-12-20 at 15 00 23 (2)

So, the `is` stuff above determines which application type is attached to this placement request and builds it up accordingly. My understanding is that in this area (i.e. `Occupancy view`) the placements requests will all be of type `ApprovedPremisesApplicationEntity` as this `Occupancy view` area is functionality used for CAS1 applications. So, my conclusion is that, yes we can guarantee that this `Application` is always an `ApprovedPremisesApplication`.

Let me know if I'm missing anything :)

gregkhawkins:
Thanks gain for the review! 
Re `Has there been an API change to return the whole application along with a placement request?` As touched on above the BE has been returning the whole application for a while from what I hve seen. Perhaps this is one to raise as a refactor if we want to split into two calls? Looking at the BE git history it has been returning the application for a while

bobmeredith (approved):
OK, well in that case LGTM.
I do think we should look to refactor as large unnecessary payloads are not a good thing.

PR #2264, opened 2024-12-20
APS-1677 AP occupancy view

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/APS-1677

# Changes in this PR

Adds the occupancy view (calendar) page with a link from the premises view action menu.

## Screenshots of UI changes

image

image

### DIFF
--- a/integration_tests/tests/manage/occupancyView.cy.ts
+++ b/integration_tests/tests/manage/occupancyView.cy.ts
@@ -1,0 +1,161 @@
+import { addDays } from 'date-fns'
+import {
+  cas1PremiseCapacityFactory,
+  cas1PremisesSummaryFactory,
+  cas1SpaceBookingSummaryFactory,
+  staffMemberFactory,
+} from '../../../server/testutils/factories'
+
+import { OccupancyViewPage, PremisesShowPage } from '../../pages/manage'
+
+import { signIn } from '../signIn'
+import { DateFormats } from '../../../server/utils/dateUtils'
+import Page from '../../pages/page'
+
+context('Premises occupancy', () => {
+  describe('show', () => {
+    const startDateObj = new Date()
+    const endDateObj = new Date(startDateObj)
+    endDateObj.setDate(endDateObj.getDate() + 84)
+
+    const startDate = DateFormats.dateObjToIsoDate(startDateObj)
+    const endDate = DateFormats.dateObjToIsoDate(endDateObj)
+    const premisesCapacity = cas1PremiseCapacityFactory.build({ startDate, endDate })
+
+    const premises = premisesCapacity.premise
+    const placements = cas1SpaceBookingSummaryFactory.buildList(30)
+    const keyworkers = staffMemberFactory.keyworker().buildList(5)
+
+    beforeEach(() => {
+      cy.task('reset')
+
+      // Given there is a premises in the database
+      cy.task('stubSinglePremises', premises)
+      cy.task('stubPremisesStaffMembers', { premisesId: premises.id, staffMembers: keyworkers })
+
+      // And it has a list of upcoming placements
+      cy.task('stubSpaceBookingSummaryList', { premisesId: premises.id, placements, residency: 'upcoming' })
+      cy.task('stubPremiseCapacity', { premisesId: premises.id, startDate, endDate, premiseCapacity: premisesCapacity })
+    })
+
+    describe('with premises view permission', () => {
+      beforeEach(() => {
+        // Given I am logged in as a future manager with premises_view permission
+        signIn(['future_manager'], ['cas1_premises_view'])
+      })
+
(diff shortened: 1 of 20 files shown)


server/testutils/factories/cas1PremisesSummary.ts
```diff
@@ -6,7 +6,7 @@ import { apAreaFactory } from './referenceData'
 
 export default Factory.define<Cas1PremisesSummary>(() => ({
   id: faker.string.uuid(),
-  name: `${faker.person}`,
```
bobmeredith, line 9: I noticed a bug here - the name of the premises was appearing as `[object:object]`

server/views/partials/_calendar.njk
```diff
@@ -0,0 +1,25 @@
+{% macro occupancyCalendar(calendar) %}
+    <div class="calendar">
+        {% for month in calendar %}
+            <section>
+                <h3 class="govuk-heading-m">{{ month.name }}</h3>
+
+                <ul class="calendar__month govuk-list">
+                    {% for day in month.days %}
+                        <li class="calendar__day calendar__day--{{ day.name.slice(0,3) | lower }}{% if day.status !== 'available' %} {{ 'govuk-tag--yellow' if day.status in ['availableForCriteria','full'] else 'govuk-tag--red' }}{% endif %}">
+                            <a class="calendar__link" href="{% if day.link %}{{ day.link }}{% else %}#{% endif %}">
```
bobmeredith, line 10: So this assumes that all calendars are rendered with a `link` property in their day structures. I could put this in the lower level day templates, but that would mean more of the day rendering, including the date name, would be duplicated.

server/testutils/factories/cas1PremisesSummary.ts
aliuk2012: Good catch, also weird to be using the `faker.person` for a premises summary factory

aliuk2012 (approved):
Looks good to me and seems sensible.

Only other nit picking suggestion I could make is to maybe squash the commits?

server/utils/premises/premisesActions.ts
```diff
@@ -21,5 +21,13 @@ export const premisesActions = (user: UserDetails, premises: Premises) => {
     })
   }
 
+  if (user.permissions.includes('cas1_premises_view') && premises.supportsSpaceBookings) {
```
bobmeredith, line 24: Changed to an explicit test for permission - but this permission is already given to users in prod.
The `supportsSpaceBookings` flag however is not set on any premises ATM, and will be set on rollout in February.

server/views/manage/premises/occupancy/view.njk
```diff
@@ -0,0 +1,96 @@
...
+{% from "govuk/components/input/macro.njk" import govukInput %}
+{% from "govuk/components/date-input/macro.njk" import govukDateInput %}
+{% from "../../../partials/showErrorSummary.njk" import showErrorSummary %}
+{% from "./_calendar.njk" import occupancyCalendar %}
+
+{%- from "moj/components/identity-bar/macro.njk" import mojIdentityBar -%}
+
+{% extends "../../../partials/layout.njk" %}
+
+{% set pageTitle = applicationName + " - " + premises.name %}
```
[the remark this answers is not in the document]
bobmeredith: Good point - will do.

server/views/manage/premises/occupancy/view.njk
```diff
@@ -0,0 +1,96 @@
...
+{% endblock %}
+
+{% set titleHtml %}
+
+{% endset %}
+
+{% block content %}
+    {{ showErrorSummary(errorSummary) }}
+    <span class="govuk-caption-l">{{ premises.name }}</span>
+    <h1 class="govuk-heading-l">View spaces in {{ premises.name }}</h1>
```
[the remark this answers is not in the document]
bobmeredith: Just following the design guv!
I've queried it will the designers so we can fix in a later revision.

integration_tests/tests/manage/occupancyView.cy.ts
```diff
@@ -0,0 +1,152 @@
...
+        cy.get('.calendar').should('not.exist')
+        // When I select a different start date and click submit
+        page.clearAndCompleteDateInputs('startDate', newStartDate)
+        page.clickSubmit()
+        // Then the new start date should be shown
+        page.shouldHaveSelectText('durationDays', '12 weeks')
+        page.dateInputsShouldContainDate('startDate', newStartDate)
+      })
+
+      it('should not be available in the premises actions menu if the premises does not support space bookings', () => {
```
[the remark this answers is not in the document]
bobmeredith: Actually, this test used to be outside the permission block, but now, the permission is the same, so it didn't make sense.
Sadly, I can't test for the permission on the action option as there is a permission check on the page it sits on, meaning I can't render the page without having the permission.

PR #2268, opened 2024-12-24
APS-1705 Default key-worker to 'Not assigned' on the premises view page placement list

bobmeredith:
# Context
https://dsdmoj.atlassian.net/browse/APS-1705

# Changes in this PR
For placements where a key-worker is not assigned, show 'Not assigned' in the table rather than blank

### Before
image

### After
image

### DIFF
--- a/server/utils/premises/index.test.ts
+++ b/server/utils/premises/index.test.ts
@@ -89,12 +89,12 @@ describe('premisesUtils', () => {
       ])
     })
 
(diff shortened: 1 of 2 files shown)


RobBoothMOJ (approved):
Approved

aliuk2012:
I noticed the column can be sorted, Does the sorting still work after setting blank rows to `Not Assigned`

PR #2271, opened 2024-12-30
Placement request page - add blue banner

gregkhawkins:
# Context

JIRA: https://dsdmoj.atlassian.net/browse/APS-1690

# Changes in this PR

- On the `Placement request` page, add the blue banner at the top (inc. key-person details)
- Check that this is included in the integration tests

### Before
before

### After
after

### DIFF
--- a/integration_tests/pages/match/searchPage.ts
+++ b/integration_tests/pages/match/searchPage.ts
@@ -3,8 +3,6 @@ import { Cas1SpaceSearchResult, Cas1SpaceSearchResults, PlacementRequestDetail }
 import Page from '../page'
 import { placementRequestSummaryListForMatching, summaryCardRows } from '../../../server/utils/match'
 import paths from '../../../server/paths/match'
-import { isFullPerson } from '../../../server/utils/personUtils'
-import { DateFormats } from '../../../server/utils/dateUtils'
 import { occupancyCriteriaMap } from '../../../server/utils/match/occupancy'
 
 export default class SearchPage extends Page {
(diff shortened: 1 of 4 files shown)


server/views/admin/placementRequests/show.njk
```diff
@@ -2,10 +2,16 @@
...
 {% from "govuk/components/notification-banner/macro.njk" import govukNotificationBanner %}
 
+{% from "../../components/keyDetails/macro.njk" import keyDetails %}
 {% extends "../../partials/layout.njk" %}
 
 {% set pageTitle = applicationName + " - View Placement Request" %}
 
+{% block header %}
+    {{ super() }}
+    {{ keyDetails(MatchUtils.keyDetails(placementRequest)) }}
```
aliuk2012, line 12: No strictly related to the changes you made, in CAS-3 we have a very similiar component to `keyDetails` called `placeContextHeader` (not a big fan of the name 🤷 ) so wondered `keyDetails` was implemented.  Not sure if its worth doing as part of these changes but I think the markup in `keyDetails` could be improved to be more semantically structured to help screen reader users. Also I'm not sure how well this particular implementation works with mobil but I know before we refactored placeContextHeader it looked terrible on mobil or zoomed in.

This is the PR we did when we refactored the component https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/1026 and here is the component itself https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/blob/main/server/views/components/place-context-header/macro.njk.  For mobile/zoomed in view we change the layout of the `dl` items so they are column layout instead inline

aliuk2012:
On CAS-3 when we rolled out the blue banner, I raised the concerns of `banner blindness` which I think is also relevant for CAS-1 designs

image

After speaking with @devansXD and Mahesh, we agreed to move the phase banner to the bottom of the page just above the footer.

Example of desktop and mobile 
image
image

integration_tests/pages/match/searchPage.ts
```diff
@@ -22,18 +20,6 @@ export default class SearchPage extends Page {
...
-    cy.get('.prisoner-info').within(() => {
-      const { person } = placementRequest
-      if (!isFullPerson(person)) throw new Error('test requires a Full Person')
-
-      cy.get('span').contains(person.name)
-      cy.get('span').contains(`CRN: ${person.crn}`)
-      cy.get('span').contains(`Tier: ${placementRequest?.risks?.tier?.value.level}`)
-      cy.get('span').contains(`Date of birth: ${DateFormats.isoDateToUIDate(person.dateOfBirth, { format: 'short' })}`)
-    })
-  }
```
aliuk2012, line 35: Does the search page not contain this component anymore? I didn't see any changes to the search page only to a show page

froddd: This looks like it has been moved to the ancestor class, `Page`, as it is quite generic and used in several places. Makes sense!

server/views/admin/placementRequests/show.njk
froddd: Yeah, the markup for this definitely needs looking at -- I'll create a ticket and will most likely use the same approach as CAS3, but we'll do this separately.

froddd (approved):
LGTM!

PR #2275, opened 2025-01-07
APS-1751 Criteria transfer to space booking

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/APS-1751

# Changes in this PR

When a user creates a space booking from the occupancy view page, the essential criteria in the booking should be populated with the criteria that have been set as filters on the occupancy view page.
These filters are defaulted to the filters that are set on the suitability search page, which in turn are defaulted to the matching criteria on the assessment.
The selected criteria are passed into the booking confirmation page as a set of query parameters.

## Screenshots of UI changes

image

image

### DIFF
--- a/integration_tests/pages/match/bookASpacePage.ts
+++ b/integration_tests/pages/match/bookASpacePage.ts
@@ -1,13 +1,16 @@
-import { ApType, PlacementDates, PlacementRequestDetail, Premises } from '@approved-premises/api'
+import {
+  ApType,
+  Cas1SpaceBookingCharacteristic,
+  Cas1SpaceCharacteristic,
+  PlacementDates,
+  PlacementRequestDetail,
+  Premises,
+} from '@approved-premises/api'
 import Page from '../page'
 import paths from '../../../server/paths/match'
 import { createQueryString, sentenceCase } from '../../../server/utils/utils'
 import { DateFormats } from '../../../server/utils/dateUtils'
-import {
-  filterOutAPTypes,
-  placementDates,
-  placementLength as placementLengthInDaysAndWeeks,
-} from '../../../server/utils/match'
+import { placementDates, placementLength as placementLengthInDaysAndWeeks } from '../../../server/utils/match'
 import { placementCriteriaLabels } from '../../../server/utils/placementCriteriaUtils'
 import { apTypeLabels } from '../../../server/utils/apTypeLabels'
 
@@ -23,8 +26,9 @@ export default class BookASpacePage extends Page {
     premisesName: Premises['name'],
     premisesId: Premises['id'],
     apType: ApType,
+    criteria: Array<Cas1SpaceBookingCharacteristic>,
   ) {
-    const queryString = createQueryString({ startDate, durationDays, premisesName, premisesId, apType })
+    const queryString = createQueryString({ startDate, durationDays, premisesName, premisesId, apType, criteria })
     const path = `${paths.v2Match.placementRequests.spaceBookings.new({ id: placementRequest.id })}?${queryString}`
     cy.visit(path)
     return new BookASpacePage(premisesName)
@@ -35,6 +39,7 @@ export default class BookASpacePage extends Page {
     startDate: string,
     duration: PlacementDates['duration'],
     apType: ApType,
+    criteria?: Array<Cas1SpaceCharacteristic>,
   ): void {
     const { endDate, placementLength } = placementDates(startDate, duration.toString())
     cy.get('dd').contains(apTypeLabels[apType])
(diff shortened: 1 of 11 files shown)


integration_tests/pages/match/bookASpacePage.ts
```diff
@@ -1,13 +1,9 @@
-import { ApType, PlacementDates, PlacementRequestDetail, Premises } from '@approved-premises/api'
+import { ApType, PlacementCriteria, PlacementDates, PlacementRequestDetail, Premises } from '@approved-premises/api'
```
froddd, line 1: I we want to be using `Cas1SpaceBookingCharacteristic` here -- there's a slight disconnect between `PlacementRequest` using `PlacementCriteria`, but from what I can see the types seem to have enough overlap that we should be able to migrate to the new one bit by bit. We may be able to update the API to start using those instead?

froddd: For reference, the `criteria` query parameter on the occupancy view already uses the new type: https://github.com/ministryofjustice/hmpps-approved-premises-ui/blob/main/server/controllers/match/placementRequests/occupancyViewController.ts#L33

server/utils/match/index.ts
```diff
@@ -121,13 +121,15 @@ export const redirectToSpaceBookingsNew = ({
...
   durationDays,
+  criteria,
 }: {
   placementRequestId: string
   premisesName: string
   premisesId: string
   apType: string
   startDate: string
   durationDays: string
+  criteria: Array<Cas1SpaceCharacteristic>
```
froddd, line 132: Should this be `Cas1SpaceBookingCharacteristic`?

server/utils/match/index.test.ts
```diff
@@ -330,6 +331,7 @@ describe('matchUtils', () => {
         apType,
         startDate,
         durationDays,
+        criteria,
       })
 
       expect(result).toEqual(
```
froddd, line 337: This assertion is a bit close to the implementation itself. Could the assert be a bit more explicit, by hradcoding the expectations?

server/controllers/match/placementRequests/occupancyViewController.test.ts
```diff
@@ -301,7 +301,7 @@ describe('OccupancyViewController', () => {
 
       const expectedDurationDays = 10
       const expectedStartDate = `${arrivalYear}-0${arrivalMonth}-${arrivalDay}`
-      const expectedParams = `apType=${apType}&startDate=${expectedStartDate}&durationDays=${expectedDurationDays}`
+      const expectedParams = `apType=${apType}&startDate=${expectedStartDate}&durationDays=${expectedDurationDays}&criteria=`
```
froddd, line 304: If `criteria` is an empty string we shouldn't need to include this parameter at all

server/controllers/match/placementRequests/spaceBookingsController.test.ts
```diff
@@ -72,7 +76,8 @@ describe('SpaceBookingsController', () => {
         errorSummary: {},
         errors: [],
         dates: placementDates(startDate, durationDays),
-        essentialCharacteristics: filterOutAPTypes(placementRequestDetail.essentialCriteria),
+        // essentialCharacteristics: filterOutAPTypes(placementRequestDetail.essentialCriteria),
```
froddd, line 79: This commented out code can be removed

server/controllers/match/placementRequests/spaceBookingsController.ts
```diff
@@ -1,5 +1,5 @@
 import type { Request, RequestHandler, Response, TypedRequestHandler } from 'express'
-import type { ApType, Cas1NewSpaceBooking } from '@approved-premises/api'
+import type { ApType, Cas1NewSpaceBooking, PlacementCriteria } from '@approved-premises/api'
```
froddd, line 2: Again, this should really reference `CasSpaceBookingCharacteristic` rather than `PlacementCriteria`. This might require `filterOutAPTypes` to have an update signature type, but it would make sense from here.

server/controllers/match/placementRequests/spaceBookingsController.test.ts
```diff
@@ -72,7 +76,8 @@ describe('SpaceBookingsController', () => {
...
+        essentialCharacteristics: filterOutAPTypes(criteria),
```
froddd, line 80: Not sure we should need to filter out AP types, if we know `criteria` is already of the `Cas1SpaceBookingCharacteristic` subset

server/controllers/match/placementRequests/spaceBookingsController.ts
bobmeredith: True.  I added a fn to filter down from `PlacementCriteria` to `Cas1SpaceBookingCharacteristic` as the existing `filterOutApTypes` returns a  `Cas1SpaceCharacteristic` which contains more characteristics than we want.

server/controllers/match/placementRequests/occupancyViewController.test.ts
bobmeredith: True. It needs a tweak to the controller but a good shout

server/controllers/match/placementRequests/spaceBookingsController.test.ts
bobmeredith: Yes, the type conversion is the filter so it doesn't need doing twice. Again, I switched for `filterToSpaceBookingCharacteristics` to reduce to only the set of criteria we want to save in the space booking.

integration_tests/tests/match/match.cy.ts
```diff
@@ -78,7 +78,7 @@ context('Placement Requests', () => {
     // And the results should have links with the correct AP type and criteria
     searchPage.shouldHaveSearchParametersInLinks(newSearchParameters)
 
-    // And the parameters should be submitted to the API
+    // // And the parameters should be submitted to the API
```
froddd, line 81: Comment/uncomment operation artefact 🤓

froddd (approved):
All good!

PR #2277, opened 2025-01-09
APS-1775: Update booking confirmation screen

froddd:
# Context

https://dsdmoj.atlassian.net/browse/APS-1775

# Changes in this PR

Implements design and content updates to the 'Confirm booking' page.

Under the hood, this ensures the booking flow relies on the arrival and departure dates entered on the Occupancy view page, rather than recalculating those based on a start date and duration. The path for the confirmation screen has been rewritten to include the premises ID, which ensures a 404 is returned if this ID is invalid/doesn't exist.

The occupancy filter query parameters are carried over to the confirmation page to ensure the back link functions correctly.

The `apType` parameter is now removed, as it is no longer used beyond the initial suitability search.

### Before

Screenshot 2025-01-09 at 15 32 18

### After

Screenshot 2025-01-09 at 15 29 25

### DIFF
--- a/integration_tests/pages/match/bookASpacePage.ts
+++ b/integration_tests/pages/match/bookASpacePage.ts
@@ -1,54 +1,53 @@
-import {
-  ApType,
-  Cas1SpaceBookingCharacteristic,
-  Cas1SpaceCharacteristic,
-  PlacementDates,
-  PlacementRequestDetail,
-  Premises,
-} from '@approved-premises/api'
+import { Cas1Premises, Cas1SpaceBookingCharacteristic, PlacementRequestDetail, Premises } from '@approved-premises/api'
+import { differenceInDays } from 'date-fns'
 import Page from '../page'
 import paths from '../../../server/paths/match'
-import { createQueryString, sentenceCase } from '../../../server/utils/utils'
-import { DateFormats } from '../../../server/utils/dateUtils'
-import { placementDates, placementLength as placementLengthInDaysAndWeeks } from '../../../server/utils/match'
-import { placementCriteriaLabels } from '../../../server/utils/placementCriteriaUtils'
-import { apTypeLabels } from '../../../server/utils/apTypeLabels'
+import { createQueryString } from '../../../server/utils/utils'
+import { DateFormats, daysToWeeksAndDays } from '../../../server/utils/dateUtils'
+import { requirementsHtmlString } from '../../../server/utils/match'
+import { allReleaseTypes } from '../../../server/utils/applications/releaseTypeUtils'
 
 export default class BookASpacePage extends Page {
-  constructor(premisesName: string) {
-    super(`Book space in ${premisesName}`)
+  constructor() {
+    super(`Confirm booking`)
   }
 
   static visit(
     placementRequest: PlacementRequestDetail,
-    startDate: string,
-    durationDays: PlacementDates['duration'],
-    premisesName: Premises['name'],
     premisesId: Premises['id'],
-    apType: ApType,
+    arrivalDate: string,
+    departureDate: string,
     criteria: Array<Cas1SpaceBookingCharacteristic>,
   ) {
-    const queryString = createQueryString({ startDate, durationDays, premisesName, premisesId, apType, criteria })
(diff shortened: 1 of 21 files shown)


server/utils/match/index.ts
```diff
@@ -195,6 +190,37 @@ export const occupancyViewSummaryListForMatchingDetails = (
...
+export const spaceBookingConfirmationSummaryListRows = (
+  placementRequest: PlacementRequestDetail,
+  premises: Cas1PremisesSummary,
+  arrivalDate: string,
+  departureDate: string,
+  criteria: Array<Cas1SpaceBookingCharacteristic>,
+): Array<SummaryListItem> => {
+  return [
+    textRow('Approved Premises', premises.name),
+    // textRow('Address', premises.fullAddress),
```
bobmeredith, line 212: could remove commented line

server/utils/match/index.ts
```diff
@@ -195,6 +190,37 @@ export const occupancyViewSummaryListForMatchingDetails = (
   ]
 }
 
+export const textRow = (label: string, value: string): SummaryListItem => ({
```
bobmeredith, line 193: These utils look like they could be used widely. there is a similar fn in `utils/placements/index.ts` for instance.
Perhaps it would be good to put these somewhere global - `utils/formUtils` say.

server/controllers/match/placementRequests/spaceBookingsController.test.ts
```diff
@@ -1,98 +1,103 @@
...
     jest.resetAllMocks()
-    spaceBookingsController = new SpaceBookingsController(placementRequestService, spaceService)
+    spaceBookingsController = new SpaceBookingsController(placementRequestService, premisesService, spaceService)
+    request = createMock<Request>({ user: { token } })
+
+    placementRequestService.getPlacementRequest.mockResolvedValue(placementRequestDetail)
+    premisesService.find.mockResolvedValue(premises)
+
+    jest.spyOn(validationUtils, 'fetchErrorsAndUserInput')
+    when(validationUtils.fetchErrorsAndUserInput as jest.Mock)
```
bobmeredith, line 47: Not sure if I'm missing something here. Isn't the idea of jest-when to allow a mocked return based on the received values?
As this there is only one when, and it returns a constant - I don't see the difference from just using `<mock>.mockReturnValue()`

bobmeredith (approved):
Nice one!
It's much better using the post body to pass those parameters and keeping the filter date persist in the query string.
Just a few minor comments and queries.

integration_tests/pages/match/bookASpacePage.ts
```diff
@@ -1,54 +1,58 @@
...
-    cy.get('dd').contains(DateFormats.isoDateToUIDate(startDate))
-    cy.get('dd').contains(DateFormats.isoDateToUIDate(endDate))
-    cy.get('dd').contains(placementLengthInDaysAndWeeks(placementLength))
-    cy.get('dd').contains(sentenceCase(placementRequest.gender))
-    ;(criteria || []).forEach(requirement => {
-      cy.get('li').contains(placementCriteriaLabels[requirement])
-    })
+    this.shouldContainSummaryListItems([
+      { key: { text: 'Approved Premises' }, value: { text: premises.name } },
+      // { key: { text: 'Address' }, value: { text: premises.fullAddress } },
```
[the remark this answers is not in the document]
froddd: Address is now available from API and showing, so this is uncommented

server/utils/match/index.ts
froddd: Address is now available from API and showing, so this is uncommented

server/controllers/match/placementRequests/spaceBookingsController.test.ts
froddd: We shouldn't actually need to mock `fetchErrorsAndUserInput` -- we can simply provide a mocked `flash()` method to the existing request mock, which would then return the default values for the above.

server/utils/match/index.ts
froddd: I've refactored the above a bit but not gone on to replace all instances -- this would seem like a sizeable change to add to this. Thoughts?

PR #2279, opened 2025-01-09
CHORE: Remove unused code referencing bed search

froddd:
This removes unused code that still references 'bed search', which hasn't been used in a long time, and is now effectively replaced by 'space search'. 

After removing unused implementations, the only leftover was the `BedSearchController`, which should have been renamed `SpaceSearchController` at the same time as the file itself -- now fixed to prevent any further confusion.


server/paths/match.ts
```diff
@@ -32,10 +31,9 @@ export default {
     index: placementRequestsPath,
     show: placementRequestPath,
     create: newPlacementRequestPath,
-    bookingNotMade: { confirm: bookingNotMadePath.path('confirm'), create: bookingNotMadePath },
-    bookings: {
-      confirm: placementRequestBookingsPath.path('confirm'),
-      create: placementRequestBookingsPath,
+    bookingNotMade: {
```
bobmeredith, line 34: Could we maybe change to `placementNotMade` for consistency

bobmeredith (approved):
Looks good.
Just one comment.

server/paths/match.ts
froddd: This could do with a small refactor -- the 'Unable to match' flow uses the old implementation, which seems to work fine, but could do with moving over to the newer `spaceBookingsController`!

PR #2280, opened 2025-01-10
Refactor placement-request page information

gregkhawkins:
# Context

* Related JIRA: https://dsdmoj.atlassian.net/browse/APS-1693
* This PR makes changes to the Summary information on the `Placement request` page in the `Find and Book` flow. This includes removing the data that overlaps with thre blue banner and moving the rest of the data to a summary list. 

# Changes in this PR

### Before
placement_request_before

### After
placement_request_after

### DIFF
--- a/server/utils/match/index.test.ts
+++ b/server/utils/match/index.test.ts
@@ -1,5 +1,6 @@
 import type {
   ApType,
+  Application,
   ApprovedPremisesApplication,
   Cas1SpaceBookingCharacteristic,
   FullPerson,
@@ -7,6 +8,7 @@ import type {
 } from '@approved-premises/api'
 import { when } from 'jest-when'
 import type { SummaryListItem } from '@approved-premises/ui'
+import applyPaths from '../../paths/apply'
 import paths from '../../paths/match'
 import {
   cas1PremisesFactory,
@@ -23,6 +25,7 @@ import {
   apManagerDetailsRow,
   apTypeLabelsForRadioInput,
   apTypeRow,
+  apTypeWithViewTimelineActionRow,
   arrivalDateRow,
   calculateDepartureDate,
   characteristicsRow,
@@ -47,10 +50,11 @@ import {
   placementLength,
   placementLengthRow,
   placementRequestSummaryListForMatching,
-  postcodeRow,
+  preferredPostcodeRow,
   premisesNameRow,
   redirectToSpaceBookingsNew,
   releaseTypeRow,
+  requestedOrEstimatedArrivalDateRow,
   requirementsHtmlString,
   spaceBookingConfirmationSummaryListRows,
   spaceBookingPersonNeedsSummaryCardRows,
@@ -69,6 +73,7 @@ import { textValue } from '../applications/helpers'
 import { preferredApsRow } from '../placementRequests/preferredApsRow'
 import { placementRequirementsRow } from '../placementRequests/placementRequirementsRow'
 import applicationFactory from '../../testutils/factories/application'
+import offlineApplicationFactory from '../../testutils/factories/offlineApplication'
 import { allReleaseTypes } from '../applications/releaseTypeUtils'
 
 jest.mock('../retrieveQuestionResponseFromFormArtifact')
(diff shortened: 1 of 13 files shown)


server/utils/match/index.ts
```diff
@@ -376,14 +409,21 @@ export const releaseTypeRow = (placementRequest: PlacementRequest) => ({
...
-  value: {
-    text: application.licenceExpiryDate ? DateFormats.isoDateToUIDate(application.licenceExpiryDate) : '',
-  },
-})
+export const licenceExpiryDateRow = (placementRequest: PlacementRequestDetail) => {
+  let licenceExpiryDate: string | undefined
+  if (placementRequest.application && 'licenceExpiryDate' in placementRequest.application) {
+    const application = placementRequest.application as ApprovedPremisesApplication
+    licenceExpiryDate = application.licenceExpiryDate
+  }
```
gregkhawkins, line 417: the integration tests saved my neck here as the `isParole` case tested in int tests does not have an application associated with the placementRequest. I assume the data i ssetup correctly for that test to reflect the parole case. Anyway, this means we need the `if (placementRequest.application)` for sure or runtime exception.
I added the `if ('licenceExpiryDate' in placementRequest.application)` check as I was growing increasingly paranoid that a `placementRequest.application` would not be of type `ApprovedPremisesApplication`. I did check this in the backed previously (when Bob flagged it on previous PR) but with recent work on offline applications etc I wondered whether we could end up with an application that does not meet this type leading to a runtime exception related to that.
Let me know if this is too defensive or if you can think of a more graceful way to do it

server/utils/match/index.ts
```diff
@@ -376,14 +409,21 @@ export const releaseTypeRow = (placementRequest: PlacementRequest) => ({
...
-  key: {
-    text: 'Licence expiry date',
-  },
-  value: {
-    text: application.licenceExpiryDate ? DateFormats.isoDateToUIDate(application.licenceExpiryDate) : '',
-  },
-})
+export const licenceExpiryDateRow = (placementRequest: PlacementRequestDetail) => {
+  let licenceExpiryDate: string | undefined
+  if (placementRequest.application && 'licenceExpiryDate' in placementRequest.application) {
```
gregkhawkins, line 414: the integration tests saved my neck here as the `isParole` case tested in int tests does not have an application associated with the placementRequest. I assume the data i ssetup correctly for that test to reflect the parole case. Anyway, this means we need the `if (placementRequest.application)` for sure or runtime exception.
I added the `if ('licenceExpiryDate' in placementRequest.application)` check as I was growing increasingly paranoid that a `placementRequest.application` would not be of type `ApprovedPremisesApplication`. I did check this in the backed previously (when Bob flagged it on previous PR) but with recent work on offline applications etc I wondered whether we could end up with an application that does not meet this type leading to a runtime exception related to that.
Let me know if this is too defensive or if you can think of a more graceful way to do it

server/utils/match/index.test.ts
```diff
@@ -430,6 +507,27 @@ describe('matchUtils', () => {
...
+      const placementRequestWithoutLicenceExpiry = {
+        ...placementRequest,
+        application: undefinedApplication,
+      }
+      expect(
+        occupancyViewSummaryListForMatchingDetails(totalCapacity, placementRequestWithoutLicenceExpiry, managerDetails),
+      ).toEqual(expectedMatchingDetailsSummaryListItems(''))
+    })
+
+    it(`should generate the expected matching details when placement-request's application is not of type ApprovedPremisesApplication`, () => {
```
gregkhawkins, line 521: justification for this test case is in a comment in the `index.ts` file of this PR

bobmeredith (approved):
Nice one!
Looks good to me.
Just a couple a minor comments (one is a tiny query on the content/design).

server/utils/match/index.ts
gregkhawkins: I guess the real question here is how confident are we that the `placementRequest.application` will always be set and always be of type `ApprovedPremisesApplication`? If very confident we cn get rid of the defensive code and always grab the `ApprovedPremisesApplication.licenseExpiryDate`.
This function is used currently in the `Find and Book` journey by the `Occupancy View page` and as of this PR the `Placement Request page` now also. It will also (in an upcoming feature) be used in the `Find a space` page also.
If we are confident that in all of these pages it will always be `ApprovedPremisesApplication` and never undefined then I can remove the defensive if statement fro sure :)
It's possible that the test data was not set up properly for the `placementRequest.application` in the integration tests around this page which lead to the runtime exception which lead to the wrong conclusion by me. If this is the case I will just setup the test data properly (to include the `placementRequest.application`) instead of including this defensive code.
Let me know the preferred approach and thanks :)

server/utils/placementRequests/placementRequestSummaryList.test.ts
```diff
@@ -0,0 +1,214 @@
...
+    expectedArrival: '2025-10-02',
+    duration: 52,
+    essentialCriteria: ['hasTactileFlooring'],
+    application,
+    notes: 'Test notes',
+  })
+
+  it('should generate the expected summary list', () => {
+    expect(placementRequestSummaryList(placementRequest).rows).toEqual(
+      expectedSummaryListItems(
```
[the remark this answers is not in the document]
gregkhawkins: will do, thanks

server/utils/placementRequests/preferredApsRow.ts
```diff
@@ -8,7 +8,7 @@ export const preferredApsRow = (placementRequest: PlacementRequestDetail): Summa
   if (premises.length) {
     const apList = premises.map(p => `<li>${p.name}</li>`)
     return {
-      key: { text: 'Preferred APs' },
+      key: { text: 'Preferred AP' },
```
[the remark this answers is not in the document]
gregkhawkins: will do, thanks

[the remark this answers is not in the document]
gregkhawkins: they agreed to stick with `Preferred APs` so will make that change now - thanks

server/utils/match/index.ts
gregkhawkins: After sitting with this for a few hours. to avoid any chance of runtime exceptions when trying to access `licenseExpiryDate` are you happy to leave as-is and stay defensive?

gregkhawkins: resolving comment now as realised you were not suggested changes here just feeding into my question. Will merge now also but can make some changes in follow-up PR (if anything on this needs addressing).

PR #2281, opened 2025-01-10
removed factory

daveawc:
removed factory


froddd:
I believe this is superseded by #2279 which goes a bit further with the clean up!

PR #2288, opened 2025-01-13
bugfix: Ensure all date errors returned at once

froddd:
When submitting a new booking without an arrival or departure date, or with either of those invalid, only one error would be indicated for the first error field, with the other field ignored. This ensures all errors are returned at once.

A small refactor of the booking post handler makes it follow the more usual `try/catch` pattern.

## Screenshot of changes

Screenshot 2025-01-13 at 16 20 06

### DIFF
--- a/server/controllers/match/placementRequests/occupancyViewController.test.ts
+++ b/server/controllers/match/placementRequests/occupancyViewController.test.ts
@@ -23,6 +23,7 @@ import {
   dayAvailabilitySummaryListItems,
 } from '../../../utils/match/occupancy'
 import { placementRequestSummaryList } from '../../../utils/placementRequests/placementRequestSummaryList'
+import { ValidationError } from '../../../utils/errors'
 
 describe('OccupancyViewController', () => {
   const token = 'SOME_TOKEN'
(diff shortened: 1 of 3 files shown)


server/controllers/match/placementRequests/occupancyViewController.test.ts
```diff
@@ -47,7 +48,7 @@ describe('OccupancyViewController', () => {
   })
 
   beforeEach(() => {
-    jest.resetAllMocks()
+    jest.clearAllMocks()
```
bobmeredith, line 51: I'm wondering why the change to just clear the mocks rather than reset? Doesn't that leave the mock implementation intact?
It's not a biggie as all the mock implementations get set up again anyway.

bobmeredith (approved):
Looks good to me.
A super minor query is all.

server/controllers/match/placementRequests/occupancyViewController.test.ts
froddd: Using `resetAllMocks` results in the `describe` block on line 301 to fail if run consecutively -- not entirely sure why, I suspect the way we're mocking requests and such may play a part!

PR #2293, opened 2025-01-15
APS-1678 AP day summary view

bobmeredith:
# Context

# Changes in this PR


server/testutils/factories/cas1PremiseCapacity.ts
```diff
@@ -96,6 +96,16 @@ class PremisesCharacteristicAvailability extends Factory<Cas1PremiseCharacterist
...
+
+  strictlyOverbooked() {
+    const availableBedsCount = faker.number.int({ min: 5, max: 20 })
+    const bookingsCount = faker.number.int({ min: availableBedsCount + 1, max: 30 })
+
+    return this.params({
+      availableBedsCount,
+      bookingsCount,
+    })
+  }
```
froddd, line 108: Nice!

server/utils/premises/occupancy.test.ts
```diff
@@ -48,4 +55,59 @@ describe('apOccupancy utils', () => {
...
+      Object.keys(occupancyCriteriaMap).map(characteristic =>
+        overbookedCharacteristics.includes(characteristic as Cas1SpaceBookingCharacteristic)
+          ? premiseCharacteristicAvailability
+              .strictlyOverbooked()
+              .build({ characteristic: characteristic as Cas1SpaceBookingCharacteristic })
+          : premiseCharacteristicAvailability
+              .available()
+              .build({ characteristic: characteristic as Cas1SpaceBookingCharacteristic }),
+      )
+
```
froddd, line 70: I'm aware we don't have designs for overall overbooking here, but it would probably be worth adding the test and implementation now anyway, which would leave only having to update the content if required.

server/utils/premises/occupancy.ts
```diff
@@ -64,3 +71,39 @@ export const durationSelectOptions = (durationDays?: string): Array<SelectOption
...
+  return overbookedCriteria.length
+    ? `This AP is overbooked on spaces with the following ${overbookedCriteria.length > 1 ? 'criteria' : 'criterion'}: ${overbookedCriteria.map(characteristic => occupancyCriteriaMap[characteristic].toLowerCase()).join(', ')}`
+    : ''
+}
+
+const summaryRow = (key: string, value: string): SummaryListItem =>
+  value && {
+    key: textValue(key),
+    value: textValue(value),
+  }
```
froddd, line 95: Could use the recent `summaryListItem` utility in `formUtils`? It doesn't check for `value` being truthy, but I am quite undecided as to whether the basic utility should 😄

froddd (approved):
Looks really good -- couple of suggestions around error checking and pre-emptively adding test and implementation for strict overbooking summary text.

server/controllers/manage/premises/apOccupancyViewController.test.ts
```diff
@@ -111,4 +115,59 @@ describe('AP occupancyViewController', () => {
...
+      expect(response.render).toHaveBeenCalledWith(
+        'manage/premises/occupancy/dayView',
+        expect.objectContaining({
+          backLink: paths.premises.occupancy.view({ premisesId }),
+          errorSummary: [{ text: 'Date "2023-02-29" is invalid.', href: '#date' }],
+        }),
+      )
+      expect(premisesService.find).toHaveBeenCalledWith(token, premisesId)
+      expect(premisesService.getDaySummary).not.toHaveBeenCalled()
+    })
```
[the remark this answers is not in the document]
froddd: Adding to that, it would also hide the error from Sentry and our alerting -- I think we should be aware if this ever happened, so that would be an issue!

server/utils/premises/occupancy.ts
bobmeredith: Makes sense.

server/controllers/manage/premises/apOccupancyViewController.ts
```diff
@@ -61,4 +68,34 @@ export default class ApOccupancyViewController {
...
+    return async (req: Request, res: Response) => {
+      const { token } = req.user
+      const { premisesId, date } = req.params
+      const { errorSummary } = fetchErrorsAndUserInput(req)
+      try {
+        DateFormats.isoToDateObj(date)
+      } catch (e) {
+        const dateError = { date: `Date "${date}" is invalid.` }
+        errorSummary.push(generateErrorSummary(dateError)[0])
+      }
```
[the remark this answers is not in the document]
bobmeredith: OK, makes sense. So I should just pass the invalid date to the API and let that fail.

server/controllers/manage/premises/apOccupancyViewController.test.ts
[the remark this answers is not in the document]
bobmeredith: OK, that's much simpler then. I don't even need an error reporting block!

server/utils/premises/occupancy.test.ts
bobmeredith: Tricky. I can come up with a text for overall overbooking, but the combination of overall and characteristic overbooking might be complex.
TBH, I think there are better ways to present this information.

server/controllers/manage/premises/apOccupancyViewController.ts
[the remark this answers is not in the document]
froddd: Yup, that should work absolutely fine!

PR #2297, opened 2025-01-16
APS-1815: Throw if page already visited

froddd:
When building the main task list or fetching statuses for building un-submitted application summaries, some answer combinations may unintentionally produce an infinite loop due to how the next page to check is determined.

To prevent this, a simple trail of pages that have already been visited is built whenever such a loop is running. If a page has already been visited, an error is now thrown: this prevents the current process running indefinitely, and will alert us of the trail of pages that caused the infinite loop, so the issue can be remedied.

In addition, when no data has been saved against a given page, the loop attempts to check the next page as defined in the Task. Due to certain logic combinations, this page may already have been checked, which may also produce an infinite loop (or, thanks to the fix above, an error!). To prevent this, pages are removed from the pool as they are checked.

# Context

https://dsdmoj.atlassian.net/browse/APS-1815

### DIFF
--- a/server/form-pages/utils/getTaskStatus.test.ts
+++ b/server/form-pages/utils/getTaskStatus.test.ts
@@ -144,4 +144,20 @@ describe('getTaskStatus', () => {
     expect(page3Instance.errors).toHaveBeenCalled()
     expect(page3Instance.next).toHaveBeenCalled()
   })
+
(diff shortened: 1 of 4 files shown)


bobmeredith (approved):
Looks good to me.
Took me a while to work out the iteration logic in the code, but the solution looks valid.

PR #2298, opened 2025-01-17
APS-1756 - Use CAS1 endpoints for booking reference data

davidatkinsuk:
This commit updates the placement service to use cas1 specific endpoints to retrieve the following:

* non arrival reasons
* departure reasons
* move on categories

Because these endpoints only return ‘active’ values, the logic for non arrival reasons has been simplified, removing filtering on this value


muhammad-elabdulla (approved):
LGTM

froddd (approved):

PR #2301, opened 2025-01-22
APS-1826: session space search state

froddd:
# Context

https://dsdmoj.atlassian.net/browse/APS-1826

# Changes in this PR

The ticket linked above covers a bug with the Find and Book process, where AP requirements selected in the suitability search would cause an error when viewing a day's availability further down the search. This bug is now fixed!

However, this PR covers a whole lot of ground: it effectively rewrites the Find and book search to rely on a search state saved in session. This removes the need for passing query parameters around between screens. This has several advantages:

- All the filtering data selected during any of the search steps is available to be passed in the payload when the booking is made, which is a new (or originally overlooked!) requirement
- The back links can simply point to urls free of query strings, as the state is responsible for rendering the page in the right, erh, state. This also means we can now populate the Arrival and Departure dates if the user clicks back from the confirmation screen -- before this they were simply blank
- There is a lot less complexity in the code to handle the occupancy view filtering, as there is no longer a need to split data from the filters from data from the booking form
- Fully removes the `apType` parameter

This also includes quite a bit of housekeeping, removing unnecessary types/factories/constants and trying to simplify search criteria -- a bit 🙂


server/utils/match/spaceSearch.ts
```diff
@@ -0,0 +1,116 @@
...
+export type SpaceSearchRoomCriteria = keyof typeof spaceSearchCriteriaRoomLevelLabels
+
+export const spaceSearchResultsCharacteristicsLabels = {
+  ...spaceSearchCriteriaApLevelLabels,
+  ...spaceSearchCriteriaRoomLevelLabels,
+  isSuitableForVulnerable: placementCriteriaLabels.isSuitableForVulnerable,
+} as const
+
+export type SpaceSearchState = {
+  applicationId: string
```
bobmeredith, line 41: Why do we need the application ID in here? The search state is always keyed on the placementRequest Id, so I would have thought the application ID would be available through that.

server/utils/placementCriteriaUtils.ts
```diff
@@ -73,30 +75,25 @@ export const specialistApTypeCriteriaLabels = filterByType<SpecialistApTypeCrite
...
+      return 'esap'
+    case 'isRecoveryFocussed':
+      return 'rfap'
+    case 'isMHAPElliottHouse':
+      return 'mhapElliottHouse'
+    case 'isMHAPStJosephs':
+      return 'mhapStJosephs'
+    default:
+      return 'normal'
+  }
```
bobmeredith, line 98: I think you could do this with a mapping object look-up.

bobmeredith (approved):
Looks really good. Much simpler than trying to pass this stuff about by querystring!

My only issue is the longevity of the session and how that ties into the design - which doesn't really indicate what's happening! The initial hit on the suitability search populates the search criteria from the PR which can be modified by the user and end up populating the booking. With this session model, if the user abandons a booking, then goes back to it later, the initial search will be populated from the session rather than the PR, and it's not at all obvious to the user.  Can we rely on users to cross check the filters with the summary info at the top?
A design change could fix this, but in the meantime, I suggest that we clear the session for a PR id if the user hits the PR details page, as I think that's the only place the journey can start. That way, the user will always start the journey 'fresh'.

server/controllers/match/placementRequests/spaceBookingsController.ts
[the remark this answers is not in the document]
froddd: I agree with the naming, and reckon `ApSearch` would be more relevant. However, the service is meant to manage data needed by the controller, and this is exactly what the session store is, so I do think it is in the right place...

server/services/spaceService.ts
```diff
@@ -22,4 +28,27 @@ export default class SpaceService {
...
+    session.spaceSearch[placementRequestId] = {
+      ...this.getSpaceSearchState(placementRequestId, session),
+      ...data,
+    }
+
+    return session.spaceSearch[placementRequestId]
+  }
+
+  removeSpaceSearchState(placementRequestId: PlacementRequestDetail['id'], session: Request['session']): void {
+    delete session.spaceSearch?.[placementRequestId]
```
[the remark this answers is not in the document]
froddd: I don't think so, this is just a courtesy utility -- although if we implement the 'clear on viewing placement request' it will become quite useful!

server/utils/match/spaceSearch.ts
froddd: Not strictly required, but it allows us to simply pass the search state to produce the parameters needed when making the actual search call -- so we can either have the `applicationId` as part of the state, or extend the signature for performing the search for both the client and service to also include an `applicationId` parameter.

server/utils/placementCriteriaUtils.ts
froddd: Definitely, though I'm tempted to leave it until we have more clarity on how the AP type will be handled with respects to switching to checkboxes -- if this happens!

server/paths/match.ts
```diff
@@ -12,6 +12,7 @@ const v2Match = {
     search: {
       spaces: v2PlacementRequestSearchPath.path('new'),
       occupancy: v2PlacementRequestSearchOccupancyPath,
+      occupancyBook: v2PlacementRequestSearchOccupancyPath.path('book'),
```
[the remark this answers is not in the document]
froddd: This is the occupancy view -- or the 'availability search' -- on which there are 2 forms now using `POST`: the filter form, and the 'Book your space' form.

PR #2307, opened 2025-01-28
Force session save before redirect

froddd:
# Context

The load tests have highlighted potential issues with the session saving mechanism, especially for the 'Record departure' flow. It seems that, in some cases, the session does not correctly get saved before the server issues the redirect to the next screen, resulting in errors being shown and the flow halting.

The Express documentation mentions that, in some cases, including when a redirect is involved, it may be useful to explicitly save the session before continuing (https://github.com/expressjs/session#sessionsavecallback). This PR ensures the session is manually saved before the redirect is called, which should remove the issue above.

This PR also removes the `cookie-session` package, which wasn't actually used anywhere and was simply causing confusion as to what session strategy was being used.


server/controllers/match/placementRequests/spaceBookingsController.ts
```diff
@@ -86,7 +86,11 @@ export default class {
           'success',
           'You have now booked a place in this AP for this person. An email will be sent to the AP, to inform them of the booking.',
         )
-        return res.redirect(`${paths.admin.cruDashboard.index({})}?status=matched`)
+        this.spaceSearchService.removeSpaceSearchState(id, req.session)
```
gregkhawkins, line 89: this is an extra? Just cleaning up session that was missed previously?

gregkhawkins (approved):

server/controllers/match/placementRequests/spaceBookingsController.ts
froddd: Yeah, just a bit more tidying up as a courtesy to our server :)

bobmeredith (approved):
Looks OK.
I'm wondering if there is a more global solution to saving the session before all redirects.
I guess this will only show up with multiple express servers and non-sticky sessions so an alternative would be to configure session stickyness but I guess this solution is simpler and more robust.

PR #2310, opened 2025-01-29
Update CRU Member Role Descriptions

davidatkinsuk:
Add clarification that all CRU Members should also have the workflow manager role assigned.

Screenshot 2025-01-29 at 17 02 46


bobmeredith (approved):
LGTM

PR #2319, opened 2025-02-04
APS-792 - Remove defunct ‘NAT’ filtering

davidatkinsuk:
As a stop gap for removing the ‘National’ AP Area we filtered it out of drop downs on the UI. This AP Area has now been removed in the backend, so the UI filtering is no longer required

Screenshot 2025-02-04 at 12 54 35


bobmeredith (approved):
LGTM

PR #2323, opened 2025-02-05
Remove step to generate types from CI pipeline

froddd:
This step is run separately, and updated types should only be brought in as and when they are needed.


bobmeredith (approved):
LGTM

PR #2328, opened 2025-02-06
APS-1910: Show booking requirements on placement details

froddd:
# Context

https://dsdmoj.atlassian.net/browse/APS-1910

# Changes in this PR

This shows the requirements for the booking exactly as they were selected during the Find and book journey.

To be explicit about the booking made, when no requirements have been selected, the text 'None' is shown in a slightly greyed out hue.

### New section on placement details tab

Screenshot 2025-02-06 at 15 21 40

### DIFF
--- a/server/utils/placements/index.ts
+++ b/server/utils/placements/index.ts
@@ -20,6 +20,9 @@ import paths from '../../paths/manage'
 import { hasPermission } from '../users'
 import { TabItem } from '../tasks/listTable'
 import { summaryListItem } from '../formUtils'
+import { requirementsHtmlString } from '../match'
+import { apTypeCriteriaLabels, specialistApTypeCriteria } from '../placementCriteriaUtils'
+import { filterApLevelCriteria, filterRoomLevelCriteria } from '../match/spaceSearch'
 
 export const statusTextMap: Record<Cas1SpaceBookingSummaryStatus, string> = {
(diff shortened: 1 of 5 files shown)


server/utils/placements/index.ts
```diff
@@ -169,6 +172,32 @@ export const departureInformation = (placement: Cas1SpaceBooking): SummaryList =
...
 }
 
+export const requirementsInformation = (placement: Cas1SpaceBooking): SummaryList => {
+  const requirements = placement.requirements.essentialCharacteristics
+  const apType =
+    apTypeCriteriaLabels[requirements.find(requirement => specialistApTypeCriteria.includes(requirement)) || 'normal']
+  const apRequirements = filterApLevelCriteria(requirements)
+  const roomRequirements = filterRoomLevelCriteria(requirements)
+
+  const noneSelected = `<span class="text-grey">None</span>`
```
bobmeredith, line 182: Nice.
There are probably other places where this could be used. In fact, couldn't `requirementsHtmlString` return it if the array is empty?

bobmeredith (approved):
LGTM.
Just one suggestion to consider.
It may be a bit far reaching for this PR though.

server/utils/placements/index.ts
froddd: I was just thinking about that as I'd started looking at the actual issue with 'change placement' when criteria is empty. I'll merge this and apply it!

PR #2336, opened 2025-02-11
APS-1892 Match occupancy day view update

bobmeredith:
# Context

# Changes in this PR

### DIFF
--- a/assets/sass/components/_previous-next.scss
+++ b/assets/sass/components/_previous-next.scss
@@ -1,5 +1,6 @@
 .prev-next {
   font-weight: bold;
+  font-size: 19px;
   margin-bottom: govuk-spacing(3);
 
   .prev-next--prev {
--- a/server/utils/premises/occupancy.test.ts
+++ b/server/utils/premises/occupancy.test.ts
@@ -27,6 +27,7 @@ import { premiseCharacteristicAvailability } from '../../testutils/factories/cas
 import { getTierOrBlank } from '../applications/helpers'
 import { laoSummaryName } from '../personUtils'
 import { spaceSearchCriteriaRoomLevelLabels } from '../match/spaceSearch'
+import { capacityWithCriteria } from '../match/occupancy.test'
 
 describe('apOccupancy utils', () => {
   describe('occupancyCalendar', () => {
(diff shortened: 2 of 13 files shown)


server/utils/premises/occupancy.test.ts
```diff
@@ -128,26 +128,87 @@ describe('apOccupancy utils', () => {
...
       })
     })
+
+    it('should generate a list of day summary rows including criteria', () => {
+      const expected = [
+        { key: { text: 'Capacity' }, value: { text: '20' } },
+        { key: { text: 'Booked spaces' }, value: { text: '21' } },
+        { key: { text: 'Out of service beds' }, value: { text: '2' } },
+        { key: { text: 'Available spaces' }, value: { text: '-3' } },
+        emptyRow,
```
froddd, line 179: This isn't great semantically -- I wonder what a screen reader would make of the resulting markup. Could this be split into 2 summary lists?

server/utils/premises/occupancy.test.ts
```diff
@@ -222,4 +283,34 @@ describe('apOccupancy utils', () => {
...
+  //
+  //       expect(summaryList).toEqual([
+  //         { key: { text: 'AP capacity' }, value: { text: '20' } },
+  //         { key: { text: 'Booked spaces' }, value: { text: '21' } },
+  //         { key: { text: 'Suitable for sex offenders spaces available' }, value: { text: '3' } },
+  //         { key: { text: 'Step-free spaces available' }, value: { text: '0' } },
+  //       ])
+  //     })
+  //   })
+  // })
```
froddd, line 315: I think this can now be removed

server/utils/premises/occupancy.ts
```diff
@@ -97,19 +99,66 @@ export const generateDaySummaryText = (daySummary: Cas1PremisesDaySummary): stri
...
+          html: `${pluralize('bed', available)}${booked ? `<span class="govuk-!-padding-right-2"></span><a href="${characteristic ? `?characteristics=${characteristic}` : '?'}"> ${pluralize('booking', booked)}</a>` : ''} ${(booked || available) && booked >= available ? `<strong class="govuk-tag govuk-tag--${booked > available ? 'red' : 'yellow'} govuk-tag--float-right">${booked > available ? 'Overbooked' : 'Full'}</strong>` : ''}`,
+        },
+      }
+    : null
+}
+
+export const daySummaryRows = (
+  daySummary: Cas1PremisesDaySummary,
+  roomCharacteristics: Array<Cas1SpaceBookingCharacteristic> = null,
+  characteristicsMode: 'singleRow' | 'doubleRow' | 'none' = 'none',
```
froddd, line 121: I was wondering why I was seeing code to do with tags and such but not seeing them in the browser... Nice!

This should _probably_ be in a separate branch, but I'll let it live here 🙂

froddd (commented):
This looks great! much better info -- would also love to see the 'updated' availability row design take precedence, it works a treat.

Prev/Next navigation could be tweaked a bit to match designs more and be a bit more usable. Other than that this is probably good to go!

server/utils/premises/occupancy.ts
bobmeredith: Yes, I jumped the gun a bit putting this up for review. I'll put a feature flag round it to turn it on.

server/utils/premises/occupancy.ts
```diff
@@ -97,19 +99,66 @@ export const generateDaySummaryText = (daySummary: Cas1PremisesDaySummary): stri
...
 }
 
-export const daySummaryRows = (daySummary: Cas1PremisesDaySummary) => {
+const availabilityRow = (
+  name: string,
+  characteristic: Cas1SpaceBookingCharacteristic,
+  available: number,
+  booked: number,
+) => {
+  return booked || available
```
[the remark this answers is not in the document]
bobmeredith: Now that is a really good question. 
I deliberately made it not appear to cover the case where the AP simply doesn't have any rooms with the criterion e.g. not every AP has wheelchair rooms I believe.
But, you're right - if the AP has a wheelchair room but it's out of service, it should probably be shown? 
Also of course if the AP has no such rooms and the CRU, in a fit of madness, booked someone in, it should show, but I think that's already covered.

assets/sass/components/_previous-next.scss
```diff
@@ -1,5 +1,6 @@
 .prev-next {
   font-weight: bold;
+  font-size: 19px;
```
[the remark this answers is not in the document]
bobmeredith: I've removed the whole file

assets/sass/components/_previous-next.scss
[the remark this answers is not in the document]
bobmeredith: So I can actually just use the comonent - all it needs is an empty array of pages and bingo!

server/utils/premises/occupancy.test.ts
bobmeredith: Actually, I don't think the blank line is needed at all with the POC layout as I don't think there is any useful info in the old table.
I'd rather not spend time improving this blank line thing until we know that it's needed.

server/utils/premises/occupancy.ts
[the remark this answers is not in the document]
bobmeredith: Hmm, this is not easy to do as the API capacity return only shows rooms with the oosbs already removed. We have enough information in the UI I guess, but it would require using the list of OOSBs to force the row to show.

server/utils/premises/occupancy.test.ts
froddd: Fair enough, let's go ahead with this!

froddd (approved):
Looks good!

PR #2341, opened 2025-02-12
CHORE: Store failed integration tests artefacts

froddd:
GitHub actions skips steps after failing ones by default, resulting in the steps to store artefacts not being run.

This ensure the steps to store any screenshots or videos from failed integration tests are always run.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -159,12 +159,14 @@ jobs:
           path: test_results/cypress
 
       - name: Store Integration tests videos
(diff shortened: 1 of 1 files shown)


bobmeredith (approved):
Looks good.
So with this, the upload will happen even if the job is cancelled. I'm not entirely sure that this matters or is a good or bad thing.

If it does, perhaps `if: success() || failure()` would do the job.

froddd:
@bobmeredith good point! Updated

PR #2343, opened 2025-02-12
APS-1893: Update booking confirmation message

froddd:
# Context

https://dsdmoj.atlassian.net/browse/APS-1893

# Changes in this PR

This updates the booking success message to include the CRN of the person and the name of the premises the booking was made for. The content is now split into heading and an extra paragraph to aid with readability.

## Screenshots of UI changes

Screenshot 2025-02-12 at 12 25 56

### DIFF
--- a/integration_tests/pages/page.ts
+++ b/integration_tests/pages/page.ts
@@ -104,8 +104,11 @@ export default abstract class Page {
     })
   }
 
-  shouldShowBanner(copy: string): void {
-    cy.get('.govuk-notification-banner').contains(copy)
+  shouldShowBanner(heading: string, body?: string): void {
+    cy.get('.govuk-notification-banner').contains(heading)
+    if (body) {
+      cy.get('.govuk-notification-banner').contains(body)
+    }
   }
 
   shouldNotShowBanner(): void {
--- a/integration_tests/pages/page.ts
+++ b/integration_tests/pages/page.ts
@@ -104,11 +104,16 @@ export default abstract class Page {
     })
   }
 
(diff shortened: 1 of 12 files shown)


integration_tests/pages/page.ts
```diff
@@ -104,8 +104,11 @@ export default abstract class Page {
     })
   }
 
-  shouldShowBanner(copy: string): void {
-    cy.get('.govuk-notification-banner').contains(copy)
+  shouldShowBanner(heading: string, body?: string): void {
+    cy.get('.govuk-notification-banner').contains(heading)
+    if (body) {
+      cy.get('.govuk-notification-banner').contains(body)
```
bobmeredith, line 110: This will only be able to handle text for the comparison and so, may not work if the html sent to the body is broken up into multiple elements.
There is a routine in `page.ts` - `parseHtml` that I think would allow full HTML comparisons.

bobmeredith (approved):
Looks really nice!
For consistency, you should really update all the other banners in the system and sort out the grammar :)

Just one small comment about possible future-proofing of the integration test

integration_tests/pages/page.ts
froddd: Very good call -- I've updated `shouldShowBanner` to make use of this. Given a couple of previous tests only cared if the banner _contained_ the text, I've also added an `options` argument: if setting `options.exact` to `false`, the assertion is that the banner _contains_ the text; the default (`exact: true`) asserts that the banner content and text are an exact match.

bobmeredith (approved):
Nice!
Looks good to go.

PR #2346, opened 2025-02-12
APS-1957 Occupancy view filter crash

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/APS-1957

# Changes in this PR
On the M&M occupancy view calendar page, if the filter is updated with a single room criterion, the post contains a string rather than an array. This is stored in the session and the view controller tries to do a `.map` causing the error.

## Screenshots of UI changes
No UI changes


froddd (approved):
Looks good!

PR #2348, opened 2025-02-13
CHORE: Install npm dependencies before starting E2E tests

froddd:
Since we have removed the orphaned `build` step, npm dependencies are no longer available to the E2E test runner -- this ensures they are installed first, before running the tests.

The E2E tests continued to run fine for a few merges, as they were restoring a cached version of the dependencies. There has since been a change to the dependencies, resulting in the cache becoming invalid, and the tests no longer having readily installed dependencies to use.


davidatkinsuk (approved):
LGTM 👍

PR #2350, opened 2025-02-13
APS-1831: Render timeline event payload based on template for Booking Changed event

froddd:
# Context

https://dsdmoj.atlassian.net/browse/APS-1831

# Changes in this PR

If the timeline event has a payload of type `booking_changed`, this now uses a template to render the content of the timeline event. I took some liberty to make the template more readable, by actually listing the changes next to bullet points.

This PR also tidies up the markup used for timelines, which had some incorrectly nested elements.

_NOTE: the API currently returns an empty list for `previousCharacteristics` when there has been no change to the room criteria. This means the UI is unable to determine whether there has been no change, or there has been a change from no criteria to a new set. https://dsdmoj.atlassian.net/browse/APS-2013 should eventually fix this, with no further change required to the UI._

## Screenshots of UI changes

Screenshot 2025-02-20 at 15 43 07

### DIFF
--- a/server/views/partials/timelineEvents/booking_changed.njk
+++ b/server/views/partials/timelineEvents/booking_changed.njk
@@ -1,0 +1,12 @@
+<p class="govuk-body">A placement at {{ premises.name }} has been changed:</p>
+<ul class="govuk-list govuk-list--bullet">
+    {% if previousExpectedArrival %}
+        <li>Arrival date changed from {{ previousExpectedArrival }} to {{ expectedArrival }}</li>
+    {% endif %}
+    {% if previousExpectedDeparture %}
+        <li>Departure date changed from {{ previousExpectedDeparture }} to {{ expectedDeparture }}</li>
+    {% endif %}
+    {% if previousCharacteristics !== 'None' or characteristics !== 'None' %}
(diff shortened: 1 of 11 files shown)


server/views/partials/timelineEvents/booking_changed.njk
```diff
@@ -0,0 +1,12 @@
...
+<ul class="govuk-list govuk-list--bullet">
+    {% if previousExpectedArrival %}
+        <li>Arrival date changed from {{ previousExpectedArrival }} to {{ expectedArrival }}</li>
+    {% endif %}
+    {% if previousExpectedDeparture %}
+        <li>Departure date changed from {{ previousExpectedDeparture }} to {{ expectedDeparture }}</li>
+    {% endif %}
+    {% if previousCharacteristics %}
+        <li>Room criteria changed from {{ previousCharacteristics }} to {{ characteristics }}</li>
+    {% endif %}
```
bobmeredith, line 11: It might be a strange edge-case, but ATM, we don't prevent an update being made where there are no changes at all!  In this case, the timeline event displays nothing. 
Perhaps it would be nice to add a 'No change' text if not  `previousExpectedArrival or previousExpectedDeparture or previousCharacteristics`

bobmeredith (approved):
Looks really nice.
At least it will when the API bug is fixed.

PR #2351, opened 2025-02-13
APS-1871 Filter bed characteristics

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/APS-1871

# Changes in this PR

This PR affects two pages - out-of-service-bed view page and bed view page. They are the only users of  `translateCharacteristic` that looks-up text for a characteristic in a `CharacteristicPair` format.

I've removed this old mechanism and replaced with a lookup - but also relocated the lookup map to characteristicUtils which seems like a better place for it to be than buried in match - hence a lot of file touches.
I've also written a routine to generate a bullet list from a set of characteristics that can be used in several places.

## Screenshots of UI changes

<details>
  <summary>Before</summary>

</details>

<details>
  <summary>After</summary>

</details>

### DIFF
--- a/server/utils/characteristicsUtils.test.ts
+++ b/server/utils/characteristicsUtils.test.ts
@@ -1,29 +1,30 @@
-import { translateCharacteristic } from './characteristicsUtils'
-import { CharacteristicPair } from '../@types/shared'
+import {
+  characteristicsBulletList,
+  characteristicsPairToCharacteristics,
+} from './characteristicsUtils'
+import { Cas1SpaceBookingCharacteristic, CharacteristicPair } from '../@types/shared'
 
-describe('characteristicsUtils.translateCharacteristic(CharacteristicPair)', () => {
-  describe('when the propertyName is found', () => {
-    const propertyNameFoundInTranslation = 'isSingle'
-    const successfulTranslation = 'Single room'
-    const characteristicWithTranslation = <CharacteristicPair>{
-      propertyName: propertyNameFoundInTranslation,
-      name: 'Is this a single room?',
-    }
+describe('characteristicsUtils', () => {
 
-    it('returns the translation (in the form of a statement) for the characteristic', () => {
-      expect(translateCharacteristic(characteristicWithTranslation)).toEqual(successfulTranslation)
+  describe('characteristicsPairToCharacteristics', () => {
+    it('should flatten and filter a list of characteristicPairs', () => {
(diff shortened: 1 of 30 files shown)


server/utils/characteristicsUtils.test.ts
```diff
@@ -1,29 +1,55 @@
...
+  //   const propertyNameNotFoundInTranslation = 'isRed'
+  //   const characteristicWithoutTranslation = <CharacteristicPair>{
+  //     propertyName: propertyNameNotFoundInTranslation,
+  //     name: 'Is this a a red room?',
+  //   }
+  //
+  //   it('returns the propertyName as a (just) human-readable fallback', () => {
+  //     expect(translateCharacteristic(characteristicWithoutTranslation)).toEqual('isRed')
+  //   })
+  // })
```
froddd, line 32: Kill it! 💥 

The `describe` above could do with a label tweak, too!

server/utils/characteristicsUtils.ts
```diff
@@ -1,6 +1,27 @@
-import { CharacteristicPair } from '@approved-premises/api'
-import characteristicLookup from '../i18n/en/characteristics.json'
+import { Cas1SpaceBookingCharacteristic, CharacteristicPair } from '@approved-premises/api'
```
froddd, line 1: Should these utilities refer to the `Cas1SpaceBookingCharacteristic` type? It seems the more generic `Cas1SpaceCharacteristic` would be more suited -- although this might only move some of the type casting around...

froddd (approved):
This looks great! Couple of small comments.

One further question (which doesn't really impact this being merged): would we then be able to deprecate the characteristics _pairs_ from the API, and instead return straight characteristics? It would be nice to fully remove the translation we still need to do for those!

bobmeredith:
Yes, we could deprecate the `characteristicsPairs` and that would make life much simpler. It seemed that the actual translations were all held in the UI anyway, so the API mappings are redundant.
I'll raise a ticket.

server/utils/characteristicsUtils.test.ts
bobmeredith: Oh yes - thanks

server/utils/characteristicsUtils.ts
bobmeredith: Yes, you're right. Cas1SpaceBookingCharacteristic is right for the map though as that only maps the names for the 6 room-level ones.
The switch didn't throw up any type errors.

PR #2357, opened 2025-02-18
Remove deprecated reports

davidatkinsuk:
This commit removes reports that should no longer be used as they have been superseded by newer versions

### Before

Screenshot 2025-02-19 at 08 35 10

### After

Screenshot 2025-02-18 at 12 48 25


Rosie-Brigham (approved):
👍

PR #2359, opened 2025-02-19
APS-1912 Extend booking

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/APS-1912

# Changes in this PR

Handles the 'Extend booking' journey - an amend on a placement post arrival (i.e. the actualArrivalDate is populated). This involved changes to the `changes` form.

- Added 'Actual arrival date' to the top summary.  I left the 'Expected arrival date' in place so the user can relate this to the arrival date shown on the confirmation page - which will not be changed.
- Removed the Arrival date input from the bottom form. 
- I re-titled the bottom section to 'Change placement' as it can change more than just dates, and changed the field labels to 'New arrival date' and 'New departure date' with the hints re-labelled as 'Current arrival date' and 'Current departure date'. This means that the form works well in both amend (both arrival and departure) and Extend (departure only) modes.
- The calendar displays from the actual arrival date, rather than the expected arrival date.
- Given that the departure is being extended, it's likely that the end date will be after the current end date. This means that the user will probably need to use the duration control in the filter to advance the period - but the current behaviour is odd in that the calendar is only shown up until the departure date, regardless of the rounding-up that occurs in the duration selector. To make the behaviour of the form slightly more logical, I forced the calendar end-date to use the rounded-up duration shown in the duration selector on the filter.

## Screenshots of UI changes

<details>
  <summary>Occupancy view page</summary>

</details>

<details>
  <summary>Error - departure must be after arrival</summary>

</details>

### DIFF
--- a/server/controllers/manage/premises/placements/changesController.ts
+++ b/server/controllers/manage/premises/placements/changesController.ts
@@ -21,7 +21,7 @@ import { createQueryString, makeArrayOfType } from '../../../../utils/utils'
 import { DateFormats, dateAndTimeInputsAreValidDates } from '../../../../utils/dateUtils'
 import { CriteriaQuery } from '../../../match/placementRequests/occupancyViewController'
 import { convertKeyValuePairToCheckBoxItems } from '../../../../utils/formUtils'
-import { durationSelectOptions } from '../../../../utils/match/occupancy'
+import { durationSelectOptions, getClosestDuration } from '../../../../utils/match/occupancy'
 import { OccupancySummary } from '../../../../utils/match/occupancySummary'
 import managePaths from '../../../../paths/manage'
 import matchPaths from '../../../../paths/match'
@@ -43,6 +43,7 @@ interface ViewRequest extends Request {
   body: ObjectWithDateParts<'arrivalDate'> &
     ObjectWithDateParts<'departureDate'> & {
       criteria: string
+      actualArrivalDate: string
     }
 }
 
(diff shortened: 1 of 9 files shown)


integration_tests/mockApis/premises.ts
```diff
@@ -90,17 +90,17 @@ const stubPremisesStaffMembers = (args: { premisesId: string; staffMembers: Arra
...
 }) =>
   stubFor({
     request: {
       method: 'GET',
       urlPath: paths.premises.capacity({ premisesId: args.premisesId }),
       queryParameters: {
-        startDate: { equalTo: args.startDate },
-        endDate: { equalTo: args.endDate },
+        startDate: args.startDate ? { equalTo: args.startDate } : undefined,
+        endDate: args.endDate ? { equalTo: args.endDate } : undefined,
```
froddd, line 103: Are these changes needed? Seems the changes in the signature clash with the changes in the expected params (only `endDate` is now marked as optional in signature, yet both `endDate` and `startDate` check for truthiness), and I think we always do pass the `premisesCapacity` too?

froddd (approved):
Couple of minor comments, seems good to go!

server/controllers/manage/premises/placements/changesController.ts
```diff
@@ -79,13 +80,19 @@ export default class ChangesController {
       const { errors, errorSummary, userInput } = fetchErrorsAndUserInput(req)
 
       const placement = await this.placementService.getPlacement(token, placementId)
-
+      let mode: 'amendment' | 'extension' = 'amendment'
```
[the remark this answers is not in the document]
bobmeredith: Yup, I guess I introduced the `mode` thinking it would need to do more than it ended up doing.

integration_tests/mockApis/premises.ts
bobmeredith: So, this change is about allowing the stubbing to be less specific.
All the parameters are always passed in the API call, but it was getting quite complex to derive the exact parameters (startDate, endDate) that need to be matched by the mock.
I guess the thing is that if we don't specify the mock's query parameters, we should check that the mock was called with the parameters we expect - which is much the same amount of effort.

Anyway, I've reverted the mock above and changed the code to supply the correct start and end dates in all cases.

froddd: I think it does make sense, and it's most likely useful to have a more flexible mock -- I was mostly confused that only `endDate` in the signature was made optional, and that the response would be `undefined` if capacity wasn't provided -- I'd expect that response to be a factory if not defined, at least?

PR #2360, opened 2025-02-20
APS-1951 Pen-test fix - Type confusion through parameter tampering

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/APS-1951

# Changes in this PR
Add recommended change (test for type string on date parameter). Also added a test. 

Note: the test passes (date validation returns false) even without the fix so I don't think there was ever an actual vulnerability. 

## Screenshots of UI changes

No UI change


froddd (approved):
👍

PR #2361, opened 2025-02-20
APS-1920 Remove root backlinks

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/APS-1920

# Changes in this PR

Removes pointless backlinks from two pages that are accessed directly from the root menu.

- Out of service beds listing for all premises.
- User management

### Before
image
image

### After
image
image


froddd (approved):
👍

PR #2363, opened 2025-02-21
APS-1922 Remove extra day on calendars

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/APS-1922

# Changes in this PR

Passes the endDate to `premisesGetCapacity` as a day earlier than the booking end date in match. Also passes the end date a one day earlier in manage.
Bookings have a start date and an end date - but the end date is the day of departure (check-out) so shouldn't be considered as included in the booking as far as the occupancy calculations are concerned.
In Manage, if the user selects to see a period of n weeks, they would expect to see exactly 7n days so the last day shown (as sent to the API) should be `start + 7n -1`

### Before
<details>
  <summary>Match occupancy view</summary>

</details>

<details>
  <summary>Manage occupancy view</summary>

</details>

### After
<details>
  <summary>Match occupancy view</summary>

</details>

<details>
  <summary>Manage occupancy view</summary>

</details>


froddd (approved):
LGTM! 👍

PR #2365, opened 2025-02-21
APS-1998 AP Calendar should show overbooked characteristics

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/APS-1998

# Changes in this PR
Given that this wasn't considered in the design, I've just forced the day status to 'overbooked' so that it shows in red if there is a characteristic overbooking e.g. wheelchair bookings exceeds wheelchair available spaces. If a characteristic is full, it doesn't override a total 'available' status as having all the special rooms full would be more-or-less expected.

Also, I noticed that the day summary back-button was a fixed path and thus loses the selected period and filters. This is very annoying - so I used the standard session-based handler for it.

## Screenshots of UI changes

<details>
  <summary>Before</summary>

</details>

<details>
  <summary>After</summary>

</details>


froddd (approved):
Nice!

PR #2366, opened 2025-02-21
Enable reports with PII role on admin ui

davidatkinsuk:
# Changes in this PR

* Make the 'report viewer with PII' role available on the user admin page
* Show reports that include PII on the reports page if the user has the required permission

### After

New role option

Screenshot 2025-02-21 at 15 53 19

Reports screen without PII role

Screenshot 2025-02-21 at 16 28 33

Reports screen with PII role

Screenshot 2025-02-21 at 16 30 43

### DIFF
--- a/server/utils/reportUtils.test.ts
+++ b/server/utils/reportUtils.test.ts
@@ -1,9 +1,11 @@
 import { reportOptions } from './reportUtils'
+import { userDetailsFactory } from '../testutils/factories'
 
 describe('reportUtils', () => {
   describe('reportOptions', () => {
-    it('should return a list of report options', () => {
-      expect(reportOptions).toEqual([
+    it('should return a list of report options, excluding pii by default', () => {
+      const user = userDetailsFactory.build({ permissions: [] })
+      expect(reportOptions(user)).toEqual([
         {
           value: 'outOfServiceBeds',
           text: 'Out of service beds',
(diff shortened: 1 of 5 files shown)


server/utils/reportUtils.test.ts
```diff
@@ -1,9 +1,10 @@
+import { fromPartial } from '@total-typescript/shoehorn'
 import { reportOptions } from './reportUtils'
 
 describe('reportUtils', () => {
   describe('reportOptions', () => {
-    it('should return a list of report options', () => {
-      expect(reportOptions).toEqual([
+    it('should return a list of report options, excluding pii by default', () => {
+      expect(reportOptions(fromPartial({ permissions: [] }))).toEqual([
```
froddd, line 7: A better approach here would be to use the user details factory to be more explicit about the argument being passed, and avoid type mocking:

```suggestion
      const user = userDetailsFactory.build({ permissions: [] })
      expect(reportOptions(user)).toEqual([
```

(This will obviously require the factory imported too, so can't be merged as is 🙂)

server/utils/reportUtils.test.ts
```diff
@@ -22,21 +23,82 @@ describe('reportUtils', () => {
...
           hint: {
-            text: 'A raw data extract providing placement matching outcomes for placement requests with an expected arrival within the month. Excludes PII.',
+            text: 'A raw data extract providing placement matching outcomes for placement requests with an expected arrival within the month.',
+          },
+        },
+      ])
+    })
+
+    it('should return a list of report options, including pii if user has permission', () => {
+      expect(reportOptions(fromPartial({ permissions: ['cas1_reports_view_with_pii'] }))).toEqual([
```
froddd, line 47: Same as above:

```suggestion
      const userWithPiiPermissions = userDetailsFactory.build({
        permissions: ['cas1_reports_view_with_pii'],
      })
      expect(reportOptions(userWithPiiPermissions)).toEqual([
```

froddd (changes requested):
Needs to use user details factory for tests, other than that looks good!

server/utils/reportUtils.test.ts
davidatkinsuk: Done! thanks for the proposed change, made it much easier 👍

froddd (approved):
Looks good!

PR #2368, opened 2025-02-24
APS-1988 Better differentiate report role names

davidatkinsuk:
Screenshot 2025-02-24 at 15 01 20
Screenshot 2025-02-24 at 15 04 03


froddd (approved):
👍

PR #2370, opened 2025-02-24
APS-2020 Daylight saving duration error

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/APS-2020

# Changes in this PR

The utility function  `differenceInDays()` of `date_fns` returns the number of whole days between dates, truncating any hour residue. When passed dates that span the autumnal daylight saving transition (BST -> GMT), one hour is removed from the duration and hence, the function returns one day fewer than it should.
The `date_fns` help recommends using the `differenceInHours` and dividing by 24 to correct this.

the bug was spotted on the Match confirmation page, but can occur wherever `differenceInDays` is used. I found other instances where this causes a problem after adjusting the unit tests to span 26 October 2025.

## Screenshots of UI changes

<details>
  <summary>Before</summary>

</details>

<details>
  <summary>After</summary>

</details>


froddd (approved):
That's a lot clearer, too!

Just one small comment about some unused parameterised test array items, which got me a tiny bit confused as I reviewed -- good to go after removing those.

PR #2393, opened 2025-03-04
APS-1935: placement change based on status

froddd:
# Context

https://dsdmoj.atlassian.net/browse/APS-1935

# Changes in this PR


server/testutils/factories/placementRequest.ts
```diff
@@ -1,49 +1,80 @@
...
+      booking: undefined,
+    })
+  }
+
+  unableToMatch() {
+    return this.notMatched().params({
+      status: 'unableToMatch',
+    })
+  }
+}
```
bobmeredith, line 42: Nice!

server/utils/bookings/index.ts
```diff
@@ -34,38 +34,10 @@ export const bookingSummaryList = (booking?: PlacementRequestBookingSummary): Su
...
-              text: 'Expected departure date',
-            },
-            value: {
-              text: DateFormats.isoDateToUIDate(booking.departureDate),
-            },
-          },
+          summaryListItem('Approved Premises', booking.premisesName),
+          summaryListItem('Date of match', DateFormats.isoDateToUIDate(booking.createdAt)),
+          summaryListItem('Expected arrival date', DateFormats.isoDateToUIDate(booking.arrivalDate)),
+          summaryListItem('Expected departure date', DateFormats.isoDateToUIDate(booking.departureDate)),
```
bobmeredith, line 40: This is so much better than all that bloat.

server/utils/placementRequests/adminIdentityBar.ts
```diff
@@ -26,33 +27,38 @@ export const adminActions = (
...
+    if (
+      placementRequest.booking.type === 'space' &&
+      ['upcoming', 'arrived'].includes(overallStatus(placementRequest.spaceBookings[0]))
+    ) {
+      matchedActions.push({
+        href: managePaths.premises.placements.changes.new({
+          premisesId: placementRequest.booking.premisesId,
+          placementId: placementRequest.booking.id,
+        }),
+        text: 'Change placement',
```
bobmeredith, line 51: I'm wondering why the difference in wording between the two types - 'Change' vs 'Amend' seems to indicate nothing! It would be nice if we could standardise on terminology e.g. 'booking' for a legacy booking and 'placement' for a `space-booking` but I know the terminology throughout the service is confused.

server/utils/placementRequests/placementSummaryList.test.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    expect(result.rows).toEqual(
+      expect.arrayContaining([
+        {
+          key: { text: 'Actual arrival date' },
+          value: { text: DateFormats.isoDateToUIDate(spaceBooking.actualArrivalDate) },
+        },
+      ]),
+    )
+    expect(result.rows).not.toEqual(
```
bobmeredith, line 48: I'm wondering why we don't show 'Expected arrival date' if there is an actual. For the CRU, I would have thought that they only really care about the expected - the booked dates. When the person actually turned up is more the concern of the AP.

server/utils/placementRequests/placementSummaryList.test.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    expect(result.rows).toEqual(
+      expect.arrayContaining([
+        {
+          key: { text: 'Actual departure date' },
+          value: { text: DateFormats.isoDateToUIDate(spaceBooking.actualDepartureDate) },
+        },
+      ]),
+    )
+    expect(result.rows).not.toEqual(
```
bobmeredith, line 72: Similarly for departure dates.

bobmeredith (approved):
Looks really good!
My only comments are about the language - and the logic of not showing the 'expected' dates on the placement summary if there are actuals. I would have though the CRU would want to see the actual dates they booked.

server/utils/placementRequests/placementSummaryList.test.ts
froddd: I think that's a very good point -- this was my call, not in designs, so I'll reinstate the 'expected' dates and only add the actual ones if available -- this should work?

PR #2394, opened 2025-03-04
APS-1963: Use updated response for Person Acct Alerts

froddd:
This swaps the deprecated `comment` for the new `description`, and renders 'Alert type' as the value of `alertTypeDescription` instead of `alertId`.

# Context

https://dsdmoj.atlassian.net/browse/APS-1963

# Changes in this PR

This swaps the deprecated `comment` for the new `description`, and renders 'Alert type' as the value of `alertTypeDescription` instead of `alertId`.


bobmeredith (approved):
LGTM

PR #2399, opened 2025-03-05
CHORE: Use upcoming booking by default for placement request factory

froddd:
This ensures a repeatable default for a property that has a bearing on the rendering of the UI itself. Any test requiring a booking of another status should instantiate it explicitly.

To ensure integrity of mock placement requests details, the booking properties (booking, legacyBooking, spaceBookings) are also now derived from the status provided: if 'notMatched' or 'unableToMatch', no linked booking is created.


bobmeredith (approved):
Makes sense.
LGTM

PR #2402, opened 2025-03-06
Use isArsonSuitable for arson risk on assessment

davidatkinsuk:
# Context

https://dsdmoj.atlassian.net/browse/APS-1902

# Changes in this PR

This commit effectively removes the ‘Is Designated Arson Room’ question (mapping to `isArsonDesignated`) and replaces it with an `Suitable for active arson risk` question (mapping to `isArsonSuitable`).

The new question continues to be prepopulated from the ‘Does the person pose an arson risk?’ response in the application

Any assessments that have been started before release of this code may have a value for isArsonDesignated and we should avoid losing this setting. 
As a temporary measure, when an assessment is loaded, if `isArsonSuitable` is not populated, it is copied from `isArsonDesignated`. This means that if the user uses the 'Matching information' or 'Check your answers' pages, the correct setting will be shown and if the assessment was complete apart from submission at the time of the update, isArsonSuitable would be populated at submission and hence carry the setting through to find and book.
  

## Screenshots of UI changes

No UI changes

### DIFF
--- a/server/controllers/assess/assessments/pagesController.test.ts
+++ b/server/controllers/assess/assessments/pagesController.test.ts
@@ -3,6 +3,7 @@ import { DeepMocked, createMock } from '@golevelup/ts-jest'
 import createError from 'http-errors'
 
 import type { DataServices, FormPages, TaskNames } from '@approved-premises/ui'
+import { Unit } from '@approved-premises/api'
 import PagesController from './pagesController'
 import { AssessmentService } from '../../../services'
 import TasklistPage from '../../../form-pages/tasklistPage'
@@ -18,6 +19,7 @@ import paths from '../../../paths/assess'
 import { viewPath } from '../../../form-pages/utils'
 
 import { assessmentFactory, clarificationNoteFactory } from '../../../testutils/factories'
+import * as matchingInformationUtils from '../../../form-pages/utils/matchingInformationUtils'
 
 const PageConstructor = jest.fn()
 
@@ -57,6 +59,9 @@ describe('pagesController', () => {
   describe('show', () => {
     const request: DeepMocked<Request> = createMock<Request>({ url: 'assessments' })
     const errorsAndUserInput = { errors: {}, errorSummary: [] as Array<never>, userInput: {} }
+    const mockedRemapArsonAssessmentData = jest
+      .spyOn(matchingInformationUtils, 'remapArsonAssessmentData')
+      .mockImplementation((assessmentData: Unit) => assessmentData)
 
     beforeEach(() => {
       ;(viewPath as jest.Mock).mockReturnValue('assessments/pages/some/view')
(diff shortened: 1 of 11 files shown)


server/controllers/assess/assessments/pagesController.ts
```diff
@@ -26,7 +27,13 @@ export default class PagesController {
     return async (req: Request, res: Response, next: NextFunction) => {
       try {
         const { errors, errorSummary, userInput } = fetchErrorsAndUserInput(req)
-        const assessment = await this.assessmentService.findAssessment(req.user.token, req.params.id)
+        const rawAssessment = await this.assessmentService.findAssessment(req.user.token, req.params.id)
+
+        // TODO: remove once arson remapping (APS-1876) is completed
```
davidatkinsuk, line 32: I think this can only be removed once all assessments that were created before APS-1876 goes live have been submitted (accepted or rejected?)

server/controllers/assess/assessments/pagesController.test.ts
```diff
@@ -18,10 +19,12 @@ import paths from '../../../paths/assess'
 import { viewPath } from '../../../form-pages/utils'
 
 import { assessmentFactory, clarificationNoteFactory } from '../../../testutils/factories'
+import { remapArsonAssessmentData } from '../../../form-pages/utils/matchingInformationUtils'
 
 const PageConstructor = jest.fn()
 
 jest.mock('../../../utils/validation')
+jest.mock('../../../form-pages/utils/matchingInformationUtils')
```
froddd, line 27: Can we use `jest.spyOn()` on the relevant method we're expecting to have been called instead of mocking the whole module? Mocking the whole module affects the rest of the test suite, and does not return actual implementation which makes it difficult to track other affected logic when/if some of the implementation itself changes.

server/controllers/assess/assessments/pagesController.ts
froddd: My understanding here is that `remapArsonAssessmentData()` is used to patch any old assessment _until_ APS-1876 has been done, in that it will not affect any assessment that is already using the new schema but will update any older one. Is this correct, @bobmeredith?

froddd (changes requested):
We need to move away from mocking whole modules wherever possible, as these mocks completely replace implementation and would hide any future updates to logic from other related parts of the codebase. Less relevant in this case, as this will be removed shortly, but should be done as a habit/best practice!

server/controllers/assess/assessments/pagesController.ts
davidatkinsuk: This PR. is part of APS-1876. The only remaining work on APS-1876 is 'fix' characteristics for assessments that have already been submitted. We're not planning on updating any assessment JSON for 'draft' assessments as part of APS-1876, so I think this remapping code will need to remain in place until we're sure that all assessments created before this change have been submitted.

server/controllers/assess/assessments/pagesController.test.ts
bobmeredith: Make sense

froddd (approved):
All good!

PR #2403, opened 2025-03-06
APS-2087: tidy up permissions

froddd:
# Context

https://dsdmoj.atlassian.net/browse/APS-2087

# Changes in this PR

This PR ensures all routes have the correct permissions to prevent unauthorised access. This completely removes reliance on user roles to determine authorisation, and instead relies fully on the more granular permissions.

- Ensure all routes have appropriate permissions set
- Move all user checks to use permissions rather than role
- Remove all uses of `workflow_manager` role
- Remove all uses of `matcher` role
- Refactor integration tests to use a consistent login strategy

## Screenshots of UI changes

n/a


integration_tests/tests/assess/shortNoticeAssessments.cy.ts
```diff
@@ -15,17 +15,15 @@ import AssessHelper from '../../helpers/assess'
...
     cy.task('reset')
-    cy.task('stubSignIn')
-    cy.task('stubAuthUser')
-  })
 
-  beforeEach(() => {
-    // Given I am logged in
-    cy.signIn()
+    // Given I am signed in as an assessor
+    signIn({ permissions: ['cas1_view_assigned_assessments', 'cas1_assess_application'] })
```
bobmeredith, line 25: I'm wondering if it's worth replicating the roles -> permissions mapping in a separate file. 
i.e. `assessorRole = ['cas1_view_assigned_assessments','cas1_assess_application']`
Then, we could preview any changes of the mapping in the API in one place, rather than having them scattered throughout the integration tests?

integration_tests/tests/signIn.ts
```diff
@@ -1,7 +1,7 @@
-import { ApprovedPremisesUserPermission } from '@approved-premises/api'
+import { ApprovedPremisesUser } from '@approved-premises/api'
 
-export const signIn = (roles: Array<string>, permissions?: Array<ApprovedPremisesUserPermission>) => {
+export const signIn = (user?: Partial<ApprovedPremisesUser>) => {
   cy.task('stubSignIn')
-  cy.task('stubAuthUser', { roles, permissions })
+  cy.task('stubAuthUser', user)
```
bobmeredith, line 5: Nice!

server/controllers/tasksController.test.ts
```diff
@@ -59,7 +59,7 @@ describe('TasksController', () => {
     beforeEach(() => {
       when(cruManagementAreaService.getCruManagementAreas).calledWith(token).mockResolvedValue(cruManagementAreas)
       when(taskService.getAll).calledWith(expect.anything()).mockResolvedValue(paginatedResponse)
-      when(userService.getUserList).calledWith(token, ['assessor', 'matcher']).mockResolvedValue(users)
+      when(userService.getUserList).calledWith(token, ['assessor', 'appeals_manager']).mockResolvedValue(users)
```
bobmeredith, line 62: Is the `appeals_manager` role a direct replacement for `matcher`?  They sound quite different!

server/routes/assess.ts
```diff
@@ -18,14 +18,24 @@ export default function routes(controllers: Controllers, router: Router, service
...
-  get(paths.assessments.index.pattern, assessmentsController.index(), { auditEvent: 'LIST_ASSESSMENTS' })
-  get(paths.assessments.show.pattern, assessmentsController.show(), { auditEvent: 'SHOW_ASSESSMENT' })
+  get(paths.assessments.index.pattern, assessmentsController.index(), {
+    auditEvent: 'LIST_ASSESSMENTS',
+    allowedPermissions: ['cas1_view_assigned_assessments'],
+  })
+
+  get(paths.assessments.show.pattern, assessmentsController.show(), {
+    auditEvent: 'SHOW_ASSESSMENT',
+  })
```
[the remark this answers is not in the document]
froddd: A base user ('applicant'?) can view links to individual assessments in a person's timeline, so adding any permission here would break this -- I'm not sure they _should_ be able to see this, but the above at least does not change current access for this.

server/routes/manage.test.ts
```diff
@@ -116,19 +114,17 @@ describe('manage routes', () => {
...
 
-  it('should allow a user with permission cas1 view out of service beds to to access the update out of service bed view', () => {
+  it('should allow a user with permission cas1 out of service beds create to access the update out of service bed view', () => {
     manageRoutes(controllers, router, services)
 
     expect(getSpy).toHaveBeenCalledWith(paths.outOfServiceBeds.update.pattern, updateOutOfServiceBedsController.new(), {
       auditEvent: 'SHOW_UPDATE_OUT_OF_SERVICE_BED',
-      allowedRoles: [],
-      allowedPermissions: ['cas1_view_out_of_service_beds'],
+      allowedPermissions: ['cas1_out_of_service_bed_create'],
```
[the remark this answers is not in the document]
froddd: These 2 permissions (`cas1_view_out_of_service_beds` and `cas1_out_of_service_bed_create`) are actually applied together to the CRU member and Future manager roles, so effectively interchangeable, but given this route is specifically to _update_ an OOSB it makes sense to use a more specific permission -- I think this could change in the future, where more users are allowed to view details of OOSB, but not to create/update them.

server/routes/placementApplications.ts
```diff
@@ -71,17 +73,22 @@ export default function routes(controllers: Controllers, router: Router, service
...
       },
     ],
   })
-  get(paths.placementApplications.review.confirm.pattern, placementApplicationReviewController.confirm())
+  get(paths.placementApplications.review.confirm.pattern, placementApplicationReviewController.confirm(), {
+    allowedPermissions: ['cas1_assess_placement_application'],
+  })
+
   get(paths.placementApplications.withdraw.new.pattern, placementApplicationWithdrawalsController.new())
   post(paths.placementApplications.withdraw.create.pattern, placementApplicationWithdrawalsController.create())
```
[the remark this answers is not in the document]
froddd: Yes, that's correct

server/utils/outOfServiceBedUtils.test.ts
```diff
@@ -74,56 +72,15 @@ describe('outOfServiceBedUtils', () => {
...
-          text: 'Ref number',
-        },
-        {
-          text: 'Manage',
-        },
-      ])
-    })
-
-    it('returns table headers for a user who is not a manager', () => {
-      const user = userDetailsFactory.build({ roles: [] })
```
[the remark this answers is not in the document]
froddd: The 'Manage' heading for the column is a misnomer, really, as this link is to view the details of an OOSB -- which I believe any user with the `cas1_view_out_of_service_beds` permission should be able to do, this permissions being necessary to view anything to do with OOSB. I'll change the column to 'View'.

However, the 'Update record' button on the individual OOSB detail screen seems to always show -- this should really be behind the `cas1_out_of_service_bed_create`, so I'll fix this.

Also the table headers above weren't showing anyway, due to template typo 🤦‍♂️  -- will also fix.

server/controllers/tasksController.test.ts
froddd: No, it isn't -- these changes are more to do with unreferencing the `matcher` role: the actual roles used in these instances don't matter (these tests only check presence/absence of any given role, and have nothing to do with actual permissions), so `matcher` was replaced with another role that will most likely continue to exist!

bobmeredith: Makes sense.  Thanks for the explanation.

PR #2406, opened 2025-03-07
APS-2078 Spellcheck on textareas

bobmeredith:
# Context
 
https://dsdmoj.atlassian.net/browse/APS-2078

# Changes in this PR
Adds spellcheck (or rather removes `spellcheck="false"`) throughout the system.  I went through apply/assess and identified all the text-areas that spellcheck was being disabled. I've no idea why this was done, but it seems to be deliberate. 
The affected text areas are listed in the Jira ticket.


froddd (approved):
Looks good code-wise, waiting to hear if we _should_ disable spell check on sensitive fields.

bobmeredith:
Dan has approved enabling spellcheck everywhere.

PR #2416, opened 2025-03-13
APS-2086: assess consistent validation

froddd:
# Context

https://dsdmoj.atlassian.net/browse/APS-2086

# Changes in this PR

Added validation as described in the ticket. 
Originally, I intended to make the additional info boxes conditional (as in apply) but the questions seem so complex that I am not sufficiently confident that the textareas would be shown in the right circumstances, so left as-is.

## Screenshots of UI changes

<details>
  <summary>Suitability assessment</summary>
  
</details>

<details>
  <summary>Required actionst</summary>
  
</details>

<details>
  <summary>Esap</summary>
  
</details>

### DIFF
--- a/server/form-pages/assess/assessApplication/suitablityAssessment/rfapSuitability.test.ts
+++ b/server/form-pages/assess/assessApplication/suitablityAssessment/rfapSuitability.test.ts
@@ -75,13 +75,27 @@ describe('RfapSuitability', () => {
 
       expect(page.errors()).toEqual({
         noDetail: 'You must provide details to support the decision',
-        unsuitabilityForRfapRationale: 'You must provide a summary of the rationale',
+        unsuitabilityForRfapRationale: 'Provide a summary of the rationale',
+      })
+    })
+
+    it('should show an error if the ESAP is identified as not suitable but the rationale box is not populated', () => {
+      const page = new RfapSuitability(
+        {
+          rfapIdentifiedAsSuitable: 'no',
+          noDetail: 'Some detail',
+        } as RfapSuitabilityBody,
+        assessment,
+      )
+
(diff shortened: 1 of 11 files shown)


froddd (commented):
Did you discuss the wording of the Location suitability assessment question as I suggested? It doesn't feel right at all, as all questions around it are of the type 'is the application of good quality', whereas this question is 'Are there factors to consider regarding the location of placement?' -- this seems like a question that only belongs to the application process.

This is fine as is, changing the question should be a separate matter!

gregkhawkins (approved):

server/form-pages/assess/assessApplication/suitablityAssessment/rfapSuitability.test.ts
```diff
@@ -51,6 +51,33 @@ describe('RfapSuitability', () => {
...
+          rfapIdentifiedAsSuitable: 'no',
+        } as RfapSuitabilityBody,
+        assessment,
+      )
+
+      expect(page.errors()).toEqual({
+        noDetail: 'You must provide details to support the decision',
+        unsuitabilityForRfapRationale: 'Provide a summary of the rationale',
+      })
+    })
```
gregkhawkins, line 80: do we need an extra case where we only get `unsuitabilityForRfapRationale` back? Preumably this would happen if this was the page setup:
```
const page = new RfapSuitability(
        {
          noDetail: 'test',
          rfapIdentifiedAsSuitable: 'no',
        } as RfapSuitabilityBody,
        assessment,
      )
```
 I don't have context on when the `noDetail` field is set so perhaps overkill :)

gregkhawkins (approved):

PR #2417, opened 2025-03-14
APS-2106 Out of service bed cancellation

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/APS-2106

# Changes in this PR

Adds (or rather restores) the OOSB cancellation journey adding a confirmation page.
 

## Screenshots of UI changes
image

image

image


server/utils/outOfServiceBedUtils.ts
```diff
@@ -111,7 +111,11 @@ export const outOfServiceBedActions = (
         {
           items: [
             {
-              text: 'Update record',
+              text: 'Cancel out of service bed',
```
froddd, line 114: Should there be any check on permissions here? I guess to be done when a new, separate `cancel OOSB` permission is introduced?

server/controllers/manage/outOfServiceBedCancellationController.ts
```diff
@@ -0,0 +1,51 @@
...
+      const outOfServiceBed = await this.outOfServiceBedService.getOutOfServiceBed(req.user.token, premisesId, id)
+
+      res.render('manage/outOfServiceBeds/cancel', {
+        pageHeading: 'Cancel out of service bed',
+        errors,
+        errorSummary,
+        errorTitle,
+        outOfServiceBed,
+        backLink: paths.outOfServiceBeds.show({ premisesId, bedId, id, tab: 'details' }),
+        selfLink: paths.outOfServiceBeds.cancel({ premisesId, bedId, id }),
```
froddd, line 24: I believe we've used `submitLink` in another place -- might be good to keep things consistent.

server/views/manage/outOfServiceBeds/cancel.njk
```diff
@@ -0,0 +1,42 @@
...
+{% from "moj/components/interruption-card/macro.njk" import interruptionCard %}
+{% extends "../../partials/layout.njk" %}
+
+{% set pageTitle = applicationName + " - " + pageHeading %}
+{% set mainClasses = "app-container govuk-body" %}
+
+{% block content %}
+<div class="govuk-grid-row">
+    <div>
+
```
froddd, line 11: These seem unnecessary -- and one of them is unclosed? The content is overflowing -- probably as a result of these.

server/views/manage/outOfServiceBeds/partials/_hiddenFields.njk
froddd, line 1: This isn't used anywhere?

froddd (commented):
The Interruption card works very well, it's a good improvement on existing design.

Some suggestions on tidying up the code, and the markup for the template needs looking at -- some unnecessary `<div>`s and the layout is overflowing the container.

integration_tests/tests/manage/future_manager/cancelsOutOfServiceBed.cy.ts
```diff
@@ -0,0 +1,76 @@
...
+    showPage.clickAction('Cancel out of service bed')
+
+    // Then I should be back on the confirmation page
+    Page.verifyOnPage(OutOfServiceBedCancelPage, outOfServiceBed)
+
+    // When I click Cancel out of service bed
+    cancelPage.clickDoCancel()
+
+    // Then the cancellation is sent to the API
+    cy.task('verifyOutOfServiceBedCancel', { premisesId, outOfServiceBedId: outOfServiceBed.id }).then(requests => {
```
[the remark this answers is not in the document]
bobmeredith: I was keeping consistent with the other verifications in this .cy file, but you're right - the more generic one is fine. I should refactor the others to change though.
The POST body is actually empty on this one, but I can verify that I guess. Should it use a 'DELETE' in fact?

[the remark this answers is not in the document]
froddd: It's a good point, it should probably be a `DELETE`, but I think I remember the API having this discussion and deciding to mostly keep to `POST` -- might be one to check with API devs again, as I have the feeling they'll want to retain `POST` -- which, incidentally, is less work all around!

server/controllers/manage/index.ts
```diff
@@ -28,7 +29,9 @@ export const controllers = (services: Services) => {
     services.sessionService,
   )
   const updateOutOfServiceBedsController = new UpdateOutOfServiceBedsController(services.outOfServiceBedService)
-
+  const outOfServiceBedCancellationController = new OutOfServiceBedCancellationController(
```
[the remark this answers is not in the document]
bobmeredith: OK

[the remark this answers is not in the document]
bobmeredith: OK. That does impact quite a few files though.

server/controllers/manage/outOfServiceBedCancellationController.ts
bobmeredith: OK, yes, there's just one usage in match - but let's stick with that.

server/views/manage/outOfServiceBeds/partials/_hiddenFields.njk
bobmeredith: True, I didn't need any hidden fields in this as it turns out.
It's quite a handy little macro for injecting hidden fields out of an object though - so may be useful at some stage.

server/utils/outOfServiceBedUtils.ts
bobmeredith: And it's there now!

froddd (approved):
Looks grand!

froddd (approved):

PR #2419, opened 2025-03-18
Enable V2 Match feature flag for rollout

froddd:
# Context

Rollout of V2 Match/Find and Book Beta to South West/South Central regions.


davidatkinsuk (approved):
LGTM 👍

PR #2424, opened 2025-03-19
Enable unable to match for all CRU members

froddd:
The 'Unable to match' functionality should be accessible to CRU members that are not part of the Beta -- this enables anyonre with the `cas1_booking_create` permission to access the feature.


bobmeredith (approved):
LGTM

PR #2427, opened 2025-03-20
APS-2087: Use role to permissions mapping from API

froddd:
# Context

https://dsdmoj.atlassian.net/browse/APS-2087

# Changes in this PR

This PR adds a step to copy the generated roles to permissions mapping from the API during the types generation, and updates all uses of the `signIn()` method in integration tests to revet to using role instead of granular permissions, for a closer to real-world usage.

## Screenshots of UI changes

n/a

### DIFF
--- a/server/utils/users/data/rolesToPermissions.json
+++ b/server/utils/users/data/rolesToPermissions.json
@@ -1,0 +1,186 @@
+[
+  {
+    "name": "assessor",
+    "permissions": [
+      "cas1_assess_appealed_application",
+      "cas1_assess_application",
+      "cas1_assess_placement_application",
+      "cas1_view_assigned_assessments"
+    ]
+  },
+  {
+    "name": "matcher",
+    "permissions": [
+      "cas1_assess_placement_application"
+    ]
+  },
+  {
+    "name": "future_manager",
+    "permissions": [
+      "cas1_out_of_service_bed_create",
+      "cas1_premises_view",
+      "cas1_premises_manage",
+      "cas1_space_booking_list",
+      "cas1_space_booking_record_arrival",
+      "cas1_space_booking_record_departure",
+      "cas1_space_booking_record_non_arrival",
+      "cas1_space_booking_record_keyworker",
+      "cas1_space_booking_view",
+      "cas1_view_out_of_service_beds"
+    ]
+  },
+  {
+    "name": "workflow_manager",
+    "permissions": [
+      "cas1_adhoc_booking_create",
+      "cas1_application_withdraw_others",
+      "cas1_booking_change_dates",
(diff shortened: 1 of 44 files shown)


server/utils/users/data/rolesToPermissions.json
```diff
@@ -0,0 +1,186 @@
...
+  {
+    "name": "change_request_dev",
+    "permissions": [
+      "cas1_appeal_create",
+      "cas1_planned_transfer_create"
+    ]
+  },
+  {
+    "name": "applicant",
+    "permissions": []
```
Rosie-Brigham, line 104: Just a sanity check - the applicant (PP) doesn't need specific permission to create, withdraw, view timeline etc?

froddd: No, this is effectively the default role: any user with access to the system is able to use Apply and Timeline. Some checks further down the line ensure users have more local permissions to submit applications or view timeline for a given CRN.

server/utils/users/data/rolesToPermissions.json
```diff
@@ -0,0 +1,186 @@
...
+    ]
+  },
+  {
+    "name": "matcher",
+    "permissions": [
+      "cas1_assess_placement_application"
+    ]
+  },
+  {
+    "name": "future_manager",
```
Rosie-Brigham, line 18: Maybe add a comment that this will be renamed in the future once legacy roles have been removed?

server/utils/users/data/rolesToPermissions.json
```diff
@@ -0,0 +1,186 @@
...
+      "cas1_application_withdraw_others",
+      "cas1_reports_view",
+      "cas1_reports_view_with_pii",
+      "cas1_request_for_placement_withdraw_others",
+      "cas1_appeal_create",
+      "cas1_planned_transfer_create"
+    ]
+  },
+  {
+    "name": "user_manager",
```
Rosie-Brigham, line 180: should this be `role_admin`? (just checking on the confluence)

server/utils/users/data/rolesToPermissions.json
froddd: This is covered in the Confluence page about roles and permissions -- besides, this file is copied over from the API, and also comments don't work well in JSON 😄

Rosie-Brigham (approved):
Look good - just a couple of questions where comparing the confluence page

server/utils/users/data/rolesToPermissions.json
froddd: Oooh good catch, it's the other way round -- the Confluence page needs updating. I'll do that now!

PR #2430, opened 2025-03-20
Update eslint

froddd:
# Context

https://dsdmoj.atlassian.net/browse/APS-2060

# Changes in this PR

Upgrade `eslint` from version `8.x.x` to version `9.x.x`.

The configuration is largely rewritten and now extends [the HMPPS config](https://www.npmjs.com/package/@ministryofjustice/eslint-config-hmpps) instead of relying on a variety of other third-party plugins.

Some updates to code as some rules are now a bit tighter around type declarations, and a lot of ignore statements have been removed (🥳).

## Screenshots of UI changes

No UI changes.


server/data/restClient.ts
```diff
@@ -156,7 +156,7 @@ export default class RestClient {
             reject(error)
           } else if (response) {
             const s = new Readable()
-            // eslint-disable-next-line no-underscore-dangle,@typescript-eslint/no-empty-function,no-empty-function
+            // eslint-disable-next-line no-underscore-dangle,no-empty-function
```
bobmeredith, line 159: Although you'll still need a disable line for the underscore, it might be nicer to add a comment inside the empty function and lose the 'no-empty-function' directive.

server/utils/azureAppInsights.ts
```diff
@@ -40,7 +40,7 @@ export function buildAppInsightsClient(name = defaultName()): TelemetryClient {
     defaultClient.addTelemetryProcessor(({ tags, data }, contextObjects) => {
       const operationNameOverride = contextObjects.correlationContext?.customProperties?.getProperty('operationName')
       if (operationNameOverride) {
-        tags['ai.operation.name'] = data.baseData.name = operationNameOverride // eslint-disable-line no-param-reassign,no-multi-assign
+        tags['ai.operation.name'] = data.baseData.name = operationNameOverride // eslint-disable-line no-multi-assign
```
bobmeredith, line 43: Suggest changing to // eslint-disable-next-line no-multi-assign on the line above to aid readability.

bobmeredith (approved):
Nice.
Just a couple of super minor suggestions.

server/data/restClient.ts
froddd: A comment explaining _why_ the function is empty would be good, but I'll be honest I'm not entirely sure I get this piece of code, so I can't add any insight! I'll leave this one as is, until it gets revisited with some more enlightenment added 😄

server/utils/azureAppInsights.ts
froddd: Good call!

PR #2433, opened 2025-03-25
APS-2152 isArsonDesignated in legacy assessment

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/APS-2152

# Changes in this PR

CAS1 uses a document mechanism to show the questions and answers as they were at the moment of submission of an application. There is no such feature for assessments.
As we have removed isArsonDesignated as a field, legacy assessments that have isArsonDesignated set in their data block, do not show a value for 'Designated arson room' but do show isArsonSuitable which gets defaulted to 'Not relevant'.

The response method on the MatchingInformationPage applies defaulting logic to its stored body on load.  This defaulting logic is useful to apply to the form, but shouldn't really be applied to the read-only view as that should show the content that was actually saved, rather than any update on it.  Hence I added a rawBody property to the page object to allow the construction of a non-defaulted response.

### Before

<details>
  <summary>Read-only view on a legacy assessment (where user selected isArsonDesignated as essential)</summary>

</details>

### After

<details>
  <summary>Read-only view on same legacy assessment as above</summary>

</details>

<details>
  <summary>New assessment - Matching information form</summary>

</details>

<details>
  <summary>Check-your-answers page on new assessment</summary>

</details>

<details>
  <summary>Read-only view on completed new assessment</summary>

</details>

### DIFF
--- a/server/form-pages/assess/matchingInformation/matchingInformationTask/matchingInformation.test.ts
+++ b/server/form-pages/assess/matchingInformation/matchingInformationTask/matchingInformation.test.ts
@@ -146,6 +146,31 @@ describe('MatchingInformation', () => {
       expect(response['Do you agree with the suggested length of stay?']).toEqual('No')
       expect(response['Recommended length of stay']).toEqual('5 weeks, 5 days')
     })
+
+    describe('a legacy assessment with isArsonDesignated defined', () => {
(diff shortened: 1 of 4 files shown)


froddd (approved):
LGTM

PR #2435, opened 2025-03-31
BUGFIX/APS-2172: dateutils BST issue

froddd:
# Context

https://dsdmoj.atlassian.net/browse/APS-2172

# Changes in this PR

The switch over to BST has uncovered a couple of bugs with the unit tests, and with the recording of arrivals and departure times. For arrivals, the time was derived from a date reconstructed from both date and time input by the user; this was resulting in the wrong time being set. For departures, the date and time were still sent as one ISO string -- the API is now able to receive separate date and time, similarly to the arrivals, so this has now been switched over tot he same mechanism.

A test for a date utility wasn't explicit with the timezone it was running in: this has also been fixed to ensure clarity and cover both timezones that happen in the UK.

## Screenshots of UI changes

n/a

### DIFF
--- a/server/controllers/manage/premises/placements/arrivalsController.ts
+++ b/server/controllers/manage/premises/placements/arrivalsController.ts
@@ -77,7 +77,7 @@ export default class ArrivalsController {
         }
 
         const placementArrival: Cas1NewArrival = {
-          arrivalTime: DateFormats.isoDateTimeToTime(arrivalDateTime),
+          arrivalTime,
           arrivalDate: DateFormats.isoDateTimeToIsoDate(arrivalDateTime),
         }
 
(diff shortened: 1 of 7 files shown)


server/controllers/manage/premises/placements/arrivalsController.ts
```diff
@@ -77,7 +77,7 @@ export default class ArrivalsController {
         }
 
         const placementArrival: Cas1NewArrival = {
-          arrivalTime: DateFormats.isoDateTimeToTime(arrivalDateTime),
+          arrivalTime,
           arrivalDate: DateFormats.isoDateTimeToIsoDate(arrivalDateTime),
```
davidatkinsuk, line 81: is this liable to issues if time is set between 00:00 and 00:59 during BST? I think isDateTimeToIsoDate is still forcing into UTC?

davidatkinsuk (approved):
Just one comment from me. As always, may be good for a typescript dev to give it a check too :)

Rosie-Brigham (approved):

server/controllers/manage/premises/placements/arrivalsController.ts
davidatkinsuk: Great 👍

PR #2436, opened 2025-04-01
BUGFIX: Ensure date inputs converted to local time

froddd:
This prevents an issue where the date derived from a set of date and time inputs ends up in UTC while the current timezone is BST, which could result in an unexpected result when checking if a datetime is in the past.

This commits adds a lot of fixtures to date utilities to ensure any timezone data from ISO strings or local dates is taken into account correctly when performing all the different formatting and validation. There is no change to implementation apart from the above.

Effectively, throughout the date utilities, any ISO with no timezone information (for instance, `2024-05-13T09:35`) is treated as a local date: in practice, this means that, during BST, it is considered to have an offset of an hour to UTC; when not in BST, there is no offset. Any ISO with timezone information (for instance, `2024-05-13T09:35:00.000Z` or `2024-05-13T09:35:00.000+01:00`) is treated as such, and converted appropriately. This ensures that, if a record is created in the API with a UTC `createdAt=2025-04-01T23:30:00.000Z` (which is during BST), the resulting date is shown as `2 Apr 2025, 00:30`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "typecheck": "tsc && tsc -p integration_tests",
     "test": "jest --collectCoverage",
     "test:debug": "jest --collectCoverage=false",
-    "test:ci": "jest --collectCoverage --runInBand --coverageThreshold=\"{}\"",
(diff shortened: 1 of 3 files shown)


davidatkinsuk (approved):
LGTM 👍

PR #2439, opened 2025-04-02
APS-2167: Add link to download CRU occupancy report

froddd:
# Context

https://dsdmoj.atlassian.net/browse/APS-2167

# Changes in this PR

This adds an action menu with a link to download the CRU occupancy report to the CRU dashboard page, for users who have the right permission.

## Screenshots of UI changes

<details><summary>CRU dashboard with actions menu open</summary>

Screenshot 2025-04-02 at 16 56 16

</details>


davidatkinsuk (approved):
LGTM 👍

gregkhawkins (approved):
LGTM! 

I added some e2e tests recently in `CAS3` for downloading reports so that we actually download the report (so it tests the API integration in dev/local for real when the e2es run). Here's where it is:
https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/blob/4c2d0bf6780be5aabb43722df5ff444708f2c232/e2e_playwright/steps/reports.ts#L186

This was modelled on the way that it had already been done in `CAS2` tbf so can't take the credit. Also, it's admittedly a little crude in the fact we just make sure the downloaded file has the expected header (i.e. we do not setup data and check the resulting file's data - only the header as mentioned). But it at least ensures that a file is downloaded for real and that the API integration is not broken. 

Anyways, obviously you've got good coverage for this through integration tests where you mock the file. Just thought worth mentioning for future reference :)

PR #2442, opened 2025-04-03
Migrate from CircleCI to GH Actions for security scans

aliuk2012:
Result of running
https://github.com/ministryofjustice/hmpps-github-actions/blob/main/docs/workflow-migration.md#automation


.circleci/config.yml.bak.20250403_120143
froddd, line 1: Don't think we should be committing this -- we can rely on git history for this

froddd (changes requested):
Looks good, but let's remove the config backup.

.github/workflows/security_trivy.yml
```diff
@@ -0,0 +1,12 @@
+name: Security trivy dependency check
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "48 7 * * MON-FRI" # Every weekday at 07:48 UTC
+jobs:
+  security-kotlin-trivy-check:
```
froddd, line 7: Not sure why this mentions `kotlin`, but given it's just a job id, no biggie

froddd (approved):
LGTM!

.github/workflows/security_trivy.yml
davidatkinsuk: suspect it's a copy paste thing from the original template? Maybe we should rename to security-trivy-check?

PR #2443, opened 2025-04-03
chore: Remove roles JSON cleanup command

froddd:
The roles to permissions JSON mapping from the API is now formatted in a more useful way for comparing diffs (https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/3281), so there is no need to run this step during the import as well.


davidatkinsuk (approved):
LGTM 👍

PR #2447, opened 2025-04-04
Setup .env with dev secrets as part of bootstrap

davidatkinsuk:
This will always use the ‘dev upstream’ configuration, and resolve any variables in the file to actual secret values


froddd:
Very nice!

PR #2448, opened 2025-04-07
Set the correct node version file for the scan to detect our node ver…

aliuk2012:
…sion

# Context

# Changes in this PR


froddd (approved):
👍

PR #2450, opened 2025-04-07
Ensure test waits for page load after click

froddd:
The stack is fast enough locally that the steps to check whether 'Breach or recall' or 'Move on' are present on the page always return true when those elements exist on the next page.

However, when running against the dev environment, the latency means that the timeout for checking those elements exist is often reached before the page has time to load, making the runner assume the elements simply do not exist.

This commit enforces a wait for the page load event before attempting to find those elements.


davidatkinsuk (approved):
LGTM 👍

PR #2456, opened 2025-04-11
Bootstrap e2e props into e2e.env and use in e2e runs

gregkhawkins:
# Context

# Changes in this PR


davidatkinsuk (approved):
very nice!

.env.template
```diff
@@ -1,31 +1,4 @@
 API_CLIENT_ID=${API_CLIENT_ID}
```
davidatkinsuk, line 1: is it worth adding a comment at the top of this saying that this file can be used to override default configuration defined in ap-tools?

package.json
```diff
@@ -24,8 +24,6 @@
     "test:e2e:ui": "npm run test:e2e:ci -- --ui",
     "test:e2e:local": "npx playwright test --config ./e2e/playwright.config.ts --project=local",
     "test:e2e:local:ui": "npm run test:e2e:local -- --ui",
-    "test:e2e:local-dev-upstream": "npx playwright test --config ./e2e/playwright.config.ts --project=local-dev-upstream",
-    "test:e2e:local-dev-upstream:ui": "npm run test:e2e:local-dev-upstream -- --ui",
```
davidatkinsuk, line 28: I think the README needs revising as it talks about testing upstream

Probbaly worth giving it a general refresh regarding overriding config from ap-tools?

PR #2469, opened 2025-04-17
APS-2160 Prevent manage actions on cancelled placement

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/APS-2160

# Changes in this PR

Adds a 'cancelled' logical status to placements. This means that the action menu on the placement details page is blank as all actions are dependent on specific statuses.

It's impossible to tell if a placement is cancelled from the `Cas1SpaceBookingSummary` hence there is a bit of a bodge in the status tests that test all statuses for both a `Cas1SpaceBooking` and `Cas1SpaceBookingSummary` but I thought it better to stick with the structure.

## Screenshots of UI changes
Both screenshots are of the same cancelled placement - as accessed through `timeline` from the CRU dashboard.

<details>
  <summary>Before</summary>

</details>

<details>
  <summary>After</summary>

</details>

### DIFF
--- a/server/testutils/factories/cas1SpaceBookingSummary.ts
+++ b/server/testutils/factories/cas1SpaceBookingSummary.ts
@@ -50,6 +50,15 @@ class Cas1SpaceBookingSummaryFactory extends Factory<Cas1SpaceBookingSummary> {
       isNonArrival: true,
     })
   }
+
+  cancelled() {
+    return this.params({
+      actualArrivalDate: undefined,
+      actualDepartureDate: undefined,
+      isNonArrival: false,
+      isCancelled: true,
+    })
+  }
 }
 
 export default Cas1SpaceBookingSummaryFactory.define(() => {
@@ -71,7 +80,7 @@ export default Cas1SpaceBookingSummaryFactory.define(() => {
     actualArrivalDate: arrivedStatuses.includes(status) ? canonicalArrivalDate : undefined,
     actualDepartureDate: status === 'departed' ? canonicalDepartureDate : undefined,
     isNonArrival: status === 'notArrived',
-    isCancelled: status === 'cancelled',
+    isCancelled: false,
     tier: faker.helpers.arrayElement(['A', 'B', 'C']),
     keyWorkerAllocation: cas1KeyworkerAllocationFactory.build(),
     characteristics: faker.helpers.arrayElements(Object.keys(roomCharacteristicMap)) as Array<Cas1SpaceCharacteristic>,
--- a/server/utils/placements/index.test.ts
+++ b/server/utils/placements/index.test.ts
@@ -39,7 +39,7 @@ describe('placementUtils', () => {
       const current = typeFactory.current()
       const departed = typeFactory.departed()
       const nonArrival = typeFactory.nonArrival()
-      const cancelled = cas1SpaceBookingFactory.cancelled()
+      const cancelled = typeFactory.cancelled()
 
       beforeEach(() => {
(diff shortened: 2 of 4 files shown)


server/testutils/factories/cas1SpaceBookingSummary.ts
```diff
@@ -50,6 +50,10 @@ class Cas1SpaceBookingSummaryFactory extends Factory<Cas1SpaceBookingSummary> {
       isNonArrival: true,
     })
   }
+
+  cancelled() {
+    return this.upcoming().params({})
+  }
```
bobmeredith, line 56: There is no way to indicate cancelled placement in the `Cas1PlacementSummary` now, so this is a bit weird.

server/utils/placements/index.test.ts
```diff
@@ -39,6 +39,7 @@ describe('placementUtils', () => {
       const current = typeFactory.current()
       const departed = typeFactory.departed()
       const nonArrival = typeFactory.nonArrival()
+      const cancelled = cas1SpaceBookingFactory.cancelled()
```
bobmeredith, line 42: Likewise, this gets round the fact that the test wouldn't pass on a `Cas1SpaceBookingSummary`

froddd: This feels like a smell! Maybe the API should return a `isCancelled` property instead? It's a bit more involved, but that's what happens for the `nonArrival` scenario, so it would make sense.

bobmeredith: True - that would be the right thing. Trouble is, currently the cancelled placements are filtered out of all of the lists so there is no material gain from doing the work, but it might be useful in the future. I'll see if the API guys are up for it.

bobmeredith: Should be febreezed now.

server/utils/placements/index.ts
```diff
@@ -53,9 +54,13 @@ type SpaceBookingStatus = keyof typeof statusTextMap
 const isSpaceBooking = (placement: Cas1SpaceBooking | Cas1SpaceBookingSummary): placement is Cas1SpaceBooking =>
   Boolean((placement as Cas1SpaceBooking).otherBookingsInPremisesForCrn)
 
+const isCancelled = (placement: Cas1SpaceBooking | Cas1SpaceBookingSummary): boolean =>
+  'isCancelled' in placement ? placement.isCancelled : Boolean(placement.cancellation)
+
 export const overallStatus = (placement: Cas1SpaceBookingSummary | Cas1SpaceBooking): SpaceBookingOverallStatus => {
   const isNonArrival = isSpaceBooking(placement) ? placement.nonArrival : placement.isNonArrival
 
+  if (isCancelled(placement)) return 'cancelled'
```
froddd, line 63: This could probably be replaced with a similar approach to the `notArrived` status below to keep things in one place (sorry, suggested change formatting is a bit whack: suggestion is to remove `isCancelled` method and simply rely on `isSpaceBooking` type guard to return the correct property, as is done with `isNonArrival`):

```suggestion
export const overallStatus = (placement: Cas1SpaceBookingSummary | Cas1SpaceBooking): SpaceBookingOverallStatus => {
  const isNonArrival = isSpaceBooking(placement) ? placement.nonArrival : placement.isNonArrival
  const isCancelled = isSpaceBooking(placement) ? placement.cancellation : placement.isCancelled

  if (isCancelled) return 'cancelled'
```

bobmeredith: Makes sense.

PR #2473, opened 2025-04-23
APS-2202 Transgender flow fix

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/APS-2202

# Changes in this PR

The original transgender logic has the NDelius 'sex' attribute being dominant if it's set to female. The requirement is that if the transgender flow is followed, the outcome of the special case board overrides the 'sex' attribute from NDelius.
I've included a check on `transgenderOrHasTransgenderHistory` so that if the transgender outcome is changed from 'yes' to 'no' in an application, any existing state of `shouldPersonBePlacedInMaleAp` is ignored.

I've also removed the old `ENABLE_WE` environment variable as it now set in all environments.


froddd (approved):
Perfect, covers all bases. Nice!

PR #2475, opened 2025-04-24
Fix reference to build step to get app_version

froddd:
For the 'Deploy to test' workflow, this was incorrectly set to `build` -- in this workflow, the build step is called `docker_build`.


bobmeredith (approved):
LGTM

PR #2480, opened 2025-04-30
APS-2139 Fix assessment duration override

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/APS-2139

# Changes in this PR

- Fixes the bug in the linked ticket. 
- In the assessment form, users are forced to enter a valid non-zero duration comprised of weeks or days. One of the input boxes can be left empty. Populated fields however must contain valid cardinal numbers. The 'day' box is not limited in magnitude so it's still possible to enter say '0' weeks and '54' days since it's possible that some users may prefer this and so the functionality should not be removed.
- Submission of the matchingDetails form no longer saves a `lengthOfStay` as a calculated day duration.
- The 'check your answers' page now displays the duration formatted from the as-entered `lengthOfStayWeeks` and `lengthOfStayDays`. 
- On submission, the day duration is calculated and written to a first-class field `duration`. This field drives the booking process.
- The error message on failing to enter valid numbers was updated. Also, there was a typo in the week/day input macro that prevented the error message from displaying so that was fixed also.
- Several tests failed because the test files mocked whole modules and, as a side effect, broke the functionality being tested. These have been refactored to leave non-mocked functions alone.

## Screenshots of UI changes

<details>
  <summary>Validation errors</summary>

</details>

### DIFF
--- a/server/form-pages/assess/matchingInformation/matchingInformationTask/matchingInformation.ts
+++ b/server/form-pages/assess/matchingInformation/matchingInformationTask/matchingInformation.ts
@@ -3,13 +3,14 @@ import type { SummaryList, TaskListErrors, YesOrNo } from '@approved-premises/ui
 import { ApprovedPremisesAssessment as Assessment } from '@approved-premises/api'
 import {
   defaultMatchingInformationValues,
+  lengthOfStay,
   suggestedStaySummaryListOptions,
 } from '../../../utils/matchingInformationUtils'
 import { DateFormats, daysToWeeksAndDays } from '../../../../utils/dateUtils'
 import { Page } from '../../../utils/decorators'
 
 import TasklistPage from '../../../tasklistPage'
-import { isCardinal, lowerCase, sentenceCase } from '../../../../utils/utils'
+import { lowerCase, sentenceCase } from '../../../../utils/utils'
 import {
   type ApTypeCriteria,
   type ApTypeSpecialist,
@@ -54,7 +55,6 @@ export type MatchingInformationBody = {
     'lengthOfStayAgreed',
     'lengthOfStayWeeks',
     'lengthOfStayDays',
-    'lengthOfStay',
     'cruInformation',
     ...placementRequirementCriteria,
     ...offenceAndRiskCriteria,
@@ -143,7 +143,7 @@ export default class MatchingInformation implements TasklistPage {
     response['Do you agree with the suggested length of stay?'] = sentenceCase(this.body.lengthOfStayAgreed)
 
     if (this.body.lengthOfStayAgreed === 'no') {
-      response['Recommended length of stay'] = DateFormats.formatDuration(daysToWeeksAndDays(this.body.lengthOfStay), [
+      response['Recommended length of stay'] = DateFormats.formatDuration(daysToWeeksAndDays(lengthOfStay(this.body)), [
         'weeks',
         'days',
       ])
@@ -183,8 +183,8 @@ export default class MatchingInformation implements TasklistPage {
     }
 
     if (this.body.lengthOfStayAgreed === 'no') {
-      if (!isCardinal(this.body.lengthOfStayWeeks) || !isCardinal(this.body.lengthOfStayDays)) {
-        errors.lengthOfStay = 'You must provide a recommended length of stay in whole weeks and days'
+      if (!(Number(lengthOfStay(this.body)) > 0)) {
+        errors.lengthOfStay = 'You must provide a recommended length of stay'
       }
     }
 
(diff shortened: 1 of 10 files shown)


server/utils/utils.test.ts
```diff
@@ -352,3 +353,18 @@ describe('joinWithCommas', () => {
     expect(joinWithCommas(list)).toEqual(expected)
   })
 })
+
+describe('isCardinal', () => {
+  it.each([
```
froddd, line 358: Probably worth trimming the input before testing — a space entered in a field before and/or after a number value should be handled fine.

server/form-pages/assess/matchingInformation/matchingInformationTask/matchingInformation.ts
```diff
@@ -182,8 +182,10 @@ export default class MatchingInformation implements TasklistPage {
       errors.lengthOfStayAgreed = 'You must state if you agree with the length of the stay'
     }
 
-    if (this.body.lengthOfStayAgreed === 'no' && !this.body.lengthOfStayWeeks && !this.body.lengthOfStayDays) {
-      errors.lengthOfStay = 'You must provide a recommended length of stay'
+    if (this.body.lengthOfStayAgreed === 'no') {
+      if (!isCardinal(this.body.lengthOfStayWeeks) || !isCardinal(this.body.lengthOfStayDays)) {
```
froddd, line 186: I’m not sure about requiring that _both_ fields are filled in. A common use case seems to be suggesting an 8-week placement instead of a 12-week one: in this case it seems like unnecessary friction to force the user to enter 0 for days.

Given we work out the resulting length in days from both fields, couldn’t the validation simply ensure the resulting length is more than `0` days, from any combination of weeks and/or days? The error message would still work fine. The `isCardinal` validation would still be useful, but only if any value has been entered (with the above suggestion to ignore whitespace).

server/views/components/formFields/form-page-day-weeks-input/macro.njk
```diff
@@ -24,7 +24,7 @@
     {% if context.errors[fieldName] %}
       <p id="{{ fieldname }}-error" class="govuk-error-message" data-cy-error-{{ fieldname }}="true">
```
froddd, line 25: Another `fieldname` with incorrect case here

server/form-pages/assess/matchingInformation/matchingInformationTask/matchingInformation.ts
froddd, line 1: It seems a bit odd that the calculation for the resulting `lengthOfStay` is done on submission only: could this be done in the body setter here to ensure the data is preemptively sanitised?

server/form-pages/assess/matchingInformation/matchingInformationTask/matchingInformation.ts
bobmeredith: Yes, I was thinking that.  This ia  bit belt'n'braces as this level of validation allows the existing downstream handling to be left untouched.
I'm now thinking it would be better to fix the handling so this field validation could be relaxed.

server/utils/utils.test.ts
bobmeredith: True. A tweak to the regexp would do that.

server/views/components/formFields/form-page-day-weeks-input/macro.njk
bobmeredith: Ouch. I guess that's a result of copy'n'paste.  This shows that we're lacking integration testing on error handling.

server/form-pages/assess/matchingInformation/matchingInformationTask/matchingInformation.ts
bobmeredith: Could be.  I guess one argument for this approach is that the result of the calculation can be shown on 'check-your-answers' and subsequently written into the document. If the response and submission handler used the same function though, I guess that would give more confidence.

froddd: What's shown on the 'Check your answers' and submitted as the document is the output of the `response` method for that page, so it could easily be done there. Equally, the `set body()` method could set the resulting total number of days straight away.

bobmeredith: Yes it does, and it means that in the JSON body, we have two versions of the truth which is bad. 
Never go to sea with two chronometers!
So, I've removed the `lengthOfStay` from the JSON block. The response (check your answers/document) calculate on the fly, as does the creation of the first-class field on submission.

froddd (approved):
Yup, that's a good one! 👍 Couple of leftover commented out bits that can be further cleaned up, everything else seems grand.

server/form-pages/utils/matchingInformationUtils.ts
```diff
@@ -219,7 +226,7 @@ const defaultMatchingInformationValues = (
       'essential',
       'notRelevant',
     ),
-    lengthOfStay: lengthOfStay(body),
+    // lengthOfStay: lengthOfStay(body),
```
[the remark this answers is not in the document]
bobmeredith: oops

PR #2481, opened 2025-05-02
APS-2267 - Booking changed timeline fixes

davidatkinsuk:
Whilst there was logic before this commit to render the booking changed timeline using structured data when the payload schemaVersion is null, the API was not populating the payload so the code was unused.

The API will start populating the payload in the near future using the same schema expected for V2 schema versions, allowing us to simplify the UI code

This PR is safe to go live immediately as the API doesn't currently return a payload for null schema version

Screenshots from when the API _is_ populating the payload for null schema version

# schema version null

Screenshot 2025-05-02 at 09 35 37

# schema version 2

Screenshot 2025-05-02 at 09 35 55


bobmeredith (approved):
Looks good to me.
Just a tiny comment on pluralization - feel free to ignore.

PR #2483, opened 2025-05-07
CHORE: remove unnecessary date utils

froddd:
The date-fns utilities `isBefore`, `isPast` and `isToday` are sufficient
for most date comparisons, so `datetimeIsInThePast` and `dateIsToday`
can be removed.

A single new utility is added, `dateIsPast`, as date-fns does not
provide for assessing whether a single date is in the past -- passing an
ISO date (with no time) to `isPast` returns true on teh same day, as the
comparison is made between now and the turn of the day.

## TODO
- [x] rebase on main once #2478 has been merged


bobmeredith (approved):
Nice one!
Tidier and much more readable.

PR #2487, opened 2025-05-09
APS-2276 - Add OOSB with PII Report

davidatkinsuk:
Add an option to download the Out of Service With PII Report if user has the report with PII role

Screenshot 2025-05-09 at 10 05 55


bobmeredith (approved):
LGTM

PR #2504, opened 2025-05-27
Update expected report filename

froddd:
All reports now follow the `{reportName}-{startDate}-to-{endDate}-{timestamp}.(csv|xlsx)` format.

Reports E2E run locally:

Screenshot 2025-06-11 at 11 08 46


froddd:
API seems to return various filename formats for the time being -- we can revisit when the UI sends a start and end date for reports rather than month and year.

vbala-moj (approved):

PR #2513, opened 2025-06-04
Chore: update non major dependencies

froddd:
This is a manual update of all non-major npm dependencies and some github actions version updates.

Renovate [attempted to update all non-major npm dependencies](https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/2146) but this caused some issues when running tests: this is due to some breaking changes and regressions introduced in `@golevelup/ts-jest` from version `0.5.1`. These are still not fully resolved (now on version `0.7.0`!), so for now we pin the version to `0.5.0`.

This also updates Github action version to simply major versions.


froddd:
Good point about the deprecations -- I'll have a look at these first.

froddd:
As discussed just now, let's merge this as is, and look at quashing Sass deprecation warnings at a later point.

PR #2515, opened 2025-06-04
Ensure minutes have 2 digits

froddd:
The previous change to this test step was stripping leading zeros from _every part_ of the datetime used to fill in the field. If the minutes where in the single digits (e.g. `15:09`), this was resulting in a time with a single minutes digit (`15:9`), which was falling foul of the validation.

This commit ensures the minutes are padded with zeros again, to ensure a double digit. Also, only the very first leading zero is removed from any datetime part, to ensure the hours do not end up empty if the time used is midnight.


e2e/pages/manage/recordArrivalPage.ts
```diff
@@ -12,13 +12,13 @@ export class RecordArrivalPage extends BasePage {
   }
 
   async recordArrival() {
-    const arrivalDateTime = addDays(new Date(), -1) // Yesterday same time
+    const arrivalDateTime = addDays(new Date('2025-06-04T00:00'), -1) // Yesterday same time
```
bobmeredith, line 15: I don't think this should be a literal. With the new validation it would fail after a week!

bobmeredith (commented):
I think that literal date-time needs to go

e2e/pages/manage/recordArrivalPage.ts
froddd: Absolute facepalm, this was for testing 🤦

bobmeredith (approved):
LGTM

PR #2528, opened 2025-06-10
Fix contract tests after removal of person risk e/p

bobmeredith:
# Context

No functional change.
Just removes an unused endpoint from person client/service so the contract tests pass.


davidatkinsuk (commented):
LGTM 👍

PR #2538, opened 2025-06-13
APS-2391 Support oasys 'hasApplicableAssessment'

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/APS-2391

# Changes in this PR
If the api response to `.../oasys/answers` and .../oasys/metadata` has the flag `hasApplicableAssessment` set to `false`, the UI:

- Sets `oasysSuccess` to false on the page - thus rendering the page the same as if a 404 had been received for the call.
- Loads the questions/answers from the response, rather than the `oasysStubs.json` file

If hasApplicableAssessment is true or not defined, the behviour is as before.
If the endpoint returns a 404, the `oasysSuccess` is set to false on the page and the questions/answers are loaded from `oasysStubs.json`

## Screenshots of UI changes

No change to UI - apart from the API flag causing the UI to render as it did for an API 404.

### DIFF
--- a/integration_tests/helpers/apply.ts
+++ b/integration_tests/helpers/apply.ts
@@ -239,7 +239,7 @@ export default class ApplyHelper {
     cy.task('stubPrisonCaseNotes404', { person: this.person })
   }
 
-  stubOasysEndpoints(excludeSection = false) {
+  stubOasysEndpoints(excludeSection = false, hasApplicableAssessment = true) {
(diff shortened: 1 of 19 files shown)


froddd (approved):
Looks good -- took me a while to figure out the reasoning behind setting `oasysSuccess` to false when the oasys retrieval was, in fact, successful, but within this context it seems to make sense.

PR #2539, opened 2025-06-16
APS-2082 AP-type on task lists

bobmeredith:
# Context

# Changes in this PR


froddd (approved):
Nice, can hardly spot the new functionality for all the cleaning up -- a good sign!

Code-wise this is great, although, as suggested elsewhere, better labels are needed. Maybe worth implementing the new set of labels now, so the content can be dropped in at a later point?

PR #2541, opened 2025-06-17
APS-2192 Change optional OASys question in E2Es

davidatkinsuk:
Previously the tests selected the optional ‘Accommodation’ section. This is no longer optional, so the tests now select ‘Relationships’ instead


bobmeredith (approved):
LGTM

PR #2544, opened 2025-06-17
BUGFIX/APS-2407: Prevent render error on `null` OASys section

froddd:
# Context

https://dsdmoj.atlassian.net/browse/APS-2407

# Changes in this PR

After removal of certain OASys sections, some application may end up with OASys section equal to `null`. This PR filters out such sections before rendering to prevent an error being thrown.

## Screenshots of UI changes

n/a


bobmeredith (approved):
LGTM

PR #2545, opened 2025-06-17
BUGFIX/APS-2408: Redact OASys answer for Sentry alert

froddd:
# Context

https://dsdmoj.atlassian.net/browse/APS-2408

# Changes in this PR

If an OASys summary is not defined, an alert containing all summaries for the OASys section is sent to Sentry. This PR ensures any user input (in the `answer` property of a summary) is redacted before the alert is created.

## Screenshots of UI changes

n/a

### DIFF
--- a/server/utils/oasysImportUtils.ts
+++ b/server/utils/oasysImportUtils.ts
@@ -117,8 +117,12 @@ export const findSummaryLabel = (questionNumber: string, summaries: Array<OASysQ
   let summary: OASysQuestion | undefined = summaries.find(i => i.questionNumber === questionNumber)
 
   if (!summary) {
+    const redactedSummaries = summaries.map(question => ({
+      ...question,
+      answer: question.answer ? '***redacted***' : undefined,
+    }))
     logToSentry(
-      `OASys summary not found for question number: ${questionNumber}. Summaries ${JSON.stringify(summaries)}`,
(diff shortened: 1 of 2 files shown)


server/utils/oasysImportUtils.ts
```diff
@@ -117,8 +117,12 @@ export const findSummaryLabel = (questionNumber: string, summaries: Array<OASysQ
   let summary: OASysQuestion | undefined = summaries.find(i => i.questionNumber === questionNumber)
 
   if (!summary) {
+    const redactedSummaries = summaries.map(question => ({
+      ...question,
+      answer: question.answer ? '***redacted***' : undefined,
```
bobmeredith, line 122: As an empty string is falsey, this would return `undefined` rather than the empty string. I don't know if that matters at all?

bobmeredith (approved):
Nice,
Just one, probably insignificant, comment.

server/utils/oasysImportUtils.ts
froddd: This is used in `JSON.stringify()`, so this results in the property not being included at all:

```js
JSON.stringify({ foo: undefined }) // returns '{}'
```

froddd: Oh, I see what you mean, sorry -- if there is an answer but it is empty, then the property won't be added at all... I wouldn't worry, the main thing here is to show 'there was some content to the answer but it is redacted'. I don't think not having the property for an empty answer is a worry, in that respect!

PR #2548, opened 2025-06-18
APS-1032: remove workflow manager role

froddd:
# Context

https://dsdmoj.atlassian.net/browse/APS-1032

# Changes in this PR

This PR fully removes the Workflow Manager role uses from the UI, and marks it as 'unused' so it can safely be removed from the API.

## Screenshots of UI changes

n/a

### DIFF
--- a/e2e/pages/admin/editUser.ts
+++ b/e2e/pages/admin/editUser.ts
@@ -7,7 +7,6 @@ export type Qualification = (typeof qualifications)[number]
 
 export const roles = [
   'Assessor',
-  'Workflow manager',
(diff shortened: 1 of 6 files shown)


bobmeredith (approved):
Looks good to me.
Nice and thorough.
Needs API update before merging though.

froddd:
This shouldn't need API update, as the role is moved to `unusedRoles` -- the API can continue specifying it, it just won't affect the UI in any way (or be shown anywhere). Once it is fully removed from the API, only the types will have to be updated -- which should allow for removal without affecting functionality.

PR #2553, opened 2025-06-23
APS-2221: Remove `aria-describedby` attribute when no error

froddd:
# Context

https://dsdmoj.atlassian.net/browse/APS-2221

# Changes in this PR

This ensures the document description textarea only has an `aria-describedby` attribute pointing to the error when there is an error for the field.

## Screenshots of UI changes

n/a


bobmeredith (approved):
LGTM

PR #2561, opened 2025-06-26
Chore: Cleanup Inactive Applications feature flag

froddd:
The `ENABLE_INACTIVE_APPLICATIONS_TAB` feature flag was only necessary to control launch of the feature, and should no longer be used. This PR completely removes the flag, ensuring the feature remains.


davidatkinsuk (approved):
LGTM 👍

PR #2565, opened 2025-06-27
APS-2401 Manage beds list view update

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/APS-2401

# Changes in this PR
A revamp of the manage beds view page.

- Removed the 'View' link column and put the details link on the bed name (first) column for consistency.
- Added sorting by bed name (AFAICS, the bed name is prefixed by the room name, so there's no point in )
- Added a list of bed characteristics on the table (I would like feedback as to whether this is needed/wanted given the filter. I think it's OK even for the worst premises but it could be made into a details expander list perhaps?)
- Added a filter for the bed characteristics
- Slight restructure of template around the filters and restructured a couple of other filters using these - we should probably switch all the filters over to get better consistency,
- Uncovered a bug in a test helper that tests column rows. This is only used in one place and couldn't handle the column restructure - so needed fixing
- While fixing the above, I found a few tiny issues with html rendering that inserts extra unnecessary spaces into mark-up. I left the fixes in place so had up update a few tests to get rid of the spare spaces.

Note, I added the class `govuk-list--compact` to the bed characteristics block in the table. This is not supported in this PR, but in https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/2547. It will remove the extra padding from that list.

### Before
<details>
   <summary>Before</summary>

</details>

### After
<details>
   <summary>After</summary>

</details>

### DIFF
--- a/server/controllers/manage/premises/bedsController.ts
+++ b/server/controllers/manage/premises/bedsController.ts
@@ -1,8 +1,21 @@
 import type { Request, RequestHandler, Response } from 'express'
 
+import { Cas1SpaceBookingCharacteristic } from '@approved-premises/api'
 import PremisesService from '../../../services/premisesService'
 import paths from '../../../paths/manage'
-import { bedActions, bedDetails, bedsActions, bedsTableRows } from '../../../utils/bedUtils'
+import {
+  bedActions,
+  characteristicsSummary,
+  bedsActions,
+  bedsTableHeader,
+  bedsTableRows,
+  calculateBedCounts,
+  filterBedsByCharacteristics,
+  generateCharacteristicsLabels,
+} from '../../../utils/bedUtils'
+import { convertKeyValuePairToCheckBoxItems } from '../../../utils/formUtils'
+import { makeArrayOfType, pluralize } from '../../../utils/utils'
+import { generateCharacteristicsSummary } from '../../../utils/premises/occupancy'
 
 export default class BedsController {
   constructor(private readonly premisesService: PremisesService) {}
@@ -12,19 +25,30 @@ export default class BedsController {
       const {
         user: { token },
         params: { premisesId },
+        query,
       } = req
+
+      const filterCharacteristics = makeArrayOfType<Cas1SpaceBookingCharacteristic>(query.characteristics)
 
       const [premises, beds] = await Promise.all([
         this.premisesService.find(token, premisesId),
         this.premisesService.getBeds(token, premisesId),
       ])
 
+      const filteredBeds = filterBedsByCharacteristics(beds, filterCharacteristics)
+
       return res.render('manage/premises/beds/index', {
         backLink: paths.premises.show({ premisesId }),
         premises,
         pageHeading: 'Manage beds',
         actions: bedsActions(premisesId, req.session.user),
-        tableRows: bedsTableRows(beds, premisesId),
+        tableRows: bedsTableRows(filteredBeds, premisesId),
+        tableHeader: bedsTableHeader(),
+        characteristicOptions: convertKeyValuePairToCheckBoxItems(
+          generateCharacteristicsLabels(calculateBedCounts(beds)),
+          filterCharacteristics,
+        ),
+        tableCaption: `Showing ${pluralize('bed', filteredBeds.length)}${generateCharacteristicsSummary(filterCharacteristics, ` that ${filteredBeds.length === 1 ? 'is' : 'are'} `)}`,
       })
     }
   }
@@ -45,7 +69,7 @@ export default class BedsController {
         premises,
         pageHeading: `Bed ${bed.name}`,
         actions: bedActions(bed, premisesId, req.session.user),
-        characteristicsSummaryList: bedDetails(bed),
+        characteristicsSummaryList: characteristicsSummary(bed.characteristics),
       })
     }
   }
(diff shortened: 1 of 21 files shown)


server/controllers/manage/premises/bedsController.ts
```diff
@@ -12,19 +24,29 @@ export default class BedsController {
...
       } = req
 
+      const filterCharacteristics = makeArrayOfType<Cas1SpaceBookingCharacteristic>(query.characteristics)
+
       const [premises, beds] = await Promise.all([
         this.premisesService.find(token, premisesId),
         this.premisesService.getBeds(token, premisesId),
       ])
 
+      const filteredBeds = filterBedsByCharacteristics(beds, filterCharacteristics as Array<Cas1SpaceCharacteristic>)
```
froddd, line 37: Can `filterCharacteristics` be instantiated directly as `makeArrayOfType<Cas1SpaceCharacteristic>(...)` to avoid this casting? It seems at this Premises manage level we shouldn't need to reference `Cas1SpaceBookingCharacteristic`.

server/controllers/manage/premises/bedsController.ts
```diff
@@ -12,19 +24,29 @@ export default class BedsController {
...
         backLink: paths.premises.show({ premisesId }),
         premises,
         pageHeading: 'Manage beds',
         actions: bedsActions(premisesId, req.session.user),
-        tableRows: bedsTableRows(beds, premisesId),
+        tableRows: bedsTableRows(filteredBeds, premisesId),
+        tableHeader: bedsTableHeader(),
+        characteristicOptions: convertKeyValuePairToCheckBoxItems(
+          generateCharacteristicsLabels(calculateBedCounts(beds)),
+          filterCharacteristics as Array<string>,
```
froddd, line 48: Don't think this needs to be cast as `Array<String>` either

server/utils/bedUtils.test.ts
```diff
@@ -40,55 +52,72 @@ describe('bedUtils', () => {
...
+      .buildList(5)
+      .sort(({ bedName: name1 }, { bedName: name2 }) =>
+        name1.localeCompare(name2, 'en', { numeric: true, sensitivity: 'base' }),
+      )
+
+    it('returns the rows of the beds table', () => {
+      const expected = bedList.map(bed => {
+        return [
+          htmlCell(
+            `<a href="/manage/premises/premisesId/beds/${bed.id}" data-cy-bedId="${bed.id}">${bed.bedName}<span class="govuk-visually-hidden">bed ${bed.bedName}</span></a>`,
```
froddd, line 66: Screen readers will read the main text content _and_ the hidden text, which means the content here is duplicated, as it is read as, for example, `1-1 bed 1-1` (this has been pointed out as a medium issue elsewhere in the accessibility audit). Ideally the `linkTo` utility should allow a hidden prefix _and/or_ suffix, so the hidden text could be `Bed name:` before the content. Or the `hiddenText` property could be ignored and the hidden element simply passed as part of the `text` argument? 

As an aside, after a quick look I'm not sure the `hiddenText` property is very judiciously used in many places! ☹️

server/utils/bedUtils.ts
```diff
@@ -21,23 +22,27 @@ export const bedsActions = (premisesId: string, user: UserDetails) =>
...
 
-export const actionCell = (bed: Cas1PremisesBedSummary, premisesId: string): TableCell => ({
-  html: bedLink(bed, premisesId),
-})
+  return sorted.map(bed => [
+    htmlCell(bedLink(bed, premisesId)),
+    textCell(bed.roomName),
+    htmlCell(characteristicsBulletList(bed.characteristics, { classes: 'govuk-list--compact' })),
+  ])
+}
```
froddd, line 35: Much nicer than those ultra-specific `bedNameCell` and `roomNameCell` classes. I've been sheepish about removing similar utilities elsewhere which I didn't find very useful; I should probably have gone ahead and got rid of them!

I'd even say we could probably get rid of `bedLink` -- if a utility is only calling another, existing one, without doing any more work, it's possibly unnecessary?

server/views/partials/filters.njk
froddd, line 1: This is very nice!

server/views/partials/filters.njk
```diff
@@ -0,0 +1,45 @@
+{% from "govuk/components/checkboxes/macro.njk" import govukCheckboxes %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
+{% from "govuk/components/select/macro.njk" import govukSelect %}
+
+{% macro filterWrapper(title) %}
+    <div class="search-and-filter">
+        <form action="{{ selfPath }}"
```
froddd, line 7: Is `selfPath` a property meant to be set for all controllers using this? Should it be `currentUrl` instead, which is set for all templates? 

https://github.com/ministryofjustice/hmpps-approved-premises-ui/blob/f2f64b1143a6a0082e1446bda6f66f9ec40cfcae/server/utils/nunjucksSetup.ts#L103

As it stands this results in an empty `action` attribute, which works the same way (submits the form to the current path), but it isn't clear why this variable should be required if that's what we're after!

froddd (commented):
Looks good -- love the work on wrapping filters, would be good to extend that as we go.

A few comments here and there!

server/views/partials/filters.njk
bobmeredith: It's a good point that maybe we should just submit to the same URL. In which case, simply omitting the action attribute would be the way to go.

PR #2566, opened 2025-06-27
updated generate-types job

daveawc:
this updates the generate types job to use a repository variable, to enable pulling spec file from dev environment


froddd (approved):
👍

PR #2570, opened 2025-06-30
Replace instances of `submitted' status

froddd:
# Context

https://dsdmoj.atlassian.net/browse/APS-866

# Changes in this PR

This PR replaces uses of the now-removed `submitted` status for applications. When an application is submitted, it now automatically gets the `awaitingAssesment` status (typo included), so most uses have been changed to that.


server/utils/match/occupancyCalendar.test.ts
bobmeredith, line 1: I don't think this change belongs in this PR? maybe the other ticket you're working on?

PR #2578, opened 2025-07-02
APS-2445 day summary capacity removal

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/APS-2445

# Changes in this PR

The single day occupancy view in manage currently makes three calls to the API. It fetches :

1. the premises
2. the day summary for that day - which contains a list of spaceBookings and OOSBs
3. the day capacity - which contains the counts of available and booked beds of each type

The day capacity is also returned as part of the day summary, but this is wasted. The secondary call to get the day capacity is needed because this endpoint supports excluding space-bookings whereas the day summary does not.

In manage however, the day occupancy page only retrieves the premises and to the day summary - and uses the capacity in that day summary.

In this PR, the manage day occupancy controller makes three requests, as in match, and uses only the capacity from the capacity response, so allowing the removal of `capacity` from the day summary.

## Screenshots of UI changes

No changes

### DIFF
--- a/server/utils/premises/occupancy.test.ts
+++ b/server/utils/premises/occupancy.test.ts
@@ -141,6 +141,7 @@ describe('apOccupancy utils', () => {
         : cas1PremiseCapacityForDayFactory.available().build({
             characteristicAvailability,
           })
+      return capacityForDay
     }
 
     it('should generate the text for an premises day with an overbooking on a single characteristic', () => {


server/controllers/manage/premises/apOccupancyViewController.ts
```diff
@@ -119,6 +119,11 @@ export default class ApOccupancyViewController {
         config.flags.pocEnabled ? characteristicsArray : undefined,
       )
 
+      const premisesCapacity = await this.premisesService.getCapacity(token, premisesId, {
```
froddd, line 122: Given this now makes 3 requests to the API, it would be good to run them in parallel -- I believe this shoudl be possible, as they're all based on the same request parameters?

PR #2582, opened 2025-07-07
APS-2430 accessibility-fix date field error link

bobmeredith:
# Context

# Changes in this PR


server/utils/nunjucksSetup.ts
```diff
@@ -116,10 +116,15 @@ export default function nunjucksSetup(app: express.Express, path: pathModule.Pla
       dev: true, // This is set to true to allow us to see the full stacktrace from errors in global functions, otherwise it gets swallowed and tricky to see in logs
     },
   )
-
+  njkEnv.addGlobal(
+    'callMacro',
+    function callMacro(macro: (context: Record<string, never>) => void, parameters: Record<string, never>) {
+      return macro({ ...this.ctx, ...parameters })
+    },
+  )
```
froddd, line 124: Context can be added to macros by importing them using ` with context`. So, in a template that uses a `dateGds` macro, you would import it like so:

```jinja
{% from "partials/dateGds.njk" import dateGds with context %}
```

This would then make the context available inside the macro, where it can be accessed with:

```jinja
{% set context = fetchContext() %}
```

Possibly cleaner than introducing this extra macro wrapper? 

Another possibly cleaner approach would be to repurpose `callMacro` as a wrapper for the parameters themselves:


```suggestion
  njkEnv.addGlobal(
    'addContext',
    (parameters: Record<string, never>) => ({ ...this.ctx, ...parameters })
  )
```

which could then be called in the template like so:

```jinja
{{ dateGds(addContext(params)) }}
```

froddd (commented):
This seems to work, but I don't like the approach at all -- it seems overly far-reaching. 

Have you explored only tweaking the error summary rendering to append the correct suffix for fields that require it, without having to change the field ID in so many places? I think this should be doable by either using a match on field name (e.g. ends with `Date`) or extending the error throwing so it can throw a bit of extra metadata about with the error itself (there's already an instance of adding the redirect URL with certain errors). 

To be discussed!

bobmeredith:
Yes, I thought about tweaking the summary, but couldn't see a reliable way to determine when and what to suffix. Parsing on 'Date' seems a bit error prone, particularly given the forthcoming switch over to date pickers.
The other way I considered was adding some client-side code to catch the focus on the wrapping div of a date field and focus onto the 'day' input. That would be minimal impact on the code, but requires client-side javascript.

froddd:
Javascript-based solution can't really be considered for obvious reasons.

The summary generation takes an object of errors of the form `Record<fieldname, string>`. This could be extended to include some metadata for each field, maybe `Record<fieldname, { message: string, type?: 'text' | 'date' }>`. We could then update the `ValidationError` type to also include some extra metadata containing which fields are to be considered date fields?

I just really don't like mapping the errors to the suffixed `-day` field name, it doesn't feel right. Should we need to start showing errors for each date part, or swap GOVUK date inputs for datepickers, we'll again come into rather large changes.

PR #2583, opened 2025-07-07
APS-1919: Use namespaced placement-request dashboard endpoint

froddd:
# Context

https://dsdmoj.atlassian.net/browse/APS-1919

# Changes in this PR

This PR remaps the dashboard endpoint for Placement Requests, `/placement-requests/dashboard`, to the new namespaced endpoint, `/cas1/placement-requests`. This changes the return type from `PlacementRequest` to `Cas1PlacementRequestDetail`, which affects a large amount of the codebase.

A new factory has been created to handle creating mocks of type `Cas1PlacementRequestSummary`. The `PlacementRequest`-type factory has been removed, with all functionality moved over to the `Cas1PlacementRequestDetail`-type one.

Also added 2 clean-up commits that were [originally earmarked for a separate branch](https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/2584), but it does make sense to bundle them here to reduce the amount of merging required to clean things up!

Screenshot 2025-08-05 at 14 59 47

## Screenshots of UI changes

No visual changes.

PR #2584, opened 2025-07-08
APS-1919: unused types cleanup

froddd:
# Context

https://dsdmoj.atlassian.net/browse/APS-1919

# Changes in this PR

This PR cleans up a lot of the types that are unused as a result of #2583, as well as a number of unused utilities.


froddd:
Rebased onto #2583

PR #2604, opened 2025-07-17
chore: Update MoJ Frontend dependency

froddd:
This latest version fixes a bug where a form that contain a datepicker cannot be submitted by pressing the Enter key when focusing on a field.


bobmeredith (approved):
Nice

PR #2606, opened 2025-07-17
APS-2520: Remove legacy booking references

froddd:
# Context

https://dsdmoj.atlassian.net/browse/APS-2520

# Changes in this PR

This PR removes any code related to 'legacy' bookings, as these are no longer used anywhere in the service.

As part of this removal, it was noted that booking confirmation emails still had links to the old URL for bookings, i.e. `/premises/{premisesId}/bookings/{bookingId}`. This has now been fixed in the API, and this PR adds redirects for both this URL and the v2 version (`/manage/...`) to the current placement page, `/manage/premises/{premisesId}/placements/{placementId}` -- this is possible as the booking IDs were retained during the migration.

Screenshot 2025-07-22 at 11 21 39

## Screenshots of UI changes

No UI changes


server/utils/bookings/index.test.ts
bobmeredith, line 1: Could the remaining functions in this file be moved to `outOfServiceBedUtils`. This filename doesn't seem right.

server/utils/bookings/index.ts
bobmeredith, line 1: Likewise

integration_tests/tests/manage/placements/viewPlacement.cy.ts
```diff
@@ -216,7 +216,7 @@ context('Placements', () => {
       // And there is an existing placement
       const { placement } = setup()
       // When I visit the placement page
-      // I should get an authorsation error
+      // I should get an authorisation error
```
bobmeredith, line 219: 😄

bobmeredith (approved):
Gordon Bennett, that's a big'ol PR!

Not much to say really.  Looks like there is still some scope for streamlining and the API could maybe rationalise and rename some types.
I can't say I looked too hard at all the deleted files, but if it builds and runs tests, it can't be too bad.

PR #2612, opened 2025-07-22
Use /cas1/assessments API instead of /assessments

davidatkinsuk:
# Context

# Changes in this PR


bobmeredith (approved):
Nice.
Much nicer using the `paths` definitions than manipulating strings.

PR #2615, opened 2025-07-23
chore: Remove unused templates

froddd:
These are not passed to any `response.render()` in any controller and can be removed.


bobmeredith (approved):
Nice!

PR #2622, opened 2025-07-29
APS-2249 National occupancy AP view

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/APS-2449

# Changes in this PR

Adds a single AP occupancy view page and a single premises single day view page as part of the national occupancy view.
The premises view has a new controller as there are too many differences between the page operating within match (in the context of a Placement request) and in the national view where there is no placement request. In this case, many of the calendar processing utilities are re-used and the calendar outer template is defined, but inheriting the majority of its functionality from the generic calendar template. The designs have not been followed in that the rendering of calendar days/cells match the national view since there should be consistency on successive pages used in conjunction.

The single day view is almost unchanged from the version in match, save for the lack of placement request context. A new route has been added to the existing controller and some of the controller logic tweaked to handle the case where the placement request id is not provided.

The connection of links is quite complex since the single premises/day view can be reached from either the national view (by clicking on a day) or from the single premises view. The back buttons should return the user to the page they accessed it from.  Also, the parameters of the search (room criteria and start date) are passed from the national view to the premises view in the session so that if the filter is updated in either view, the changes are persisted to the other. The duration selected in the premises view is also persisted in the session. This session object is not cleared so whenever the user returns to the national view within the same session, the last filter choices will be retained - although the results will not load automatically.

Also, a tweak to the rendering of cells in the national week view. The choice of single number or ratio should be dependent on whether the user has chosen any room criteria or not, and be independent of the `forRoomCharacteristic` field returned for each cell. 

## Screenshots of UI changes

<details>
  <summary>Single premises calendar view</summary>
   
</details>

<details>
  <summary>Single premises day view</summary>
   
</details>

### DIFF
--- a/integration_tests/pages/admin/nationaOccupancy/premisesOccupancyViewPage.ts
+++ b/integration_tests/pages/admin/nationaOccupancy/premisesOccupancyViewPage.ts
@@ -1,0 +1,66 @@
+import type { Cas1PremiseCapacity, Cas1Premises, Cas1SpaceBookingCharacteristic } from '@approved-premises/api'
+import Page from '../../page'
+import { DateFormats } from '../../../../server/utils/dateUtils'
+import paths from '../../../../server/paths/manage'
+import { dayAvailabilityCount, dayAvailabilityStatus } from '../../../../server/utils/match/occupancy'
+
+export default class PremisesOccupancyViewPage extends Page {
+  constructor(title: string) {
+    super(title)
+  }
+
+  static visit(premises: Cas1Premises): PremisesOccupancyViewPage {
+    cy.visit(paths.premises.occupancy.view({ premisesId: premises.id }))
+    return new PremisesOccupancyViewPage(`View spaces in ${premises.name}`)
+  }
+
+  static visitUnauthorised(premises: Cas1Premises): PremisesOccupancyViewPage {
+    cy.visit(paths.premises.occupancy.view({ premisesId: premises.id }), {
+      failOnStatusCode: false,
+    })
+    return new PremisesOccupancyViewPage(`Authorisation Error`)
+  }
+
+  shouldSeeValidationErrors() {
+    this.shouldShowErrorMessagesForFields(['arrivalDate'], {
+      arrivalDate: 'Enter a valid arrival date',
+    })
+  }
+
+  shouldShowCalendarKey(): void {
+    cy.get('#calendar-key').within(() => {
+      cy.contains('Available')
+      cy.contains('Full or overbooked')
+    })
+  }
+
+  shouldShowCalendar(
+    premisesCapacity: Cas1PremiseCapacity,
+    criteria: Array<Cas1SpaceBookingCharacteristic> = [],
+  ): void {
+    cy.get('#calendar').find('li').should('have.length', premisesCapacity.capacity.length)
+    cy.get('#calendar')
+      .find('li')
+      .each((day, index) => {
+        const capacity = premisesCapacity.capacity[index]
+        const { availableBedCount, date } = capacity
+        const bookableCount = dayAvailabilityCount(capacity, criteria)
+        const dayStatus = dayAvailabilityStatus(capacity, criteria)
+        const expectedClass = { overbooked: 'govuk-tag--red', full: 'govuk-tag--red', available: 'govuk-tag--green' }[
+          dayStatus
+        ]
+        cy.wrap(day).within(() => {
+          cy.contains(criteria.length ? `${bookableCount}` : `${bookableCount}/${availableBedCount}`)
+          cy.contains(DateFormats.isoDateToUIDate(date, { format: 'longNoYear' }))
+        })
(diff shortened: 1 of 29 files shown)


server/controllers/admin/nationalOccupancyController.test.ts
```diff
@@ -37,14 +46,23 @@ describe('NationalOccupancyController', () => {
...
+
   const errorsAndUserInput = createMock<ErrorsAndUserInput>({ errors: {}, errorSummary: [] })
 
   let nationalOccupancyController: NationalOccupancyController
 
-  const mockRequest = (query: Record<string, string | Array<string>> = {}, session: Record<string, string> = {}) =>
+  const mockRequest = (
+    query: Record<string, string | Array<string> | number> = {},
+    session: Record<string, string | Array<string> | number> = {},
+    params: Record<string, string> = {},
```
froddd, line 60: Could these types follow their actual implementation?

This seems to reveal a couple of inconsistencies with the session types, so would be worth tidying up, but would also remove the need for a couple of the castings further down.

```suggestion
    query: ParsedQs = {},
    session: NationalSpaceSearchFormData = {},
    params: ParamsDictionary = {},
```

server/controllers/admin/nationalOccupancyController.ts
```diff
@@ -134,4 +142,70 @@ export default class NationalOccupancyController {
...
+        })
+      }
+
+      const { arrivalDate: arrivalDateSlash, roomCriteria, durationDays } = { durationDays: 84, ...sessionData }
+      const roomCharacteristics = makeArrayOfType<Cas1SpaceBookingCharacteristic>(roomCriteria || [])
+
+      let calendar: Calendar
+      let premises: Cas1Premises
+
+      const arrivalDate = DateFormats.datepickerInputToIsoString(arrivalDateSlash as string)
```
froddd, line 168: Don't think `as` is needed here

integration_tests/pages/admin/nationaOccupancy/premisesOccupancyViewPage.ts
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  shouldShowCalendarKey(): void {
+    cy.get('#calendar-key').within(() => {
+      cy.contains('Available')
+      cy.contains('Full or overbooked')
+    })
+  }
+
+  shouldShowCalendar(
```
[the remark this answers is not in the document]
bobmeredith: Parameterising this does show up something interesting:

There are two entirely different ways to calculate the status of an AP for a day using the same capacity data.
In match and the national view where the intention is to create a booking, the state of the AP is based on a specified set of characteristics but in manage, it's based on all the returned characteristics.

For instance, if you want to see if you can place a person on a day who needs a wheelchair, you need to know that the overall capacity - overall bookings >= 1 and that the wheelchair spaces - wheelchair bookings >= 1.  You actually don't care if the AP is overbooked on single rooms - that's someone else's problem.
For AP managers though, when looking at a day, they want to know if overall capacity - overall bookings < 0 or capacity - bookings < 0 for _any_ characteristic as that means they have an overbook that needs action.

server/controllers/admin/nationalOccupancyController.test.ts
```diff
@@ -136,12 +157,12 @@ describe('NationalOccupancyController', () => {
     })
 
     it('should render the form with data merged from the session and query with query dominant', async () => {
-      const queryData = {
+      const queryData: { arrivalDate: string; apType: ApType } = {
         arrivalDate: '16/12/2025',
-        apType: Object.keys(apTypeLongLabels)[0],
+        apType: makeArrayOfType<ApType>(Object.keys(apTypeLongLabels))[0],
```
froddd, line 162: I have noticed this issue with TypeScript before: surely it should infer the type of the keys as `ApType` for `Object.keys()`, given the object itself is typed as `Record<ApType, string>`? I'm not sure why we have to be more explicit than that. 🤷‍♂️

integration_tests/pages/page.ts
```diff
@@ -814,4 +818,51 @@ export default abstract class Page {
 
     this.clearInput(name)
   }
+
+  shouldShowCalendarKey(type: 'twoColour' | 'threeColour'): void {
```
froddd, line 822: Is this technically the same as the `verbose` argument in `shouldShowCalendar`? In fact, this could probably be integrated into `shouldShowCalendar`, as I don't think we ever show a calendar without a key?

integration_tests/tests/admin/nationalOccupancy.cy.ts
```diff
@@ -180,7 +188,7 @@ context('National occupancy view', () => {
 
     THEN('I am on the AP view page again with all my settings unchanged')
     page.checkOnPage()
-    page.shouldBePopululatedWith({ durationText: 'Up to 1 week', arrivalDate: filterSettings.arrivalDate })
```
froddd: 😄

PR #2632, opened 2025-08-01
APS-2484 Restrict task allocation

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/APS-2484
https://dsdmoj.atlassian.net/browse/APS-2482

# Changes in this PR
This PR tackles two tickets as the required functionality for each is so closely related.

- Remove links to the task allocation page if:
  - the user doesn't have `cas1_tasks_allocate` permission 
  - or the task is completed (i.e on the completed tab)
- Restrict access to the task allocation page if the user doesn't have `cas1_tasks_allocate` permission.

## Screenshots of UI changes

<details>
  <summary>After</summary>

</details>

### DIFF
--- a/integration_tests/tests/tasks/allocations.cy.ts
+++ b/integration_tests/tests/tasks/allocations.cy.ts
@@ -22,11 +22,7 @@ context('Task Allocation', () => {
 
   const cruManagementAreas = [...cruManagementAreaFactory.buildList(4), cruManagementArea]
   const qualifications = qualificationFactory.buildList(2)
-  // Given there are some users in the database
-  const users = [...userWithWorkloadFactory.buildList(3), userWithWorkloadFactory.build({ qualifications })]
-  const selectedUser = users[0]
 
-  // And there is an allocated task
   const applicantUserDetails = applicationUserDetailsFactory.build()
   const caseManagerUserDetails = applicationUserDetailsFactory.build()
   const application = applicationFactory.build({
@@ -63,6 +59,10 @@ context('Task Allocation', () => {
   describe('when logged in as CRU member', () => {
     beforeEach(() => {
       cy.task('reset')
+      GIVEN('there are some users in the database')
+      const users = [...userWithWorkloadFactory.buildList(3), userWithWorkloadFactory.build({ qualifications })]
+
+      AND('there is an allocated task')
       cy.task('stubGetAllTasks', {
         tasks,
         allocatedFilter: 'allocated',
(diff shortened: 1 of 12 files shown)


integration_tests/tests/tasks/allocations.cy.ts
froddd, line 1: Might be worth implementing the `GIVEN/WHEN/THEN` on any touched integration test from now on?

integration_tests/tests/tasks/allocations.cy.ts
```diff
@@ -18,210 +18,220 @@ import { fullPersonFactory } from '../../../server/testutils/factories/person'
...
 
-  it('allows me to allocate a task', function test() {
-    cy.task('stubTaskAllocationCreate', {
-      task: { ...this.task, applicationId: this.application.id, allocatedToStaffMember: this.selectedUser },
-      reallocation: reallocationFactory.build({ taskType: this.task.taskType, user: this.selectedUser }),
+  const tasks = [task, taskWithRestrictedPerson]
+  describe('when logged in as CRU member', () => {
+    beforeEach(() => {
+      cy.task('reset')
+      // cy.task('stubSignIn')
```
froddd, line 66: Can be removed

server/controllers/tasksController.test.ts
froddd, line 1: There are a lot of superfluous assertions in this test suite, it's quite difficult to make sense of what each test is actually testing. It would be useful to clean this up a bit and only include the calls the tests are expecting to be different, and targeting changed template variables only with `expect.objectContaining()`.

froddd (approved):
Looks good.

I notice it's still technically possible to access the 'Allocate task' page by going to the URL directly, though without access to the database it would be quite tricky to figure out how -- not sure it's worth worrying about!

integration_tests/tests/tasks/allocations.cy.ts
```diff
@@ -18,210 +18,219 @@ import { fullPersonFactory } from '../../../server/testutils/factories/person'
...
-    const applicationForRestrictedPerson = applicationFactory.withReleaseDate().build({
-      apType: 'pipe',
-      person: restrictedPerson,
-      isWomensApplication: false,
-    })
+  const cruManagementArea = cruManagementAreaFactory.build()
+
+  const cruManagementAreas = [...cruManagementAreaFactory.buildList(4), cruManagementArea]
+  const qualifications = qualificationFactory.buildList(2)
+  // Given there are some users in the database
```
froddd, line 25: Missed a `GIVEN` and an `AND` 😄

froddd (approved):
Nice -- passing and overriding default render parameters works very well actually, and does cover all parameters so that's a nice bonus -- I know a couple of places where I will apply that in my current PR!

integration_tests/tests/tasks/allocations.cy.ts
bobmeredith: Ah no, they are outside the test 'it()' so a `cy.log` without the context, blows up!

PR #2634, opened 2025-08-04
APS-2622: Send requested placement duration as first-class

froddd:
# Context

https://dsdmoj.atlassian.net/browse/APS-2622

# Changes in this PR

This PR adds the requested placement duration as a first-class field on submission of the application. The utility to determine the default duration for a given application has been moved to the `applications` directory, as it is no longer explicitly used during the assessment process. This has resulted in the updated to a fair few import paths! The utility also now considers the `apType` for an application as optional: its requirement is only useful with regards to submission, where the value for `apType` is checked in a throwable manner anyway.

## Screenshots of UI changes

No UI changes.


froddd:
Very much so! 😄

PR #2646, opened 2025-08-12
APS-2449 National-occupancy - change single AP calendar rendering to match 'match'

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/APS-2449

# Changes in this PR

Changes the rendering of individual days in the single AP calendar view of the national occupancy view, to match the rendering of the calendar view in match.
Also, tightens the integration tests on said calendar; since the change initially didn't cause the integration tests to fail.

## Screenshots of UI changes

<details>
  <summary>Before</summary>

</details>

<details>
  <summary>After</summary>

</details>


server/views/admin/nationalOccupancy/partials/_singleApCalendar.njk
froddd, line 1: The contents of this file are now exactly the same as `server/views/match/placementRequests/occupancyView/partials/_occupancyCalendar.njk` -- this should probably be deleted.

froddd (changes requested):
There's now a fully duplicated template that should probably be deleted!

PR #2652, opened 2025-08-19
APS-2686: Validate duration weeks and days fields

froddd:
# Context

https://dsdmoj.atlassian.net/browse/APS-2686

# Changes in this PR

This introduces validation for the duration fields of the Placement duration and move-on page, which is shown when the assessor does not agree with the requested placement duration. The validation matches that already used on another such weeks/days set of inputs: both fields are cardinal numbers, or one is with the other empty.

## Screenshots of UI changes

Error when any field contains something else than a cardinal number:

Screenshot 2025-08-19 at 16 00 35

### DIFF
--- a/server/form-pages/apply/move-on/placementDuration.test.ts
+++ b/server/form-pages/apply/move-on/placementDuration.test.ts
@@ -7,6 +7,7 @@ import { applicationFactory } from '../../../testutils/factories'
 import { addResponsesToFormArtifact } from '../../../testutils/addToApplication'
 import { arrivalDateFromApplication } from '../../../utils/applications/arrivalDateFromApplication'
 import { DateFormats } from '../../../utils/dateUtils'
+import * as formUtils from '../../../utils/formUtils'
 
 jest.mock('../../../utils/applications/getDefaultPlacementDurationInDays')
 jest.mock('../../../utils/applications/arrivalDateFromApplication')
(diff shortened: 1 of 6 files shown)


server/form-pages/apply/move-on/placementDuration.test.ts
```diff
@@ -106,22 +106,44 @@ describe('PlacementDuration', () => {
...
+    //   expect(page.errors()).toEqual({
+    //     duration: 'You must specify the duration of the placement',
+    //   })
+    //
+    //   page = new PlacementDuration({ differentDuration: 'yes', durationDays: '1', reason: 'Some reason' }, application)
+    //
+    //   expect(page.errors()).toEqual({
+    //     duration: 'You must specify the duration of the placement',
+    //   })
+    // })
```
bobmeredith, line 146: Can go!

server/form-pages/apply/move-on/placementDuration.test.ts
```diff
@@ -106,22 +106,44 @@ describe('PlacementDuration', () => {
       })
     })
 
-    it('returns an error if the different duration response is yes but both durations are not set', () => {
-      let page = new PlacementDuration(
-        { differentDuration: 'yes', durationWeeks: '1', reason: 'Some reason' },
+    it.each([
```
bobmeredith, line 109: I'm thinking this is overkill as the same tests are applied to `validWeeksAndDaysDuration` All that really needs doing here is to test the generated error message.  Mock validWeeksAndDaysDuration perhaps?

bobmeredith (approved):
Nice.
I've made a couple of comments about unit tests.

server/form-pages/apply/move-on/placementDuration.test.ts
froddd: Oh of course -- I think I added it here first then realised it could be shared elsewhere. I'll get rid, and instead spy and check it's called as suggested.

froddd: I've updated the above now, and also updated the test for the other handler that uses the same validation -- that was also doing a whole lot of logic testing, which wasn't necessary.

PR #2656, opened 2025-08-21
APS-2717 - Use ‘submitted at’ for RfP made on date

davidatkinsuk:
Up until recently the request for placement ‘created at’ and ‘submitted at’ values were typically the same because the request for placement had to be created and submitted in the same session (they were lost if the user didn’t complete them)

Now we’ve started capturing initial request for placement details in the database as placement applications these dates can be very different e.g.

created at. - when the original application as created submitted at - when the original application was submitted

this could be days/weeks apart.

given this, when showing ‘request made on’ on the ‘requests for placements’ tab, we should submitted at instead


froddd (approved):
Good catch, nice!

PR #2669, opened 2025-09-12
APS-2755: Add sorting to booked date and premises columns

froddd:
# Context

https://dsdmoj.atlassian.net/browse/APS-2755

# Changes in this PR

This PR adds sorting to the 'Booked arrival date' and 'Approved Premises' columns on the 'Booked' tab of the CRU dashboard. It also defaults the tab to sort placement requests by 'Booked arrival date', ascending, by default.

## Screenshots of UI changes

<details><summary>'Booked' CRU dashboard with results sorted by booked arrival date by default</summary>

Screenshot 2025-09-12 at 16 22 27

</details>


server/controllers/admin/cruDashboardController.ts
```diff
@@ -53,7 +53,7 @@ export default class CruDashboardController {
 
       const {
         pageNumber = 1,
-        sortBy = 'expected_arrival',
+        sortBy = status === 'matched' ? 'canonical_arrival_date' : 'expected_arrival',
```
bobmeredith, line 56: Hmm, I got confused by the terms here, but it's quite correct. 
Sadly, we have an `expectedArrival` in `Cas1PlacementRequestDetail` but also have `expectedArrivalDate` in `Cas1SpaceBookingDetail`. They are very different things though.

bobmeredith (approved):
Nice

server/controllers/admin/cruDashboardController.ts
froddd: Indeed. The UI now does a decent job of showing the right wording (in this case the column is 'Booked arrival date'), but this is the property name as dictated by the API. AT least this only appears in the query string!

PR #2684, opened 2025-10-06
APS-2724 One application -> View page

bobmeredith:
# Context

# Changes in this PR


froddd:
The top block could easily be tailored to the various states:

- If assessed as suitable: 
  ```
  | Assessed as suitable: {date}
  | Application expires on: {date of expiry}
  |
  | [Create placement request]
  ```
- If rejected:
  ```
  | Assessed as not suitable: {date}
  ```
- If not yet assessed:
  ```
  | Application awaiting assessment
  ```

I don't think showing an empty assessment document works. When there is no assessment available, the tab should either be disabled or fully hidden, or the contents should at least be changed to something similar to the Placement request tab when there is no placement request, with wording adjusted to the current status ('awaiting assessment', etc):

Screenshot 2025-10-08 at 10 45 12

integration_tests/tests/apply/viewApplication.cy.ts
```diff
@@ -120,28 +123,28 @@ context('show applications', () => {
...
+    WHEN(`I click the 'PLacement request' tab`)
     cy.task('stubApplicationRequestsForPlacement', {
       applicationId: updatedApplication.id,
       requestsForPlacement: [],
     })
-    showPage.clickRequestAPlacementTab()
+    showPage.clickLink('Placement request')
 
-    THEN(`I do not see the 'Create request for placement' button`)
+    THEN(`I do not see the 'Create placement request' button`)
```
froddd, line 133: Given the button now appears before the tabs (if it does appear at all!), this assertion could be moved up -- there should be no need to click the placement request tab at all.

integration_tests/tests/apply/viewApplication.cy.ts
froddd, line 1: I think the test names could be a bit more consistent and descriptive in this suite, something along the lines of:

- shows an application awaiting assessment
- shows an application assessed as suitable
- shows an application that has been rejected
- shows an application that has expired
- shows an application that has been withdrawn
- shows an offline application
- etc

This suite assumes all the views are submitted, read-only applications, so there's probably no need to have 'read-only' in the titles.

server/utils/applications/adjacentPageFromSentenceType.ts
```diff
@@ -27,6 +27,6 @@ export const adjacentPageFromSentenceType = (sentenceType: SentenceTypeOption) =
     case 'life':
       return 'release-type'
     default:
-      throw new Error('The sentence type is invalid')
+      throw new Error(`The sentence type "${sentenceType}" is invalid`)
```
froddd, line 30: This seems out of place, but nice nonetheless 😄

server/utils/applications/utils.ts
```diff
@@ -379,14 +367,33 @@ const lengthOfStayForUI = (duration: number) => {
...
+  assessment: 'assessment',
   timeline: 'timeline',
   placementRequests: 'placementRequests',
 }
 
 export type ApplicationShowPageTab = keyof typeof applicationShowPageTabs
 
-export const applicationShowPageTab = (id: Application['id'], tab: ApplicationShowPageTab) =>
-  `${paths.applications.show({ id })}?tab=${applicationShowPageTabs[tab]}`
+export const applicationShowPageTab = (applicationId: string, tab: ApplicationShowPageTab): string => {
```
froddd, line 377: Seems fine as an arrow function, can lose the `{ return ... }`

server/utils/applications/utils.ts
```diff
@@ -379,14 +367,33 @@ const lengthOfStayForUI = (duration: number) => {
 
 export const applicationShowPageTabs = {
   application: 'application',
+  assessment: 'assessment',
```
froddd, line 370: I'm not sure this enum replacement is terribly useful -- it might work better to simply have the map of tab => labels and enforce that the tab is of type `ApplicationShowPageTab` in the controller?

server/utils/applications/utils.ts
```diff
@@ -379,14 +367,33 @@ const lengthOfStayForUI = (duration: number) => {
...
+  return `${paths.applications.show({ id: applicationId })}?tab=${tab}`
+}
+
+export const getApplicationShowPageTabs = (applicationId: string, tab: ApplicationShowPageTab) => {
+  const tabLabels: Record<ApplicationShowPageTab, string> = {
+    application: 'Application',
+    assessment: 'Assessment',
+    placementRequests: 'Placement requests',
+    timeline: 'Timeline',
+  }
```
froddd, line 387: See above comment about using this to define available tabs

froddd (commented):
Mostly fine, but I think we can move things further towards the ticket AC quite easily -- without having to revisit at a later stage.

PR #2877, opened 2026-03-25
FM-83 Personal contacts

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/FM-83

# Changes in this PR

Adds personal contacts.

- These are contained in the existing `caseDetail` object so no new API calls were needed.
- The contacts are split between Professional, Personal and other, using a hardcoded mapping in the UI. At some stage, we may want to pull this mapping from an API, but no such API exists in probation integration at the moment. 
- No email address is provided by the API. There appear to be very few email addresses populated in NDelius.
- Both mobile and landline phone numbers are shown if present. This deviates from NDelius which only shows landline numbers. Many mobile numbers are populated in NDelius, rather more than landlines in fact.
- Any unrecognised contact types are added to the 'Other' group.
- The original content of the page is shown in production until we can verify the behaviour of the page using real data in pre-prod.

## Screenshots of UI changes

<details>
  <summary>Personal contacts page</summary>
image
</details>

### DIFF
--- a/server/utils/resident/contactUtils.ts
+++ b/server/utils/resident/contactUtils.ts
@@ -1,0 +1,63 @@
+import { PersonalContact } from '@approved-premises/api'
+import { SummaryListWithCard, TableRow } from '@approved-premises/ui'
+import { htmlCell, textCell } from '../tableUtils'
+import { card } from './index'
+
+export type ContactGroup = 'PROF' | 'PERS' | 'OTH'
+
+export const contactGroupMapping: Record<string, ContactGroup> = {
+  CA: 'PROF',
+  MW: 'PROF',
+  CH: 'PERS',
+  RT04: 'OTH',
+  MC: 'PROF',
+  CE: 'OTH',
+  RT01: 'PROF',
+  ME: 'PERS',
+  OF: 'PERS',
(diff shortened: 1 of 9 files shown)


server/utils/resident/contactUtils.ts
```diff
@@ -0,0 +1,63 @@
+import { PersonalContact } from '@approved-premises/api'
+import { SummaryListWithCard, TableRow } from '@approved-premises/ui'
+import { htmlCell, textCell } from '../tableUtils'
+import { card } from './index'
+
+export type ContactGroup = 'PROF' | 'PERS' | 'OTH'
+
+export const contactGroupMapping: Record<string, ContactGroup> = {
```
AvraamK, line 8: Wonder if we can put this somewhere more central, but I guess it's fine here for now :)

AvraamK (approved):
LGTM! All good :)

server/utils/resident/contactUtils.ts
bobmeredith: It's specific to the resident profile at the moment, hence putting it here. PI are actually going to build an API to deliver the mapping so it will have to change anyway.

PR #2889, opened 2026-04-01
FM-398 Case notes - deploy

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/FM-398

# Changes in this PR

- Format case notes with body font size rather than small.
- Make visible in production

The embedded summary list component reduces the size of text to 0.9em. The designers requested that we use the normal bas font, and since this seems like it should be the default behaviour of `embeddedSummaryListItem()` I added a class parameter an changed the default to render an unchanged font, with a class to reduce the size.

## Screenshots of UI changes

<details>
  <summary>Before</summary>
image
</details>

<details>
  <summary>After</summary>
image
</details>


AvraamK (approved):
All good! :)

PR #2894, opened 2026-04-10
Updated "Back to top" jump links to behave like GDS site

robanderson-moj:
# Context

Based on additional feedback of the back of [FM-391](https://dsdmoj.atlassian.net/browse/FM-391)

Made the "Back to top" link behave the same way as on the GDS website: scroll down on [this page](https://design-system.service.gov.uk/components/footer/) and see the behaviour

Jump link appears when the side navigation is no longer on screen, and sticks to the bottom left until the footer appears when it sticks above the footer. The link takes the user to the very top of the page

After discussing with Beth, I've enabled it on all Manage a Resident pages, but it only appears if the content is long enough to be able to scroll until the side navigation no longer visible

I've made `backToTop` into a nunjucks macro so we can use it across other pages too

# Changes in this PR

https://github.com/user-attachments/assets/1defc422-f895-4a13-885e-ae7dbf81a35a

### DIFF
--- a/assets/js/backToTop.js
+++ b/assets/js/backToTop.js
@@ -1,0 +1,29 @@
+function backToTop(jumpLink) {
+  // browser doesn't support InteractionObserver, just use default behaviour
+  if (!('IntersectionObserver' in window)) {
+    return
+  }
+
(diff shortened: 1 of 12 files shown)


AvraamK (approved):
tested and all good!

PR #2904, opened 2026-04-21
Add preview text above details block

AvraamK:
# Context

Jira: https://dsdmoj.atlassian.net/browse/FM-372

# Changes in this PR

Added missing preview text in the risk page, above the now extended details block. 

### Before

<details>
  <summary>Before - details block closed</summary>
Screenshot 2026-04-21 at 16 08 06
</details>

<details>
  <summary>Before - details block expanded</summary>
Screenshot 2026-04-21 at 16 08 11
</details>

### After

<details>
  <summary>After - details block closed</summary>
Screenshot 2026-04-21 at 16 06 06
</details>

<details>
  <summary>After - details block closed</summary>
Screenshot 2026-04-21 at 16 06 13
</details>

### DIFF
--- a/server/utils/resident/riskUtils.ts
+++ b/server/utils/resident/riskUtils.ts
@@ -4,6 +4,7 @@ import nunjucks from 'nunjucks'
 import {
   card,
   detailsBody,
+  detailsBodyWithPreview,
   insetText,
   loadingErrorMessage,
   ndeliusDeeplink,
@@ -12,7 +13,7 @@ import {
 } from './index'
 import paths from '../../paths/manage'
 import { DateFormats } from '../dateUtils'
-import { ApiOutcome, splitFirstLineForDetails } from '../utils'
+import { ApiOutcome } from '../utils'
 import { htmlCell, textCell } from '../tableUtils'
 import config from '../../config'
 
@@ -163,9 +164,8 @@ export const registrationRows = (registrations: Array<Registration>): Array<Tabl
 
     if (registration.riskNotesDetail.length > 0) {
       const [{ note }] = registration.riskNotesDetail // As agreed we need only the first (latest) note to render
-      const { previewText, remainingText } = splitFirstLineForDetails(note)
       notesHtml = isOasysImportedFlag
-        ? detailsBody(`View full OASys notes for ${registration.description.toLowerCase()}`, remainingText, previewText)
+        ? detailsBodyWithPreview(`View full OASys notes for ${registration.description.toLowerCase()}`, note)
         : `<p class="govuk-body govuk-body__text-block">${note}</p>`
     }
 
--- a/server/utils/resident/riskUtils.ts
+++ b/server/utils/resident/riskUtils.ts
@@ -12,7 +12,7 @@ import {
 } from './index'
 import paths from '../../paths/manage'
 import { DateFormats } from '../dateUtils'
-import { ApiOutcome } from '../utils'
+import { ApiOutcome, splitFirstLineForDetails } from '../utils'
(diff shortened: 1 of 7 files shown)


server/utils/utils.ts
```diff
@@ -174,6 +174,20 @@ export const makeArrayOfType = <T>(input: unknown): Array<T> => {
 export const numberToOrdinal = (number: number | string): string =>
   ['First', 'Second', 'Third', 'Fourth', 'Fifth'][Number(number)]
 
+/**
+ * Splits teext for use with a details component by taking only the first line as preview text
```
bobmeredith: Little typo.

server/utils/utils.test.ts
```diff
@@ -342,6 +343,22 @@ describe('utils', () => {
...
 
+  describe('splitFirstLineForDetails', () => {
+    it('returns preview text and remaining text for multiline input', () => {
+      expect(splitFirstLineForDetails('first line\nsecond line\nthird line')).toEqual({
+        previewText: 'first line',
+        remainingText: 'second line\nthird line',
+      })
+    })
+
+    it('returns no preview text for single line input', () => {
```
bobmeredith: I'm wondering what should happen if there is one line, but terminated with a linefeed.
it looks like the code would show the line as a preview, but with an empty expander.

As I understand it, this is just input from a text box, so users are free to type whatever they like and add linefeeds wherever they want.

Having said that, I looked in pre-prod and all the cases I could find seem to have the same format so maybe it's nothing to worry about.

server/utils/utils.ts
```diff
@@ -174,6 +174,20 @@ export const makeArrayOfType = <T>(input: unknown): Array<T> => {
...
+ * and keeps the remaining lines as the expandable content.
+ */
+export const splitFirstLineForDetails = (text: string) => {
```
bobmeredith: This looks too specialised a thing to be a general utility. Maybe it should be in the risk utils.

server/utils/resident/riskUtils.ts
```diff
@@ -163,8 +163,9 @@ export const registrationRows = (registrations: Array<Registration>): Array<Tabl
 
     if (registration.riskNotesDetail.length > 0) {
       const [{ note }] = registration.riskNotesDetail // As agreed we need only the first (latest) note to render
+      const { previewText, remainingText } = splitFirstLineForDetails(note)
       notesHtml = isOasysImportedFlag
-        ? detailsBody(`View full OASys notes for ${registration.description.toLowerCase()}`, note)
+        ? detailsBody(`View full OASys notes for ${registration.description.toLowerCase()}`, remainingText, previewText)
```
bobmeredith: IMHO, it would be neater to do the splitting of the first line _inside_ the `detailsBody` function as that's surely the only reason you'd want it split.
Could either pass a boolean parameter, or have a different fn e.g. `detailsBodyWithPreview()`

bobmeredith (approved):
Seems to work OK with the data I can see in preprod although it might not handle arbitrary text very well.
I'd re-arrange things a bit though and put the splitter inside the details body fn.

server/utils/utils.ts
AvraamK: Funny thing I had it originally inside the riskUtils but I thought the exact opposite as maybe it can be used elsewhere as it is essentially a light extension of the details Component (which now looks like a "view more.." component).  I will move it back in though to keep everything in one place!

server/utils/resident/riskUtils.ts
AvraamK: fair point, moved!

PR #2907, opened 2026-04-29
Enable for prod ndelius risk flags

AvraamK:
# Context

Jira: https://dsdmoj.atlassian.net/browse/FM-372

# Changes in this PR

Removed the feature flags so that ndelius risk flags will show on prod


robanderson-moj (commented):
Has this previously been enabled in production using the feature flags?

If this has only been tested in preprod so far it might be worth leaving the feature flags in for a while until we have confidence it works in prod. Just saves us the bother of reverting and adding the feature flags back in if there are any issues

AvraamK:
> Has this previously been enabled in production using the feature flags?
> 
> If this has only been tested in preprod so far it might be worth leaving the feature flags in for a while until we have confidence it works in prod. Just saves us the bother of reverting and adding the feature flags back in if there are any issues

fair point! I restored the whole logic and just enabled them on prod!
will remove later once this works (it should work though as at least it works on pre-prod!)

robanderson-moj (approved):
👍 Looks good to me

PR #2908, opened 2026-05-01
FM-535 No valid OASys assessment handling

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/FM-535

# Changes in this PR
This feature is to improve the handling and reporting of missing or outdated OASys assessments on the risk page by removing all the OASys cards and replacing them with some content.

After investigation, it seems that in the case of a missing, or old assessment, the API returns valid structures to all requests and does not 404.
This PR therefore implements a check on the `/cas1/people/{crn}/oasys/answers` endpoint request for the `roshSummary` and checks the value of `assessmentMetadata.hasApplicableAssessment`. If false, the OASys sections are removed and replaced by the new content, otherwise normal handling is carried out.

This treatment is tentative as it's a best guess at what's returned in these circumstances. It needs to be checked in pre-prod for users that lack an up-to-date assessment.

## Screenshots of UI changes

<details>
  <summary>Error display</summary>
image
</details>

### DIFF
--- a/server/utils/resident/risk.ts
+++ b/server/utils/resident/risk.ts
@@ -1,10 +1,8 @@
 import { Cas1OASysGroup, PersonRisks } from '@approved-premises/api'
-import { card, insetText, subHeadingH2, subHeadingH3, TabData } from './index'
-import { DateFormats } from '../dateUtils'
+import { card, insetText, TabData } from './index'
 import { TabControllerParameters } from './TabControllerParameters'
-import { ndeliusRiskCards, riskOasysCards, roshWidget } from './riskUtils'
-import { linkTo, settlePromisesWithOutcomes } from '../utils'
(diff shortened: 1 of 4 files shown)


server/utils/resident/risk.ts
```diff
@@ -23,23 +21,23 @@ export const riskTabController = async ({
     personService.getOasysAnswers(token, crn, 'offenceDetails'),
     personService.riskProfile(token, crn),
   ])
-
+  // roshSummary.assessmentMetadata.hasApplicableAssessment = false
```
robanderson-moj: Can this be removed?

bobmeredith: it can indeed.

robanderson-moj (approved):
👍

PR #2919, opened 2026-05-12
Update OASys risk error messaging

AvraamK:
# Context

Jira: https://dsdmoj.atlassian.net/browse/FM-534

# Changes in this PR

If the latest OASys assessment is missing or is over 6 months old, OASys summary cards on Manage Resident pages now show "Not available from OASys" instead of being blank.

Pages: Health and disability, Mental health, Risk information, Offence and sentence, Drug and alcohol. 

Valid (< 6 months) assessments should render as before.

### DIFF
--- a/server/utils/resident/healthUtils.ts
+++ b/server/utils/resident/healthUtils.ts
@@ -11,7 +11,7 @@ import { SummaryListWithCard } from '@approved-premises/ui'
 import { card, cellMetaData, combineResultAndContent, insetText, loadingErrorMessage, ResidentProfileSubTab } from '.'
 import paths from '../../paths/manage'
 import { dateCellNoWrap, textCell } from '../tableUtils'
-import { oasysQuestionDetailsByNumber, summaryCards, tableRow } from './riskUtils'
+import { summaryCards, tableRow } from './riskUtils'
 import { ApiOutcome, linkTo } from '../utils'
 import { bulletList, summaryListItem } from '../formUtils'
 
@@ -124,19 +124,9 @@ export const healthDetailsCards = ({
 
   let cards = [card({ html: insetText(linkText) })]
 
-  const assessentIso = supportingInformation?.assessmentMetadata?.dateCompleted
-  if (assessentIso && assessentIso > '2025-04-09T18:00') {
-    const definition = oasysQuestionDetailsByNumber['13.1']
-    cards = cards.concat(
-      card({
-        title: definition.label,
-        html: `<p>We cannot load general health - any physical or mental health conditions right now.</p>
-<p>Go to OASys to check if any general health details have been entered.</p>`,
-      }),
-    )
-  } else {
-    cards = cards.concat(summaryCards(['13.1'], supportingInformation, supportingInformationOutcome))
-  }
+  cards = cards.concat(
+    summaryCards(['13.1'], supportingInformation, supportingInformationOutcome, { showUnavailableFromOasys: true }),
+  )
 
   cards = cards.concat([
     dietCard(dietAndAllergy, dietAndAllergyOutcome),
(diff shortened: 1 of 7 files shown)


server/utils/resident/healthUtils.ts
```diff
@@ -124,19 +124,9 @@ export const healthDetailsCards = ({
...
-        html: `<p>We cannot load general health - any physical or mental health conditions right now.</p>
-<p>Go to OASys to check if any general health details have been entered.</p>`,
-      }),
-    )
-  } else {
-    cards = cards.concat(summaryCards(['13.1'], supportingInformation, supportingInformationOutcome))
-  }
+  cards = cards.concat(
+    summaryCards(['13.1'], supportingInformation, supportingInformationOutcome, { showUnavailableFromOasys: true }),
+  )
```
bobmeredith: I think this is wrong - it should be left as-is.
This specific answer won't show because of a change to the OASys question structure - it's nothing to do with the 6 month cutoff.

server/utils/resident/healthUtils.ts
```diff
@@ -161,8 +151,10 @@ export const mentalHealthCards = ({
   supportingInformationOutcome: ApiOutcome
 }) => [
   card({ html: insetText('Imported from Digital Prison Service and OASys') }),
-  ...summaryCards(['FA62', 'FA63', 'FA64', 'R8.1.1', 'R8.2.1', 'R8.3.1'], riskToSelf, riskToSelfOutcome),
-  ...summaryCards(['10.9'], supportingInformation, supportingInformationOutcome),
+  ...summaryCards(['FA62', 'FA63', 'FA64', 'R8.1.1', 'R8.2.1', 'R8.3.1'], riskToSelf, riskToSelfOutcome, {
+    showUnavailableFromOasys: true,
+  }),
+  ...summaryCards(['10.9'], supportingInformation, supportingInformationOutcome, { showUnavailableFromOasys: true }),
```
bobmeredith: You don't need to parameterize the handling of errors I think. 
It will apply to all the OASys questions, apart from those on the risk page that are being hidden.

server/utils/resident/sentenceUtils.ts
```diff
@@ -123,7 +123,7 @@ export const sentenceCards = (caseDetail: CaseDetail, caseDetailOutcome: ApiOutc
 }
 
 export const oasysOffenceCards = (oasysAnswers: Cas1OASysGroup, callResult: ApiOutcome): Array<SummaryListWithCard> => [
-  ...summaryCards(['2.1', '2.12'], oasysAnswers, callResult),
+  ...summaryCards(['2.1', '2.12'], oasysAnswers, callResult, { showUnavailableFromOasys: true }),
```
bobmeredith: Again, can lose the parameter I think.

server/utils/resident/riskUtils.ts
```diff
@@ -128,23 +134,29 @@ export const summaryCards = (
...
+      const error =
+        !isUnavailableFromOasys &&
+        result &&
+        loadingErrorMessage(result, oasysGroupMapping[definition.groupName], 'oasys')
+      const html =
+        error ||
+        (isUnavailableFromOasys
+          ? oasysUnavailableMessage
+          : question &&
+            `${oasysMetadataRow(qNumber, oasysGroupMapping[definition.groupName], block)}${question.answer ? detailsBody('View information', question.answer) : '<p class="govuk-hint">Not entered in OASys</p>'}`)
```
bobmeredith: This can be a bit simpler I think. 
`error` is just a text field that gets displayed instead of the data if defined - so you could maybe just do something like this:
```
let error = block?.assessmentMetadata?.hasApplicableAssessment === false && oasysUnavailableMessage
error = error || result && loadingErrorMessage(result, oasysGroupMapping[definition.groupName], 'oasys')
```
then leave the rest of the code the same.

bobmeredith (commented):
As far as I can see, you don't need to parameterise as all the OASys questions can be treated the same.

server/utils/resident/healthUtils.ts
AvraamK: oh right yes! That was an accident, was looking at the test file for dates!

bobmeredith (approved):
Looks good to go.

PR #2921, opened 2026-05-15
Fix integration tests - fixed release date

bobmeredith:
# Context

# Changes in this PR

Fixes failing integration tests.

Some integration tests use a JSON fixture to build an application for the tests. 
This fixture has a hardcoded release date of 14 November 2026 which means that when tests were run after 14 May 2026, the application would fall inside the proper timeframe. As such, the assessment journey is altered and hence, the tests fail.

This PR updates the date loaded from the fixture to 7 months from the current date - as is done elsewhere.
The release date in the fixture has been left unchanged to prove that the mechanism works correctly.

[] I have run the E2E tests locally and they passed
```
Running 15 tests using 2 workers
  15 passed (4.8m)
```

## Screenshots of UI changes
None


davidatkinsuk (approved):
LGTM 👍

PR #2928, opened 2026-05-22
FM-604 Update pre-arrival risks

bobmeredith:
# Context
https://dsdmoj.atlassian.net/browse/FM-604

# Changes in this PR
These changes are controlled by the 'experimental' permission. Your user needs the experimental role to see them.

This PR adds a widget onto the risk tab to show information from the pre-arrival task list.

On the Risk tab, there is a new sub-tab 'Placement risks' that shows a risk widget containing the data from the pre-arrival task.
Initially, the widget displays a message with a link to the pre-arrival tasklist.

<details>
  <summary>Unpopulated widget</summary>
image
</details>

On 'complete' of the pre-arrival task-list, the data are copied from the pre-arrival form to a new form, currently called 'profile'. This blob is loaded on every page in the profile (since I expect there to be data on every page from it) and hence the completion of the pre-arrival task can be detected within the action menu.

The 'Placement risks' widget now shows the risk information enetered in the pre-arrival task.
<details>
  <summary>Populated widget</summary>
image
</details>

A 'Change' link on the widget opens a page, driven by the same page controller as handles the pre-arrival tasks, but detects that it's editing the data from the 'profile' journey instead of the pre-arrival journey and sets the save-redirect and backlinks accordingly. The page templates themselves are given access to the detected journey so that some content differences can be made - the 'where is this displayed' sections are removed in this case.

<details>
  <summary>Change page</summary>
image
</details>

The page follows the normal multi-page task flow but any data that are saved, are written to the 'profile' blob rather than the pre-arrival task blob.  Data are only saved if there are changes, so preventing the 'Last updated' time from being changed if a user simply chooses 'Save and exit' without altering anything.


integration_tests/tests/manage/placements/residentProfile.cy.ts
```diff
@@ -206,6 +211,7 @@ context('ResidentProfile', () => {
 
       THEN('the Risk tab should be selected')
       page.shouldHaveActiveTab('Risk')
+      page.shouldHaveActiveSideNav(`Risk information`)
```
robanderson-moj: Feel free to ignore, but backticks can just be regular single quotes here

robanderson-moj (approved):
I assume it wasn't a requirement, but there's no date validation on the pre arrival contact date currently

integration_tests/tests/manage/placements/residentProfile.cy.ts
bobmeredith: Cor blimey, that's a detailed bit of reviewing!
I'm surprised lint didn't automatically fix it.

bobmeredith:
> I assume it wasn't a requirement, but there's no date validation on the pre arrival contact date currently

True, there were no requirements around this, but it's a good point.

PR #2940, opened 2026-06-08
FM-506 Set suitability strategy for get oasys calls

davidatkinsuk:
For calls to oasys metadata & answers in `apply`, use the `completed_in_last_six_months` strategy.

For calls to oasys answers in `residents profile`, use the `allow_all` strategy.


bobmeredith (approved):
Nice one!
It's surprising just how many places OASys is being called from.

Let's hope all this is not all in vain! 😃

PR #2946, opened 2026-06-10
FM-724 Add error handling for submit placement application

davidatkinsuk:
The API is currently returning some specific ‘generic’ errors, although typically the UI stops these from ever happening (e.g. creating a placement application against an expired application, creating a placement application with no placement dates). If these occur the user will see a generic error page.

We have a requirement to add some validation in the API to catch a specific edge case (0 day placements) to minimise change required in the UI. This commit adds an error handler so we can present errors returned by the API to the user, instead of showing a generic error page.

The screenshot below shows a test message being rendered

Screenshot 2026-06-10 at 10 37 13


bobmeredith (approved):
LGTM

PR #2947, opened 2026-06-10
FM-724 Remove use of deprecated placement app fields

davidatkinsuk:
These fields on SubmitPlacementApplication are due to be removed from the API


robanderson-moj (approved):
👍 LGTM

PR #2966, opened 2026-06-24
FM-794 Harden actions/checkout and use hmpps renovate config

davidatkinsuk:


danhumphreys-moj (approved):
Sets `persist-credentials` to `false` in line with [this](https://github.com/actions/checkout/issues/485) recommendation and updates `renovate.json` in line with template https://github.com/ministryofjustice/hmpps-template-typescript/blob/main/renovate.json

PR #2983, opened 2026-07-07
FM-829 Post application document at every update

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/FM-829

# Changes in this PR

The document is now built and posted with every put to `/cas1/applications/{applicationId}`.
This should have been fairly simple, but I found a problem with rendering the document when the pages are not all properly completed that was highlighted by a failing integration test.
The page traversal mechanism worked as follows:

- Start with the first page (index 0 in the page index)
- If that page has data in the JSON blob, call the `next()` method on that page to find the next page.
- If that page does not have data, go to the next page in the index, or, if the last page in the index is reached, stop.

There is also a 'visited' mechanism that detects if the same page is visited twice(which would lead to an infinite loop) and throws an error.  This mechanism was being tripped during the integration test.

So, the problem is that if a populated page is reached by travelling forwards through the page list, skipping non-populated pages, then reaches a populated page that jumps backwards, the 'visited' mechanism would trip.

Since we are now rendering the document on all possible states of completion of that form, we cannot tolerate throwing on a loop, since there are certain user paths that actually include a loop.  Because of this, the task page sequencer has been alterered to:

1. skip initial non-populated pages until it finds the first populated page
2. terminate on hitting a visited page, rather than throw an error

Since this change _could_ affect the existing document output, I built a temporary test that used the old and new page sequencers and compared the list of pages calculated by the old algorithm with the new for the `applicationData.json` fixture.  From this, I updated the `applicationDocument.json` to match included a test that checks that the document renderer matches the document fixture, given the data fixture.

## Screenshots of UI changes

No UI change

### DIFF
--- a/integration_tests/fixtures/applicationDocument.json
+++ b/integration_tests/fixtures/applicationDocument.json
@@ -3,27 +3,27 @@
     {
       "If you need to update your details, select the relevant box below": "Phone number, Email address, Name, Area",
       "Applicant name": "Bob Smith",
-      "Applicant email address": "bob@test.com",
+      "Applicant email address": "bob@test.gov.uk",
       "Applicant phone number": "01234567890",
       "Applicant AP area": "Greater Manchester",
       "Do you have case management responsibility?": "No"
     },
     {
       "Case manager name": "Bob Smith",
-      "Case manager email": "bob@test.com",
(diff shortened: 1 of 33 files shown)


AvraamK (approved):
Nice! All good!

PR #2986, opened 2026-07-13
FM-830 Document render endpoint

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/FM-830

# Changes in this PR
Adds an endpoint that allows the generation of document objects from non-submitted applications to allow document backfill.

- The endpoint is available on `/render-application`
- it is not secured and does not require any token.
- The endpoint supports only POST requests and expects the body to contain a complete `Cas1Application` object. 
- In particular, the `data` field of the application should be populated with an application in a partly or fully completed state. 
- If the data field is not populated, the resulting document will simply be a list of section headings with empty contents for each. 
- If the document field is populated, it will be ignored.
- The response is a JSON document object.
- The JSON body parser limit has been increased to 1Mb from its default of 100Kb to allow applications to be posted.

To test a local instance, save a complete application in a file as JSON, then use the following curl command:
```
curl -H "Content-Type: application/json" --request POST --data '@application.json' http://localhost:3000/render-application
```

## Screenshots of UI changes

No UI changes


davidatkinsuk (approved):
Very neat! LGTM 👍

PR #2991, opened 2026-07-14
Document render endpoint - token protection

bobmeredith:
# Context

Following on from https://dsdmoj.atlassian.net/browse/FM-830
The ingress rule appears to block all access to the `/application-render` endpoint, so this ticket removes the rule and replaces it by using awt authentication.

The endpoint now expects to see a `Bearer` awt in the `authorization` header. This should have the role `ROLE_PROBATION_API__APPROVED_PREMISES__CASE_DETAIL`

- If no valid token is provided, a 401 response is returned.
- If the token does not have the required role, a 403 response is returned.


robanderson-moj (approved):
👍 Looks good as far as I can tell

PR #2994, opened 2026-07-20
FM-815 Placement duration calculation from API

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/FM-815

# Changes in this PR
Removes the internal UI calculation for default duration and replaces it with a call to the API.
The required placement duration is not stored in the Application unless overridden and hence, the default duration was being calculated whenever it was required. The calculation was a pure function, so this was quite straightforward.

For the calculation to be carried out in the API instead, all the pure function callls need to be async and hence, all the chain of calling functions need to be async too. Also, the application service ad token have to be passed down the function chain. These changes also flow through to quite an impact to the unit and end-to-end tests.

## Screenshots of UI changes

No UI change

### DIFF
--- a/integration_tests/tests/assess/assess.cy.ts
+++ b/integration_tests/tests/assess/assess.cy.ts
@@ -1,3 +1,4 @@
+import { DataServices } from '@approved-premises/ui'
 import { Cas1Assessment as Assessment } from '../../../server/@types/shared'
 import {
   assessmentFactory,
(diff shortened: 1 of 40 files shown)


integration_tests/tests/assess/assess.cy.ts
```diff
@@ -340,9 +342,9 @@ context('Assess', () => {
   it('does not invalidate the check your answers step if an answer is reviewed and not changed', function test() {
     // Given there is a complete application in the database
 
-    let assessment = assessmentFactory.build({ data: this.assessment.data, status: 'in_progress' })
+    // let assessment = assessmentFactory.build({ data: this.assessment.data, status: 'in_progress' })
```
robanderson-moj, line 345: Can this comment be removed?

server/form-pages/apply/move-on/placementDuration.ts
```diff
@@ -1,6 +1,6 @@
 import { addDays, weeksToDays } from 'date-fns'
-import type { PageResponse, TaskListErrors, YesOrNo } from '@approved-premises/ui'
-import { Cas1Application } from '@approved-premises/api'
+import type { DataServices, PageResponse, TaskListErrors, YesOrNo } from '@approved-premises/ui'
+import { Cas1Application as Application, Cas1Application } from '@approved-premises/api'
```
robanderson-moj, line 3: `Cas1Application` is import twice here

robanderson-moj (commented):
Couple minor things I noticed, but looks good 👍

bobmeredith:
No longer needed

PR #2997, opened 2026-07-21
FM-844 Use new createApplication endpoint

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/FM-844

# Changes in this PR

Switch application create to use new endpoint `/cas1/applications/create`

The new endpoint returns a rather stripped-down response from the old creation endpoint.  Rather than the full application, we only receive the tier and new applicationId.

Decisions are to be made regarding the next page the user is directed to. These are currently based on the person's `sex` and `tier`. Without the new `application` in the response, the controller now uses the `person` as retrieved immediately prior to the application creation, but substituting the `tier` as returned from application creation since it's possible that this is a new CRN and the tier may not have been available prior to application creation.
Methods that used to be passed the `Cas1Application`, have changed to take `person` instead.

Note that if the user immediately navigates away after application creation and returns later, they will be taken to the correct page as the correct (live) `tier` will now be returned in the person as part of the `Cas1Application` 

```
Running 15 tests using 1 worker
  15 passed (7.0m)
```

## Screenshots of UI changes
No UI change

### DIFF
--- a/integration_tests/mockApis/journeyUtils.ts
+++ b/integration_tests/mockApis/journeyUtils.ts
@@ -1,5 +1,10 @@
 import { SuperAgentRequest } from 'superagent'
-import { Cas1Application as Application, Cas1Assessment as Assessment } from '@approved-premises/api'
+import {
+  Cas1Application,
+  Cas1Application as Application,
+  Cas1Assessment as Assessment,
+  FullPerson,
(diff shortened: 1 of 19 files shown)


robanderson-moj (approved):
Apologies for taking so long, looks good 👍

PR #3002, opened 2026-07-24
Swap out deprecated ApprovedPremisesApplication type

bobmeredith:
# Context

# Changes in this PR

Replace all occurrences of the deprecated `ApprovedPremisesApplication` with `Cas1Application`
There are two locations where type is referenced that's not included in `Cas1Application` and these are 'bodged' until a type correction goes in.

## Screenshots of UI changes
No changes apart from type checking.


AvraamK (approved):
ooof quite a big one but yes you were right it is just a type change. nothing seems to be broken but the tests are the ultimate guide on this one! hopefully not so many merge conflicts, I just noticed this is open 1.5 months ago already! no one wanted to touch it for some reason 😄 
bear in mind I also didnt' run e2e tests but at this stage the unit ones should had picked any issues anyway

PR #3007, opened 2026-07-28
FM-928 Get placement duration from API and persist

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/FM-928

# Changes in this PR
based on https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/2994, This PR uses the API to calculate placement durations in two places.

- On the placement duration page, the duration is calculated by a call to the API.  The resulting `defaultDurationDays` and `maxDurationDays` are stored in the JSON data along with and `duration` override the user selects.
- When the application is submitted, the `defaultDurationDays` OR `duration` override is stored in the first-class `duration` field on the application.
- A mandatory error check has been added to the persisted duration field, on the `placementDuration` page. Hence, for in-flight applications in which the duration field has not been populated, the task will appear 'in-progress' and the user will not be able to submit. Simply by hitting the duration page again, the field will be populated and the task-list will show as 'completed'.
- When the application is assessed, the duration is read from the application duration field.
- On placementApplication creation, the duration is calculated using the API, and, using the sentence-type as selected by the user within the placementApplication flow, rather than just use the application logic.

## Screenshots of UI changes

No UI change

### DIFF
--- a/integration_tests/tests/assess/assess.cy.ts
+++ b/integration_tests/tests/assess/assess.cy.ts
@@ -1,4 +1,3 @@
-import { DataServices } from '@approved-premises/ui'
 import { Cas1Assessment as Assessment } from '../../../server/@types/shared'
 import {
   assessmentFactory,
@@ -91,7 +90,7 @@ context('Assess', () => {
       expect(requests).to.have.length(1)
 
       const body = JSON.parse(requests[0].body)
-      const expected = await acceptanceData(this.assessment, {} as DataServices, '')
+      const expected = acceptanceData(this.assessment)
       expect(body).to.deep.equal(expected)
     })
   })
@@ -341,8 +340,6 @@ context('Assess', () => {
 
   it('does not invalidate the check your answers step if an answer is reviewed and not changed', function test() {
     // Given there is a complete application in the database
-
-    // let assessment = assessmentFactory.build({ data: this.assessment.data, status: 'in_progress' })
 
     const assessment = addResponsesToFormArtifact<Assessment>(this.assessment, {
       page: 'application-timeliness',
--- a/integration_tests/tests/assess/assess.cy.ts
+++ b/integration_tests/tests/assess/assess.cy.ts
@@ -1,3 +1,4 @@
+import { DataServices } from '@approved-premises/ui'
 import { Cas1Assessment as Assessment } from '../../../server/@types/shared'
 import {
   assessmentFactory,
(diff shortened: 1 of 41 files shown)


integration_tests/tests/assess/assess.cy.ts
```diff
@@ -86,11 +86,12 @@ context('Assess', () => {
     assessHelper.completeAssessment()
 
     // Then the API should have received the correct data
-    cy.task('verifyAssessmentAcceptance', this.assessment).then(requests => {
+    cy.task('verifyAssessmentAcceptance', this.assessment).then(async requests => {
```
robanderson-moj: Does this need to be async?

server/form-pages/utils/matchingInformationUtils.test.ts
```diff
@@ -514,7 +514,7 @@ describe('matchingInformationUtils', () => {
       })
 
       describe('when the start date is not the same as the release date', () => {
-        it('returns the suggested stay from the application as summary list options, using the start date from the placement date screen', () => {
+        it('returns the suggested stay from the application as summary list options, using the start date from the placement date screen', async () => {
```
robanderson-moj: Again I'm not entirely sure why this has been made async

server/services/placementApplicationService.ts
```diff
@@ -73,12 +73,17 @@ export default class PlacementApplicationService {
     }
   }
 
-  async submit(token: string, placementApplication: PlacementApplication, application: Application) {
+  async submit(
+    token: string,
+    placementApplication: PlacementApplication,
+    application: Application,
+    dataServices: DataServices,
```
robanderson-moj: It feels a bit odd to be passing service classes into another service class, can the placement application submission data be fetched at an earlier point and be passed through rather than the service?

robanderson-moj (commented):
A couple of questions, mainly just that passing the data services into other services feels like a strange pattern

server/services/placementApplicationService.ts
bobmeredith: Yes, it sucks.
The issue is that what was originally a simple pure function to calculate the placement duration from the data in the application, has now become an async API request. I kept the structure the same, but it means that embedded in methods in the placementApplication service, there needs to be a call to the applicationService to get the duration.

I'll try unrolling the whole thing out into the controller and see where that gets me

bobmeredith: @robanderson-moj OK, maybe that looks a bit better. The worst impact was the unit tests though.  Auto-mocking whole modules is a pretty bad idea IMHO.
It does mean that there needs to be a test on `releaseType` outside the service, so as to know whether the duration needs fetching, but I guess that's a small price to pay if the code is more readable.

PR #3006, opened 2026-07-28
FM-960 - Show live tiers on ‘my applications’

davidatkinsuk:
# Context

https://dsdmoj.atlassian.net/browse/FM-960

# Changes in this PR

If feature flag ‘USE_LIVE_TIER’ is ‘true’, the person’s live tier value will be shown when listing applications across all tabs on the ‘my application screen’.

This commit also adds support for v3 tiers, using a different badge function, allowing us to use different style rules for v3 tiers

Also add missing coverage to dateCellSortable

### Before

Screenshot 2026-07-28 at 09 39 54

### After

when using v2 tiers (tier was manually changed in the `cases` table to differentiate from 'static' tier)

Screenshot 2026-07-28 at 09 24 00

when using v3 tiers

Screenshot 2026-07-28 at 09 30 25

PR #3010, opened 2026-07-30
Feature/fm 963 show live tier in application listing

SaintRhys:
# Context

[This](https://dsdmoj.atlassian.net/browse/FM-963) updates the assess table applications to use live tier.

# Changes in this PR

### Before
image

### After
image

With v3
image


server/utils/placementApplications/table.test.ts
```diff
@@ -94,7 +95,7 @@ describe('table', () => {
             html: stubLink,
           },
           crnCell(tasks[0]),
-          tierCell(tasks[0].tier?.value?.level),
+          htmlCell(getVersionedTierOrBlank(tasks[0].personSummary, tasks[0].tier?.value)),
```
davidatkinsuk: noting that i'm not a ui dev :) ...

in [my PR](https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/3006/changes) I added a `versionedTierCell`, but it doesn't seem that i used it anywhere. I think the intention was for this to be a drop in replacement for `tierCell`, but i didn't quite finish it.

Could that be enhanced to also take the fall back static tier value and used here instead?

e.g.

```
export const versionedTierCell = (tier: TierDto) => ({
  html: versionedTierBadge(tier),
  attributes: { 'data-sort-value': tierSortKey(tier.tierScore) },
})
```

becomes something like

```
export const versionedTierCell = (person: Person | PersonSummary, tierOnApplicationCreation: RiskTier) => ({
  html: getVersionedTierOrBlank(person, tierOnApplicationCreation),
  attributes: { 'data-sort-value': tierSortKey(tier.tierScore) },
})
```

and then used like

```
versionedTiercell(tasks[0].personSummary,tasks[0].tier?.value)
```

SaintRhys: Wasn't quite sure if you wanted it kept simple on purpose (should have asked!).
Didn't need the sort on these ones so I skipped it, but more than happy to adjust `versionedTiercell` now

davidatkinsuk: Sorry, slow response - Yeah i was wondering about the sort field, but as it was using `tierCell` already i guess that was (maybe?) intentional? @bobmeredith any strong opinions on this?

SaintRhys: I've switched to using an updated `versionedTierCell` as that would be good to use going forward

bobmeredith (approved):
Looks good to me.
The other tables should be simple now.

PR #3011, opened 2026-07-30
FM-863 Tier filter population

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/FM-863

# Changes in this PR

- Pulls the list of tiers used on the CRU dashboard/search tab, from an API reference data endpoint, rather than using a hardcoded list.
- The selector list is presented in the order received, i.e. the UI does no sorting.
- The tier filter field presented to the API `placement_request` endpoint is now switched depending on the state of the `USE_LIVE_TIER` feature flag. If enabled, the selected tier is passed in the `personTier` field, otherwise in the `tierOnApplicationCreation` field. I chose to keep the page search URL search parameter as `tier` rather than reflect the actual field names since those are not very 'user-friendly'.

## Screenshots of UI changes

No UI changes - the selector should be the same.

<details>
  <summary>Tier filter selector populated from API</summary>
image
</details>

### DIFF
--- a/integration_tests/tests/admin/searchPlacementRequests.cy.ts
+++ b/integration_tests/tests/admin/searchPlacementRequests.cy.ts
@@ -1,9 +1,11 @@
 import { PlacementRequestDashboardSearchOptions } from '@approved-premises/ui'
+import { AvailableTierDto } from '@approved-premises/api'
 import SearchPage from '../../pages/admin/placementApplications/searchPage'
 
 import { cas1PlacementRequestSummaryFactory } from '../../../server/testutils/factories'
 import { normaliseCrn } from '../../../server/utils/normaliseCrn'
 import { signIn } from '../signIn'
+import { AND, GIVEN, THEN, WHEN } from '../../helpers'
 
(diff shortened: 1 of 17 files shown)


server/@types/ui/index.d.ts
```diff
@@ -468,6 +468,8 @@ export type MiddlewareSpec = {
 export type PlacementRequestDashboardSearchOptions = {
   crnOrName?: string
   tier?: RiskTierLevel
```
SaintRhys: Could this be safely switched to just `string`?

integration_tests/tests/admin/searchPlacementRequests.cy.ts
```diff
@@ -102,22 +109,22 @@ context('Search placement Requests', () => {
       sortDirection: 'desc',
     })
 
-    // When I visit the search page
+    WHEN('I visit the search paget')
```
SaintRhys: Rogue 't' here

SaintRhys (approved):
Small comments but looks good to me!

PR #3012, opened 2026-07-30
Feature/fm 964 assess display live tier in banner

SaintRhys:
# Context
[This](https://dsdmoj.atlassian.net/browse/FM-964) adds tier into the assess banner.

I've mirrored how another banner handles tier and just use a plain tier rather than a badge. Unless we specifically want the badge styling?

### Before
image

### After
image

**V3**
image

### DIFF
--- a/server/utils/assessments/utils.ts
+++ b/server/utils/assessments/utils.ts
@@ -12,7 +12,7 @@ import { getApplicationType } from '../applications/utils'
 import { applicationAccepted, decisionFromAssessment } from './decisionUtils'
 import { formattedArrivalDate } from './dateUtils'
 import { getResponseForPage } from '../applications/getResponseForPage'
-import { displayName, getVersionedTier } from '../personUtils'
+import { displayName, getVersionedTierValue } from '../personUtils'
 import { DateFormats } from '../dateUtils'
 import applyPaths from '../../paths/apply'
 import assessPaths from '../../paths/assess'
@@ -144,7 +144,7 @@ const assessmentKeyDetails = (assessment: Assessment): KeyDetailsArgs => {
       },
       {
         key: { text: 'Tier' },
-        value: { text: getVersionedTier(assessment.application.person, assessment.application.risks?.tier?.value) || 'Not available' },
+        value: { text: getVersionedTierValue(assessment.application.person, assessment.application.risks?.tier?.value) || 'Not available' },
       },
       {
         key: { text: 'Arrival Date' },
--- a/server/utils/assessments/utils.ts
+++ b/server/utils/assessments/utils.ts
@@ -12,7 +12,7 @@ import { getApplicationType } from '../applications/utils'
 import { applicationAccepted, decisionFromAssessment } from './decisionUtils'
 import { formattedArrivalDate } from './dateUtils'
 import { getResponseForPage } from '../applications/getResponseForPage'
-import { displayName } from '../personUtils'
+import { displayName, getVersionedTier } from '../personUtils'
 import { DateFormats } from '../dateUtils'
 import applyPaths from '../../paths/apply'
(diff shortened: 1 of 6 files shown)


server/utils/personUtils.ts
```diff
@@ -31,6 +33,13 @@ const versionedTierBadge = (tier: TierDto): string => {
   return tierBadgeV3(tier.tierScore)
 }
 
+const getVersionedTier = (person: Person | PersonSummary, tierOnApplicationCreation?: RiskTier): string => {
+  if (!config.flags.useLiveTiers) {
+    return tierOnApplicationCreation?.level || ''
+  }
+  return personTier(person)?.tierScore || ''
+}
```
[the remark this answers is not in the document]
SaintRhys: I think having a shared helper would prove to be awkward and is fine this way

davidatkinsuk:
> I've mirrored how another banner handles tier and just use a plain tier rather than a badge. Unless we specifically want the badge styling?

that's correct, no need for styling here (FYI as part of the v3 switch we're removing styling from the tier badges too, but we'll handle that via tier v3 tickets)

server/utils/assessments/utils.ts
```diff
@@ -142,6 +142,14 @@ const assessmentKeyDetails = (assessment: Assessment): KeyDetailsArgs => {
         key: { text: 'CRN' },
         value: { text: assessment.application.person.crn },
       },
+      {
```
bobmeredith: There is a utility function `summaryListItem()` that  neatens this kind of stuff up quite a lot. You could use it for just this row or better still, refactor the whole banner and lose a few lines of code.

server/utils/personUtils.test.ts
```diff
@@ -311,4 +313,49 @@ describe('personUtils', () => {
...
+    })
+
+    describe('when live tiers are disabled', () => {
+      beforeEach(() => {
+        config.flags.useLiveTiers = false
+      })
+
+      it('returns the tier from application creation when present', () => {
+        const tierOnApplicationCreation = tierEnvelopeFactory.build({ value: { level: 'static', version: 'V2' } }).value
+        const personWithLiveTier = personFactory.build({ tier: tierDtoFactory.v2().build({ tierScore: 'live' }) })
```
bobmeredith: These test data are duplicated quite a bit. They could go into the outer describe perhaps.

server/utils/personUtils.test.ts
```diff
@@ -311,4 +313,49 @@ describe('personUtils', () => {
...
+      it('returns the live person tier when present', () => {
+        const tierOnApplicationCreation = tierEnvelopeFactory.build({ value: { level: 'static', version: 'V2' } }).value
+        const personWithLiveTier = personFactory.build({ tier: tierDtoFactory.v2().build({ tierScore: 'live' }) })
+
+        expect(getVersionedTierValue(personWithLiveTier, tierOnApplicationCreation)).toEqual('live')
+      })
+
+      it('returns an empty string when the person has no tier', () => {
+        const tierOnApplicationCreation = tierEnvelopeFactory.build({ value: { level: 'static', version: 'V2' } }).value
+        const personWithLiveTier = personFactory.build({ tier: undefined })
```
bobmeredith: `personWithLiveTier` is maybe not the best name given that they don't have a tier.

bobmeredith (approved):
Looks good!
Just a couple of comments but nothing important.

server/utils/personUtils.test.ts
SaintRhys: Shifted out

server/utils/assessments/utils.ts
SaintRhys: Updated, though left the bottom html one as is

PR #3015, opened 2026-07-31
Unsubmitted applcation create utility

bobmeredith:
# Context

Adds an npm/playwright script to create a new, fully populated application without submitting it.
This is useful to allow manual testing of application pages, without the hassle of walking through the whole of the application process.

Use as 
```
npm run create-application:standard:unsubmitted
```


davidatkinsuk (approved):
LGTM 👍

PR #3029, opened 2026-08-06
FM-1023 Fix application firstpage bug

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/FM-1023

# Changes in this PR

Fixes a long-existing bug in CAS1 that directs users to the manual risk entry page on returning to an application for a qualifying user, if they leave the apply process from the 'Confirm your details' page without submitting any information.

For further information, please see attached ticket https://dsdmoj.atlassian.net/browse/FM-1023.

## Screenshots of UI changes

No visible change

### DIFF
--- a/server/utils/taskListUtils.ts
+++ b/server/utils/taskListUtils.ts
@@ -1,5 +1,5 @@
 import { TaskStatus as TaskListStatus, TaskWithStatus, UiTask } from '@approved-premises/ui'
-import { Cas1Application as Application, Cas1Assessment as Assessment, FullPerson } from '../@types/shared'
+import { Cas1Application as Application, Cas1Assessment as Assessment } from '../@types/shared'
 import applyPaths from '../paths/apply'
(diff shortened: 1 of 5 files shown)


server/utils/taskListUtils.ts
```diff
@@ -20,7 +20,8 @@ export const taskLink = (
           task: task.id,
           page: firstPage || '',
         })
-      : (task.id === 'basic-information' && tierQualificationPage(applicationOrAssessment)) ||
+      : (task.id === 'basic-information' &&
+          firstPageOfApplicationJourney(applicationOrAssessment.id, applicationOrAssessment.person as FullPerson)) ||
```
SaintRhys: Could you change `firstPageOfApplicationJourney` to just accept `Person`, then have the `isFullPerson` deal with if it's a `FullPerson` or not, then you could avoid the cast here

SaintRhys (approved):
Nice! One small comment but good to go

PR #3034, opened 2026-08-07
FM-983 is-exceptional-case page content updates

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/FM-983

# Changes in this PR
Updates the content and error message on the 'is-exceptional-case' page.
The check-your-answers page and document are now populated using the radio-button question, rather than the page title as-was. 

## Screenshots of UI changes

<details>
  <summary>Before</summary>
image

</details>

<details>
  <summary>After</summary>
image
<hr/>
<h3>Errors</h3>
image

</details>

### DIFF
--- a/integration_tests/tests/apply/apply.cy.ts
+++ b/integration_tests/tests/apply/apply.cy.ts
@@ -22,12 +22,11 @@ import SubmissionConfirmation from '../../pages/apply/submissionConfirmation'
 import { mapApiPersonRisksForUi } from '../../../server/utils/utils'
 import { setup } from './setup'
 import { AND, GIVEN, THEN, WHEN } from '../../helpers'
-import { fullPersonFactory } from '../../../server/testutils/factories/person'
 
 context('Apply', () => {
   beforeEach(setup)
 
-  it('allows completion of the form', function test() {
+  xit('allows completion of the form', function test() {
     AND('I complete the application')
     const uiRisks = mapApiPersonRisksForUi(this.application.risks)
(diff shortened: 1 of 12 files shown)


integration_tests/tests/apply/apply.cy.ts
```diff
@@ -22,12 +22,11 @@ import SubmissionConfirmation from '../../pages/apply/submissionConfirmation'
 import { mapApiPersonRisksForUi } from '../../../server/utils/utils'
 import { setup } from './setup'
 import { AND, GIVEN, THEN, WHEN } from '../../helpers'
-import { fullPersonFactory } from '../../../server/testutils/factories/person'
 
 context('Apply', () => {
   beforeEach(setup)
 
-  it('allows completion of the form', function test() {
+  xit('allows completion of the form', function test() {
```
SaintRhys: Skipped test here

bobmeredith: Thanks - good spot.
I thought the linter would pick that up actually.  Maybe it only notices `it.only()`

SaintRhys: I would have thought that too! I've been caught out with a few leftover `fit` in mine

bobmeredith:
Yes, @SaintRhys I noticed that, but it then it would be inconsistent with all the other pages and, in my book, consistency overrides designs.
If it was one of those funny pages where there are no actual form controls then maybe.

PR #3035, opened 2026-08-10
FM-991 Exceptional case details page update

bobmeredith:
# Context

https://dsdmoj.atlassian.net/browse/FM-991
https://dsdmoj.atlassian.net/browse/FM-993

# Changes in this PR

- Updates content on exceptional details page
- Removes radio buttons and reveal on the page
- Updates error messages accordingly

## Screenshots of UI changes

<details>
  <summary>Before</summary>
image
</details>

<details>

  <summary>After</summary>
image
<hr/>
image
</details>

### DIFF
--- a/server/form-pages/apply/reasons-for-placement/basic-information/exceptionDetails.ts
+++ b/server/form-pages/apply/reasons-for-placement/basic-information/exceptionDetails.ts
@@ -1,5 +1,4 @@
 import type { ObjectWithDateParts, TaskListErrors, YesOrNo } from '@approved-premises/ui'
-import { sentenceCase } from '../../../../utils/utils'
 import { Page } from '../../../utils/decorators'
 import { DateFormats, dateAndTimeInputsAreValidDates } from '../../../../utils/dateUtils'
 
@@ -18,13 +17,12 @@ export type ExceptionDetailsBody = ObjectWithDateParts<'agreementDate'> & {
   mergeBody: true,
 })
 export default class ExceptionDetails implements TasklistPage {
-  title = 'Provide details for exemption application'
+  title = 'Exceptional case details'
 
   questions = {
-    agreedCaseWithManager: 'Have you agreed the case with a senior manager?',
-    managerName: 'Name of senior manager',
-    agreementDate: 'What date was this agreed?',
-    agreementSummary: 'Provide a summary of the reasons why this is an exempt application',
+    managerName: 'Name of senior manager who approved exemption',
+    agreementDate: 'Date of approval',
+    agreementSummary: 'Reason for exceptional case',
   }
 
   body: ExceptionDetailsBody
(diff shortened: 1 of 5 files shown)


server/form-pages/apply/reasons-for-placement/basic-information/exceptionDetails.ts
```diff
@@ -74,24 +65,19 @@ export default class ExceptionDetails implements TasklistPage {
...
-      }
-
-      if (!this.body.managerName) {
-        errors.managerName = 'You must provide the name of the senior manager'
-      }
+    if (
+      !this.body.agreementDate ||
+      !dateAndTimeInputsAreValidDates(this.body as ObjectWithDateParts<'agreementDate'>, 'agreementDate')
+    ) {
+      errors.agreementDate = 'Enter a date of approval'
```
SaintRhys, line 70: Not called out in the ticket so can ignore, but I can see other places in cas we make a distinction between a blank date and an invalid date when it comes to error messages.

server/form-pages/apply/reasons-for-placement/basic-information/exceptionDetails.ts
```diff
@@ -18,13 +17,12 @@ export type ExceptionDetailsBody = ObjectWithDateParts<'agreementDate'> & {
   mergeBody: true,
 })
 export default class ExceptionDetails implements TasklistPage {
```
SaintRhys, line 18: Are we able to safety remove the rest of the `agreedCaseWithManager`, or would that mess with in progress applications

server/form-pages/apply/reasons-for-placement/basic-information/exceptionDetails.ts
bobmeredith: Yes, apologies but I should have commented somewhere.
The ticket didn't specify a message for the 'invalid' date case so I checked with Pauline and the response was that the message would be the same.
This isn't really consistent with the rest of the site, but frankly, date validation is all a bit of a mess.  The designers want to move to validating each date field separately as GDS specify, but that is clearly a site-wide thing.

server/form-pages/apply/reasons-for-placement/basic-information/exceptionDetails.ts
bobmeredith: Having thought about it, I think I can tidy up the `exceptionDetails` page itself, but I need to leave where `agreedCaseWithManager` affects the `isInelligble() `state test. 
Existing applications where users have selected `yes` on the `is-exceptional-case` page, but then `no` to the `agreedCaseWithManager` on this page, will find the application is, incorrectly, no longer ineligible. Since all the page's error conditions are already satisfied, they would be able to continue the application to submission even though ineligible.

